Issue #59, opened 2019-07-25
Genesis block is different on re-compilation


tjwelde:
Figure out, what is different between the genesis blocks of the running testnet and the re-compilation.

We might already want to introduce a static json file for the testnet genesis block

LeonFLK:
"Because rust -> wasm builds aren't "reproducible", each person will get a slightly different wasm blob which will break consensus if each participant generates the file themselves" from https://substrate.dev/docs/en/tutorials/start-a-private-network-with-substrate#start-your-private-blockchain

AWS Image compiled different wasm binary than the DockerHub Image.

AWS chain spec can be exported, converted into raw format, then included in different Image or local non docker build for execution, resulting in identical genesis block when included with --chain /path/to/mount/chainSpecRawAWS.json upon execution.

This wont work with our start-node.sh script since this does not include the option to state a spec file, so we would have to extend it's capabilities.

Right now it only works with explicit node execution:

``
docker run -v  /home/leon/HubAccess/chainSpecRawAWS.json:/HubAccess/chainSpecRawAWS.json:ro -p 9941:9944 kiltprotocol/mashnet-node:latest ./target/debug/node --chain /HubAccess/chainSpecRawAWS.json --port 30333 --bootnodes /ip4/18.184.59.35/tcp/30333/p2p/Qmf9Vcxjf5woQP9Znv7xnahCLbA6vXFm8PfnqWcDGgr4Ve /ip4/52.59.203.231/tcp/30333/p2p/QmTKngF1X4Zawh5Zi5sUq6F6o1NQbTPFnrXY8QpfpnkstH --name leonDockerHubHackedNode --base-path ./testnetdata/leondevnode --telemetry-url `ws://telemetry-backend.kilt-prototype.tk:1024
``


this is the AWS chain spec in RAW format
[chainSpecRaw](https://github.com/KILTprotocol/mashnet-node/files/3505964/chainSpecRawAWS.zip)

LeonFLK:
[start-node.sh.zip](https://github.com/KILTprotocol/mashnet-node/files/3520200/start-node.zip)

updated start-node.sh taking -s --chain-spec as optional parameter

have to mount the spec.json at creation time or have the file existing in the image before.

My proposal would be to rebuild the Docker-Hub Image to include the updated start-node.sh and the AWS raw chainspec.json file

LeonFLK:
fixed by #61

PR #61, opened 2019-08-22
chainspec in RAW format and updated start-node.sh

LeonFLK:
- chainspec in RAW format provided
- .start-node.sh now automatically uses the RAW chainspec provided

~~Do We need to update the dockerfile to copy these files?~~
e.g. 
`
COPY --from=builder /build/chainspec.json ./chainspec.json
`
- Yes


tjwelde:
fixes #59

LeonFLK:
README.md was not necessary to update

PR #63, opened 2019-08-30
Release Mode Building Preparation

LeonFLK:
[Documentation on different profile defaults](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections)

Our current build mode is the default invoked dev profile, which is unoptimized (slower code), but has debugging and checking options enabled and building is faster:

```
[profile.dev]
opt-level = 0   
debug = true      
rpath = false      
lto = false        
debug-assertions = true 
codegen-units = 16 
panic = 'unwind'   
incremental = true
overflow-checks = true
```

The release build mode has high optimization thus slower build time, no debugging/checks enabled, but resulting code is much faster.
```
[profile.release]
opt-level = 3
debug = false
rpath = false
lto = false
debug-assertions = false
codegen-units = 16
panic = 'unwind'
incremental = false
overflow-checks = false
```

My proposed changes would have the docker image automatically building in release mode.

On my machine no size changes occurred in different build modes all images 256 MB.


Dockerfile
```diff
@@ -28,9 +28,6 @@ RUN rustup target add wasm32-unknown-unknown --toolchain nightly-2019-07-14
 # (polkadot)
 RUN cargo +nightly install --git https://github.com/alexcrichton/wasm-gc
 
-# setup default stable channel
-RUN rustup default nightly-2019-07-14
```
tjwelde: Why can we remove this in release mode?

LeonFLK: It's a redundant command, nothing to do with release mode. 
We already set it in line 24.

PR #66, opened 2019-09-18
feat #60: substrate update

### LINKED ISSUE #60: WSH - Substrate Update

bekolb:
## fixes #60 
Update node and runtime to current substrate node template 1.0.0
Also renames all occurrences of `node-template` to `mashnet-node`

## How to test:
Launch a mashnet and run through every functionality of KILT that involves the blockchain.

## Checklist:

- [x] I have verified that the code works (by running a single node)


tjwelde:
@LeonFLK Please test with following instructions:
- Run a set of nodes (more than 2 / mix of validators and full-nodes)
- Compile the wasm-runtime from this PR and update the chain (via the chain-explorer)
- Natively update the nodes

Observe, what happens and report, if something went wrong.
Thanks

LeonFLK:
Upgrading via https://substrate-ui.parity.io/#/toolbox with the sudo module bricks the chain with the error message:

**Aura authorship task terminated unexpectedly. Restarting**

The Chain stops producing blocks at this point. The new substrate runtime version introduced breaking changes and is not compatible with our Testnet.
I will now test the chain functionality with our demo-client

LeonFLK:
Created new chainspec to share in our network.
Addresses seem to be calculated differently, sudo key and authority and endowed ids changed last 4 characters.
demo client is not able to really call the API/RPC, getting errors like: 
 - RPC-CORE: getMetadata (block: Hash): Metadata:: Object.values(...)[index] is not a constructor
 - API/DECORATOR: loadMeta ExtError: getMetadata (block: Hash): Metadata:: Object.values(...)[index] is not a constructor

tjwelde:
I tested the whole functionality with the demo client and it works very good.

tjwelde:
I think we need to use the new chainspec and restart the chain. We might need to reconfigure our validators on aws.

PR #70, opened 2019-10-09
feature: start-node.sh is takes node key and seed

LeonFLK:
## fixes https://github.com/KILTprotocol/ticket/issues/40
start-node.sh now takes the node's key and seed as parameters when giving --validator as argument


tjwelde:
We can only do this, when we make sure, that we can use environment variables for these values on our servers

PR #75, opened 2020-01-30
Portablegabi

weichweich:
## Store portablegabi accumulator on chain
The accumulator which is used to revoke anonymous attestation can be stored on chain.

## How to test:

Once the SDK is ready, create and revoke attestations.

## Checklist:

- [ ]  wait for #74 to be merged and rebase


wischli:
Successfully tested `portablegabi` module with Polkadot API, no problems encountered.

PR #78, opened 2020-02-13
feat: return Option from queries

flechtnergalani:
Fixes https://github.com/KILTprotocol/ticket/issues/304
Fixes https://github.com/KILTprotocol/ticket/issues/222
The return value for queries where the query does not match anything is defined as follows in substrate:
a. If the return type is an Option or wrapped in an option, return `Option<null>`
b. If the return type is not an Option, return the fallback value, which defaults to a Hex zero value of the appropriate length (`0x00000...`)

So far, we were relying on fallback values, but these make results hard to interpret in the SDK ([see this issue, for example](https://github.com/KILTprotocol/sdk-js/pull/218)). 
This set of changes wraps return values in an Option where query methods were relying on fallback values. 
As an example, `ctype.ctypes()` will now return an `Option<AccountID>`. When the Ctype hash provided is not stored on chain, the `Option<null>` returned will will be decoded to `null` by `.toJSON()`, while the result of a successful query is decoded directly to the account address string. 

These are non-breaking changes w/r/t the SDK. Changes to reflect the new return types are proposed in https://github.com/KILTprotocol/sdk-js/pull/235

## How to test:
In addition to unit tests:
1. (Re)build node and runtime
2. purge chain
3. Clone the SDK and run the following test file with yarn jest filename:
```
import getCached from './blockchainApiConnection'

test('query ctype', async () => {
  await expect(
    getCached()
      .then(bc => bc.api.query.ctype.cTYPEs('0x1235435'))
      .then(r => r.toJSON())
  ).resolves.toBeNull()
})

test('attestations', async () => {
  await expect(
    getCached()
      .then(bc => bc.api.query.attestation.attestations('0x1235435'))
      .then(r => r.toJSON())
  ).resolves.toBeNull()
})

test('delegations', async () => {
  await expect(
    getCached()
      .then(bc => bc.api.query.delegation.delegations('0x12309734'))
      .then(r => r.toJSON())
  ).resolves.toBeNull()
})

test('delegation root', async () => {
  await expect(
    getCached()
      .then(bc => bc.api.query.delegation.root('0x12309734'))
      .then(r => r.toJSON())
  ).resolves.toBeNull()
})

test('did', async () => {
  await expect(
    getCached()
      .then(bc => bc.api.query.did.dIDs('0x12309734'))
      .then(r => r.toJSON())
  ).resolves.toBeNull()
})

afterAll(async () => {
  const bc = await getCached()
  return bc.api.disconnect()
})
```

## Checklist:

- [x] I have checked that the SDK is compatible with the Option return values

### DIFF
--- a/runtime/src/attestation/mod.rs
+++ b/runtime/src/attestation/mod.rs
@@ -71,11 +71,10 @@ decl_module! {
 				// has a delegation
 				Some(d) => {
 					// check if delegation exists
-					if !<delegation::Delegations<T>>::exists(d.clone()) {
-						return Self::error(<delegation::Module<T>>::ERROR_DELEGATION_NOT_FOUND);
-					}
-					let delegation = <delegation::Delegations<T>>::get(d.clone())
-						.ok_or_else(||Self::error(<delegation::Module<T>>::ERROR_DELEGATION_NOT_FOUND).err().unwrap())?;
+					let delegation = <error::Module<T>>::ok_or_deposit_err(
+						<delegation::Delegations<T>>::get(d.clone()),
+						<delegation::Module<T>>::ERROR_DELEGATION_NOT_FOUND
+					)?;
 					if delegation.4 {
 						// delegation has been revoked
 						return Self::error(Self::ERROR_DELEGATION_REVOKED);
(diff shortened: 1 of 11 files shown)


runtime/src/attestation/mod.rs
```diff
@@ -84,7 +84,7 @@ decl_module! {
 						return Self::error(Self::ERROR_DELEGATION_NOT_AUTHORIZED_TO_ATTEST);
 					} else {
 						// check if CTYPE of the delegation is matching the CTYPE of the attestation
-						let root = <delegation::Root<T>>::get(delegation.0.clone());
+						let root = <delegation::Root<T>>::get(delegation.0.clone()).ok_or(<delegation::Module<T>>::ERROR_ROOT_NOT_FOUND.1)?;
```
weichweich: We need to change this. I will talk to you in person, but we need to use the error module...

flechtnergalani:
@tjwelde SDK is compatible, ready for merge

PR #80, opened 2020-03-04
chore: fixed a few more lint errors missed earlier

flechtnergalani:
a follow-up for #74. Catches a few more linter errors.

## How to test:
- [x] unit tests passing
- [x] SDK integration tests passing

## Checklist:

### DIFF
--- a/runtime/src/attestation/tests.rs
+++ b/runtime/src/attestation/tests.rs
@@ -18,7 +18,7 @@
 
 use super::*;
 use parity_codec::Encode;
-use primitives::{ed25519 as x25519, Blake2Hasher, H256, *};
+use primitives::{ed25519 as x25519, Blake2Hasher, H256, Pair};
 use runtime_io::with_externalities;
 use support::{assert_err, assert_ok, impl_outer_origin};
 use system;
--- a/runtime/src/attestation/tests.rs
+++ b/runtime/src/attestation/tests.rs
@@ -18,8 +18,7 @@
 
 use super::*;
 use parity_codec::Encode;
-use primitives::*;
-use primitives::{ed25519 as x25519, Blake2Hasher, H256};
+use primitives::{ed25519 as x25519, Blake2Hasher, H256, *};
 use runtime_io::with_externalities;
 use support::{assert_err, assert_ok, impl_outer_origin};
 use system;
--- a/runtime/src/attestation/tests.rs
+++ b/runtime/src/attestation/tests.rs
@@ -21,7 +21,6 @@ use parity_codec::Encode;
 use primitives::{ed25519 as x25519, Blake2Hasher, H256, Pair};
 use runtime_io::with_externalities;
 use support::{assert_err, assert_ok, impl_outer_origin};
-use system;
 
 use runtime_primitives::{
 	testing::{Digest, DigestItem, Header},
--- a/runtime/src/attestation/tests.rs
+++ b/runtime/src/attestation/tests.rs
@@ -82,7 +82,7 @@ fn new_test_ext() -> runtime_io::TestExternalities<Blake2Hasher> {
 }
 
 fn hash_to_u8<T: Encode>(hash: T) -> Vec<u8> {
-	return hash.encode();
+	hash.encode()
 }
 
 #[test]
@@ -93,21 +93,21 @@ fn check_add_attestation() {
 		let account_hash = pair.public();
 		assert_ok!(CType::add(
 			Origin::signed(account_hash.clone()),
-			hash.clone()
+			hash
 		));
 		assert_ok!(Attestation::add(
 			Origin::signed(account_hash.clone()),
-			hash.clone(),
-			hash.clone(),
+			hash,
+			hash,
 			None
 		));
 		let existing_attestation_for_claim = {
-			let opt = Attestation::attestations(hash.clone());
+			let opt = Attestation::attestations(hash);
 			assert!(opt.is_some());
 			opt.unwrap()
 		};
-		assert_eq!(existing_attestation_for_claim.0, hash.clone());
-		assert_eq!(existing_attestation_for_claim.1, account_hash.clone());
+		assert_eq!(existing_attestation_for_claim.0, hash);
+		assert_eq!(existing_attestation_for_claim.1, account_hash);
 		assert_eq!(existing_attestation_for_claim.3, false);
 	});
 }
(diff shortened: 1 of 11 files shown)


weichweich:
A few more suggestions:

- [ ] add `cargo clippy -- -D warnings` to CI pipeline
- [ ] maybe use `apt-get` instead of `apt` inside the dockerfile
- [ ] add target directory (cargo build directory) to `.dockerignore`

I will add this to the PR if I have time and its ok for you.

flechtnergalani:
> A few more suggestions:
> 
> * [ ]  add `cargo clippy -- -D warnings` to CI pipeline
> * [ ]  maybe use `apt-get` instead of `apt` inside the dockerfile
> * [ ]  add target directory (cargo build directory) to `.dockerignore`
> 
> I will add this to the PR if I have time and its ok for you.

absolutely! go ahead

flechtnergalani:
> * [ ]  add `cargo clippy -- -D warnings` to CI pipeline
there are some linter warnings we haven't gotten rid of yet - maybe we could use this instead:
https://github.com/marketplace/actions/rust-clippy-check

weichweich:
> there are some linter warnings we haven't gotten rid of yet - maybe we could use this instead:
> https://github.com/marketplace/actions/rust-clippy-check

I like that idea very much! Maybe we should also exclude the tests from the docker build and run them separately. We could cache build dependencies and speed up the tests.

flechtnergalani:
now that I think about it, the docker part is not really in the scope of this PR. I'll cherry-pick & open another for this. The CI linter improvements are a little more appropriate here, but more a feat than a chore, right?

weichweich:
It could also make sense to split up the compilation in first only the dependencies and second the whole project. (like it's done [here](http://whitfin.io/speeding-up-rust-docker-builds/))

And yes that might be a different PR

flechtnergalani:
> > there are some linter warnings we haven't gotten rid of yet - maybe we could use this instead:
> > https://github.com/marketplace/actions/rust-clippy-check
> 
> I like that idea very much! Maybe we should also exclude the tests from the docker build and run them separately. We could cache build dependencies and speed up the tests.

This does not run in the docker though, I guess we'd have to introduce a new job just for linting.  Not sure that's worth it?

weichweich:
> This does not run in the docker though, I guess we'd have to introduce a new job just for linting. Not sure that's worth it?

In my opinion having a dedicated linting step is definitely worth it. I don't like putting stuff into the docker container...

runtime/src/attestation/tests.rs
```diff
@@ -18,8 +18,7 @@
 
 use super::*;
 use parity_codec::Encode;
-use primitives::*;
-use primitives::{ed25519 as x25519, Blake2Hasher, H256};
+use primitives::{ed25519 as x25519, Blake2Hasher, H256, *};
```
weichweich: Either we import everything by name or we import using the wildcard. We could remove `Blake2Hasher, H256`. But I don't like wildcard imports. There might even be an option to convert wildcard imports to explicit imports in RLS.

runtime/src/delegation/tests.rs
```diff
@@ -18,8 +18,7 @@
 
 use super::*;
 use parity_codec::Encode;
-use primitives::*;
-use primitives::{ed25519 as x25519, Blake2Hasher, H256, H512};
+use primitives::{ed25519 as x25519, Blake2Hasher, H256, H512, *};
```
weichweich: same here

runtime/src/did/tests.rs
```diff
@@ -17,8 +17,7 @@
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
 use super::*;
-use primitives::*;
-use primitives::{Blake2Hasher, H256};
+use primitives::{Blake2Hasher, H256, *};
```
weichweich: and here

runtime/src/attestation/tests.rs
flechtnergalani: ok let's see what we can do about these

runtime/src/did/tests.rs
flechtnergalani: what works here is `use primitives::{ed25519, Blake2Hasher, H256, Pair};` - it does complain about unused imports though (Pair)

runtime/src/attestation/tests.rs
flechtnergalani:
```
use primitives::{ed25519, ed25519 as x25519, Blake2Hasher, H256, Pair};
```
works here - but we'd have to suppress unused imports warnings

runtime/src/delegation/tests.rs
flechtnergalani: this could be ```use primitives::{ed25519, ed25519 as x25519, Blake2Hasher, H256, H512, Pair};
```

File: runtime/src/attestation/tests.rs (hunk 1 above)
Comment (weichweich) [reply runtime/src/attestation/tests.rs]: And we need the Pair import?

Comment (weichweich) [reply runtime/src/attestation/tests.rs]: Ah yes you can remove `ed25519` from the imports. There should be no unused warning after that...

File: runtime/src/did/tests.rs (hunk 3 above)
Comment (weichweich) [reply runtime/src/did/tests.rs]: Strange, I don't get a warning here. 😕

File: runtime/src/attestation/tests.rs (hunk 1 above)
Comment (flechtnergalani) [reply runtime/src/attestation/tests.rs]: we need the pair import: 
```
error[E0599]: no function or associated item named `from_seed` found for struct `substrate_primitives::ed25519::Pair` in the current scope
   --> runtime/src/attestation/tests.rs:246:36
    |
246 |         let pair_charlie = x25519::Pair::from_seed(*b"Charlie                         ");
    |                                          ^^^^^^^^^ function or associated item not found in `substrate_primitives::ed25519::Pair`
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use substrate_primitives::crypto::Pair;
```
and the Pair import then required the ed25519 import

weichweich: `use primitives::{ed25519 as x25519, Blake2Hasher, H256, Pair};` works for me.

weichweich: ah ok. I see. You get a clippy warning there. That's strange and probably a bug in clippy...

weichweich: Try calling clippy with `cargo clippy --tests`. It's probably unused because clippy doesn't see the test methods or something like that...

flechtnergalani:
I think we have them all now!

PR #81, opened 2020-03-26
chore: Update to Substrate v2.0.0-rc4

weichweich:
## Update to Substrate 2.0.0

- fixes KILTProtocol/ticket#533
- closes KILTprotocol/ticket/issues/549
- closes KILTprotocol/ticket/issues/406

## ToDos

- [x] update the chain specs
- [x] "Restore" v1.0 payment fees
- [x] Remove weight?
- [x] The block time after the update is 6s, it should be 10s.
     - it should be configured in the genesis block?!
- [x] What was the index for? Do we need an index?

## Checklist:


runtime/src/lib.rs
```diff
@@ -139,82 +137,107 @@ pub fn native_version() -> NativeVersion {
...
 }
 
-impl consensus::Trait for Runtime {
-	/// The identifier we use to refer to authorities.
-	type SessionKey = AuthorityId;
-	// The aura module handles offline-reports internally
-	// rather than using an explicit report system.
-	type InherentOfflineReport = ();
-	/// The ubiquitous log type.
-	type Log = Log;
```
weichweich, line 178: replaced by grandpa+aura? This might need review/research...

runtime/src/lib.rs
```diff
@@ -139,82 +137,107 @@ pub fn native_version() -> NativeVersion {
...
-impl indices::Trait for Runtime {
-	/// The type for recording indexing into the account enumeration. If this ever overflows, there
-	/// will be problems!
-	type AccountIndex = u32;
-	/// Use the standard means of resolving an index hint from an id.
-	type ResolveHint = indices::SimpleResolveHint<Self::AccountId, Self::AccountIndex>;
-	/// Determine whether an account is dead.
-	type IsDeadAccount = Balances;
-	/// The ubiquitous event type.
-	type Event = Event;
```
weichweich, line 190: What was the purpose of this? It's removed and seems to make no difference. This definitely needs a review...

First guess is that it's for enumerating all available accounts?

Take from [substrate](https://github.com/paritytech/substrate/blob/29454c30501ccf3a09fe82b965dd50e1a9e5aa24/frame/indices/src/lib.rs):
>//! An index is a short form of an address. This module handles allocation
>//! of indices for a newly created accounts.

runtime/src/lib.rs
```diff
@@ -139,82 +137,107 @@ pub fn native_version() -> NativeVersion {
...
+	pub const TransactionByteFee: Balance = 1;
+}
+
+impl transaction_payment::Trait for Runtime {
+	type Currency = balances::Module<Runtime>;
+	type OnTransactionPayment = ();
+	type TransactionBaseFee = TransactionBaseFee;
+	type TransactionByteFee = TransactionByteFee;
+	type WeightToFee = ConvertInto;
+	type FeeMultiplierUpdate = ();
```
weichweich, line 235: This changes the way transaction payments are calculated

weichweich: But that's the pallet for calculating transaction fees. We might need to configure this to our liking. 😄

weichweich:
- [ ] The block time after the update is 6s, it should be 10s.

runtime/src/lib.rs
weichweich: added again ffa5d49603f9d3e797a65f55d726d4ceafb60951

weichweich:
before the update is merged we need to update the chain specs! 😱

runtime/src/ctype/tests.rs
```diff
@@ -18,68 +18,96 @@
...
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Clone, Eq, PartialEq, Debug)]
 pub struct Test;
+
+parameter_types! {
+    pub const BlockHashCount: u64 = 250;
+    pub const MaximumBlockWeight: Weight = 1_000_000_000;
+    pub const MaximumBlockLength: u32 = 5 * 1024 * 1024;
+    pub const AvailableBlockRatio: Perbill = Perbill::from_percent(75);
+    // FIXME: Wait for workaround
```
wischli, line 48: Placeholder to be aware of this.

wischli: Fixed in 466ab19 by removing `const` declaration from `MaximumExtrinsicWeight` 😅

src/command.rs
```diff
@@ -0,0 +1,72 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
```
flechtnergalani, line 15: figure out what to do with this license header

Dockerfile
```diff
@@ -1,41 +1,8 @@
...
-# Copy WASM build crate files
-COPY ./runtime/wasm/build.sh ./runtime/wasm/Cargo.lock ./runtime/wasm/Cargo.toml ./runtime/wasm/
-COPY ./runtime/wasm/src ./runtime/wasm/src
-
-# get build script and build
-COPY ./scripts/build.sh /scripts/build.sh
-RUN /bin/bash /scripts/build.sh
-
-# this container builds the mashnet-node binary from source files, the runtime library and the WASM file built previously
 FROM rustlang/rust@sha256:9ac425a47e25a7a5dac999362b89de2b91b21ce70c557a409c46280393f7b1f1 as builder
```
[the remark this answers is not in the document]
flechtnergalani: it can! so I'd suggest we go back to `rustlang/rust:nightly`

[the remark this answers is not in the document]
wischli: Or we could also use the Parity rust image [`paritytech/ci-linux:production`](https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux) which we already use in the [Portablegabi Node](https://github.com/KILTprotocol/portablegabi-node/blob/master/Dockerfile). This used to be called [`paritytech/rust-builder`](https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux).

src/command.rs
wischli: @weichweich Remove it?

weichweich: > @weichweich Remove it?

I guess GPL doesn't allow that. Just keep it. We may be able to add our copyright to the one of parity?

```
 // Copyright 2017-2020 Parity Technologies (UK) Ltd. 
 // Copyright 2019-2020 Botlabs GmbH
```

Or something like that...

wischli: I like that. We should check with @GustavHemmelmayr.

src/chain_spec.rs
```diff
@@ -88,22 +102,24 @@ impl Alternative {
...
 					"kilt_testnet",
+					ChainType::Live,
 					|| {
 						testnet_genesis(
 							vec![
-					x25519::Public::from_raw(hex!("58d3bb9e9dd245f3dec8d8fab7b97578c00a10cf3ca9d224caaa46456f91c46c")),
-					x25519::Public::from_raw(hex!("d660b4470a954ecc99496d4e4b012ee9acac3979e403967ef09de20da9bdeb28")),
-					x25519::Public::from_raw(hex!("2ecb6a4ce4d9bc0faab70441f20603fcd443d6d866e97c9e238a2fb3e982ae2f")),
-				],
+							get_authority_keys_from_seed("0x58d3bb9e9dd245f3dec8d8fab7b97578c00a10cf3ca9d224caaa46456f91c46c"),
```
tjwelde, line 109: This is supposed to be a public key

runtime/src/lib.rs
```diff
@@ -19,353 +19,476 @@
...
-	spec_name: create_runtime_str!("mashnet-node"),
-	impl_name: create_runtime_str!("mashnet-node"),
-	authoring_version: 3,
-	spec_version: 4,
-	impl_version: 4,
-	apis: RUNTIME_API_VERSIONS,
+    spec_name: create_runtime_str!("mashnet-node"),
+    impl_name: create_runtime_str!("mashnet-node"),
+    authoring_version: 4,
+    spec_version: 4,
```
tjwelde, line 128: let' increase this, too

runtime/src/lib.rs
```diff
@@ -19,353 +19,476 @@
...
+    impl_version: 5,
```
tjwelde, line 129: Can this start at 0/1 again?

wischli: Me
> Short question about the incrementing of the runtime version. What would be the recommended best practice when we switch from Substrate 1.x to 2.0?
1. Increment all version numbers
2. "Reset" to 1 if current versions > 1

Jam from Parity

> It only really matters if you have external tooling pointing to your chains and you have some kind of continuity of the infrastructure between versions. I generally don't think it's a good idea to go backwards in version number unless you are changing the name or something.

runtime/src/lib.rs
weichweich: fixed 7b3fdb9

src/chain_spec.rs
weichweich: fixed in 8d75804

runtime/src/lib.rs
tjwelde: Not sure, if he did understand the questions correctly. We only want to put `impl_version` to 1, while increasing `authoring_version` and `spec_version`.
Like going from `0.2.8` to `0.3.0`.
But we actually do not have the restrictions, he is talking about, so I guess it doesn't matter. Would still be interesting to know, how it is supposed to be done.

tjwelde: Gav said something about this: https://github.com/paritytech/substrate/issues/241

tjwelde: Also only really matters, when updating the runtime on-chain.

src/chain_spec.rs
tjwelde: Have a look at how polkadot is doing it: https://github.com/paritytech/polkadot/blob/848e1a892f99c16015b75117fdae00b393cde770/service/src/chain_spec.rs#L365
Can't we do it similarly?

tjwelde:
LGTM

src/command.rs
GustavHemmelmayr: well, comirights for two companies for same software is confusing unless we really programmed together - stuff that is done by us can be labled for BOTLabs GmbH (please spell with the first three letters as capital letters) and stuff that is done by Parity is labeld by them and we just refer to it, right?

GPL does allow a reference to the company, please always use the original wording in the license part of the software.

PR #82, opened 2020-03-26
feature: faster docker image creation

flechtnergalani:
## relates to https://github.com/KILTprotocol/ticket/issues/415 ; cherry-picked from KILTprotocol/mashnet-node#80
Speeds up docker builds by using a rust-nightly image, thus skipping the rust installation bit.
Also compiles dependencies in separate steps to avoid cache invalidation by source file changes.
A `docker-compose.yml` is included to quickly spin up a dev node on port `9944` with a single `docker-compose up dev-node`. You can also run sdk integration tests against your local code with `docker-compose run sdk yarn test:integration` !

## How to test:
build and run the image with `docker-compose build` and use it. Run integration tests against it with `docker-compose run sdk yarn test:integration`.

## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 .git
 .github
 task-definition*
+**/target
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,5 @@
 task-definition*
 **/target
 docker-compose.yml
-LICENSE
 Dockerfile
 task-definition*
-README*
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,10 @@
 .git
 .github
+.gitignore
 task-definition*
 **/target
+docker-compose.yml
+LICENSE
+Dockerfile
+task-definition*
+README*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,6 @@
-FROM rustlang/rust:nightly as builder
-
-WORKDIR /build
-
-RUN apt-get -y update && \
-	apt-get install -y --no-install-recommends \
-	clang
+# the WASM build of the runtime is completely indepedent 
+# we can avoid cache invalidations by running it in an extra container
+FROM rustlang/rust:nightly as wasm_builder
 
 # install wasm toolchain for polkadot
 RUN rustup target add wasm32-unknown-unknown --toolchain nightly
@@ -16,6 +12,7 @@ RUN cargo +nightly install --git https://github.com/alexcrichton/wasm-gc
 # show backtraces
 ENV RUST_BACKTRACE 1
 
+## not sure which of theses ENV are actually needed for this step
 #compiler ENV
 ENV CC gcc
 ENV CXX g++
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -1,3 +1,10 @@
 .git
 .github
+.gitignore
 task-definition*
+**/target
+docker-compose.yml
+LICENSE
+Dockerfile
+task-definition*
+README*
```
weichweich, line 10: Why do you ignore `.gitignore`, `LICENSE`, `README` and the `task-definitions`?

Dockerfile
```diff
@@ -1,23 +1,6 @@
...
-
-# setup rust beta and nightly channel's
-RUN rustup install nightly
-RUN rustup install stable
+# the WASM build of the runtime is completely indepedent 
+# we can avoid cache invalidations by running it in an extra container
+FROM rustlang/rust:nightly as wasm_builder
 
 # install wasm toolchain for polkadot
 RUN rustup target add wasm32-unknown-unknown --toolchain nightly
```
weichweich, line 6: instead of that, you could just call the init script?

Dockerfile
```diff
@@ -29,11 +12,40 @@ RUN cargo +nightly install --git https://github.com/alexcrichton/wasm-gc
...
+# Copy WASM build crate files
+COPY ./runtime/wasm/build.sh ./runtime/wasm/Cargo.lock ./runtime/wasm/Cargo.toml ./runtime/wasm/
+COPY ./runtime/wasm/src ./runtime/wasm/src
 
+# get build script and build
+COPY ./scripts/build.sh /scripts/build.sh
+RUN /bin/bash /scripts/build.sh
+
+# this container builds the mashnet-node binary from source files, the runtime library and the WASM file built previously
+FROM rustlang/rust:nightly as builder
```
weichweich, line 36: couldn't you just use the same container and build the binaries?

Dockerfile
```diff
@@ -42,20 +54,36 @@ ENV CXX g++
...
+COPY --from=wasm_builder /runtime/wasm/target/wasm32-unknown-unknown/release ./runtime/wasm/target/wasm32-unknown-unknown/release
+# build source again, dependencies are already built
 RUN cargo build --release
+
+# test
 RUN cargo test --release -p mashnet-node-runtime
 
 
-FROM ubuntu:xenial
+FROM debian:stretch
```
weichweich, line 81: Why not use `buster` which is the current release?

.dockerignore
flechtnergalani: I tried to exclude anything that's not build-relevant; why should they be copied to the container if they're useless there?

Dockerfile
flechtnergalani: absolutely, but was like this and did not see a reason to change

Dockerfile
flechtnergalani: this would lead to earlier cache invalidation in case the runtime is changed, I think

Dockerfile
flechtnergalani: the rust nightly image is build on stretch, wanted to keep things as similar as possible

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+
+services:
+  dev-node:
+    build: .
+    command: ./target/release/mashnet-node --dev --ws-port 9944 --ws-external
+    ports:
+      - 9944:9944
+  sdk:
+    build:
+      context: https://github.com/KILTprotocol/sdk-js.git#rf-docker-compose
```
flechtnergalani, line 11:
```suggestion
      context: https://github.com/KILTprotocol/sdk-js.git
```
will have to change once the Dockerfile is merged to develop on the sdk

.dockerignore
weichweich: I am no license expert, but maybe we should include a license with our container? Not sure if it is safe to remote it.

flechtnergalani: sounds reasonable. I'll put LICENSE and README back in

flechtnergalani:
@weichweich The LICENSE and README are now in the image. I also set the build branch for the SDK to develop, even though we haven't merged the branch containing a dockerfile there.

PR #83, opened 2020-04-15
docs: updated node binary paths in README

flechtnergalani:
## fixes KILTProtocol/ticket#166
The path of the freshly built mashnet node binary was wrong in the README, which would confuse anyone trying to use it (incl me when I started in December). Fixed this.

## Checklist:

- [ ] ~I have verified that the code works~ 
- [ ] ~I have verified that the code is easy to understand~
  - [ ] ~If not, I have left a well-balanced amount of inline comments~


flechtnergalani:
> Ok.
> 
> maybe you could also remove the `Updating to latest substrate-node-template` section? Since there is no `substrate-node-new` command!? 😅

Sure, you would know best! You've just been through the whole updating process.

PR #84, opened 2020-04-23
feat: portablegabi

weichweich:
## Checklist:

### DIFF
--- a/runtime/src/portablegabi.rs
+++ b/runtime/src/portablegabi.rs
@@ -1,0 +1,148 @@
+use support::{
+	decl_event, decl_module, decl_storage, ensure, dispatch::Result, StorageMap,
+};
+use rstd::vec::Vec;
+use system::ensure_signed;
+
+/// The pallet's configuration trait.
+pub trait Trait: system::Trait {
+	/// The overarching event type.
+	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as TemplateModule {
+		/// The AccumulatorList contains all accumulator. It is a map which
+		/// maps an account id and an index to an accumulator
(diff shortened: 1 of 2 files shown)


flechtnergalani:
Ponder, my brother: If I allocate a chunk of memory on the heap, but nothing ever reads from it; would I still be justified in calling it "memory" ?

runtime/src/portablegabi.rs
```diff
@@ -0,0 +1,148 @@
...
+		fn deposit_event<T>() = default;
+
+		/// Updates the attestation
+		pub fn update_accumulator(origin, accumulator: Vec<u8>) -> Result {
+			let attester = ensure_signed(origin)?;
+
+			let counter = if !<AccumulatorCount<T>>::exists(&attester) {
+				0
+			} else {
+				<AccumulatorCount<T>>::get(&attester)
```
tjwelde, line 37: Might be enough to just use this, when it returns 0 for uninitialized ones

runtime/src/portablegabi.rs
```diff
@@ -0,0 +1,148 @@
...
+			};
+
+			let next = counter.checked_add(1).ok_or("Overflow increasing accumulator index")?;
+			ensure!(!<AccumulatorList<T>>::exists((attester.clone(), next)),
+					"Inconsistent accumulator counter");
```
tjwelde, line 42: Please use error module

runtime/src/portablegabi.rs
weichweich: fixed in cb2ed89659a451adee7333d9459bcbb2fe6a8ee1

runtime/src/portablegabi.rs
weichweich: fixed in cb2ed89659a451adee7333d9459bcbb2fe6a8ee1

PR #85, opened 2020-06-10
publishing dev images on dockerhub

flechtnergalani:
## part of KILTProtocol/ticket#530
On pushes to develop, the docker image built for EWS and deployed to the devnet will also be pushed to the mashnet-node repository on docker hub under a dedicated tag (currently set to `develop`)

## How to test:
the latest commit 1554634 has triggered a successful test run pushing to docker hub ([here](https://github.com/KILTprotocol/mashnet-node/actions/runs/131928883)). You can also pull the image with `kiltprotocol/mashnet-node:develop` and test it.

## Checklist:

### DIFF
--- a/.github/workflows/aws-dev.yml
+++ b/.github/workflows/aws-dev.yml
@@ -151,10 +151,10 @@ jobs:
     - name: Tag and push dev image to Docker Hub
       env:
         ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
-        SOURCE_IMAGE: $ECR_REGISTRY/$ECR_REPOSITORY:$ECR_IMAGE_TAG
         DOCKER_REPOSITORY: kiltprotocol/mashnet-node
         DOCKER_IMAGE_TAG: develop
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-dev.yml
```diff
@@ -120,3 +122,38 @@ jobs:
...
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: eu-central-1
+
+    - name: Login to Amazon ECR
+      id: login-ecr
+      uses: aws-actions/amazon-ecr-login@v1
```
weichweich, line 141: Can we write comments in here? Maybe write a comment, that we need to login because we pull the Image from AWS. I think that might not be obvious at first.

flechtnergalani: sure, can do! hope this works?

PR #86, opened 2020-06-16
ci: cache wasm builder

flechtnergalani:
## no ticket
Given that we introduced another build step during building of docker images, the CI needs to be updated to reflect this change in order to benefit from reduced build times.

## How to test:
a new image will only be pushed when merging to develop; only then the build speed will increase. ~~I could make a temporary change to the workflow to trigger a build on this branch though.~~ <- can't; this would also deploy to the devnet

## Checklist:


weichweich:
Good ideat, but the separate WASM building will be gone after the update... So i wouldn't merge.

flechtnergalani:
ouhhh damn you're right

PR #95, opened 2020-08-14
refactor: move runtime modules to pallets

wischli:
## fixes KILTProtocol/ticket#690
- moves all runtime modules to pallets (enables no code duplication when used in other projects such as kilt-collator)
- refactors dependencies by removing aliases

## How to test:
Check CI

## Checklist:


weichweich:
@wischli Maybe it would be better to put all the modules in a single crate? It looks like a big overhead to maintain all the cargo.tomls separately?

wischli:
> @wischli Maybe it would be better to put all the modules in a single crate? It looks like a big overhead to maintain all the cargo.tomls separately?

We could go with that as well, I just sticked to how it is done on Substrate. IIUC I refactor all 5 pallets to be one instead?

weichweich:
> IIUC I refactor all 5 pallets to be one instead?

Depends on what you think is better. I would be fine with both possibilities.

wischli:
> > IIUC I refactor all 5 pallets to be one instead?
> 
> Depends on what you think is better. I would be fine with both possibilities.

My lazy side says keep as is since this would result in both refactoring here and the collator repo. But it might be annoying in future updates to have to repeat tasks for multiple Cargo.tomls if it cannot be handled by easy search-replacing. So if you ask me, I would probably keep as is.

weichweich:
> My lazy side says keep as is since this would result in both refactoring here and the collator repo. But it might be annoying in future updates to have to repeat tasks for multiple Cargo.tomls if it cannot be handled by easy search-replacing. So if you ask me, I would probably keep as is.

OK lets keep it like it is. If it is annoying we still can refactor that...

PR #96, opened 2020-09-01
arm X-compile

weichweich:


ARM_INSTALL.md
```diff
@@ -0,0 +1,27 @@
+# How to cross compile to arm
+
+download [cross compiler arm-linux-gnueabihf 8.3](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads)
```
tjwelde, line 3: I would add pre-conditions for these instructions (or with which setup you wrote it), like x64 linux debian based using fish shell and normal rust setup

tjwelde (changes requested):

PR #99, opened 2020-09-25
feat: substrate two update

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/740

Substrate 2 update.

## Checklist:


tjwelde (commented):
Does it work as a runtime upgrade?

Dudleyneedham:
Yes, it does work with a runtime update.

I have also checked the sdk + friends. 

It seems the SDK is breaking from the changes, but with https://github.com/KILTprotocol/sdk-js/pull/299 it fixes these changes.

Would be best to roll out with the following changes to how the transactions are handled.

wischli (approved):
LGTM! However, we should wait with the merge until all integration tests inside the SDK are working again.

Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ['KILT <info@kilt.io>']
 build = 'build.rs'
 edition = '2018'
 name = 'mashnet-node'
-version = '0.23.0'
+version = '0.24.0'
```
weichweich, line 6: We should not increase the version number straight away, but wait until we plan to publish a new release.

src/service.rs
```diff
@@ -1,37 +1,17 @@
...
-// The KILT Blockchain is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
-// If you feel like getting in touch with us, you can do so at info@botlabs.org
-
```
weichweich, line 18: maybe we should keep this

Cargo.toml
wischli: Good catch.

src/service.rs
Dudleyneedham: Ahh, yes.

Cargo.toml
Dudleyneedham: But this has to be done to do the new release, right?

src/service.rs
Dudleyneedham: I have re-added this.

Dockerfile
```diff
@@ -42,8 +42,11 @@ COPY . /build
 RUN cargo build --release
 
 # test
-RUN cargo test --release -p mashnet-node-runtime
-
+RUN cargo test --release -p ctype
+RUN cargo test --release -p delegation
+RUN cargo test --release -p did
+RUN cargo test --release -p error
+RUN cargo test --release -p portablegabi
```
weichweich, line 49:
```suggestion
RUN cargo test --all
```

Cargo.toml
weichweich: but we might not want to release right away.

Dockerfile
tjwelde: We should still use the release version, since it was already built in the previous step

PR #102, opened 2020-11-05
docs: runtime upgrade

tjwelde:
## fixes KILTProtocol/ticket#764
Just some notes about the runtime upgrade

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,9 @@ Substrate Documentation:
     - [Examples](#examples)
       - [Running a local node that connects to KILT prototype testnet in AWS](#running-a-local-node-that-connects-to-kilt-prototype-testnet-in-aws)
       - [Running a node with local image, which runs a dev-chain](#running-a-node-with-local-image-which-runs-a-dev-chain)
-  - [Development with AWS images](#development-with-aws-images)
+  - [Development](#development)
+    - [Devnet runtime upgrade](#devnet-runtime-upgrade)
+    - [Development with AWS images](#development-with-aws-images)
   - [Node Modules functionalities](#node-modules-functionalities)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -237,8 +239,19 @@ run chain in dev mode locally
...
 
-## Development with AWS images
+### Devnet runtime upgrade
+After merging a PR to the develop branch, the devnet images will be built and deployed, while keeping the data of the nodes.
+If the runtime changed, you will have to do a runtime upgrade. Here is how you do that:
+
+1. Go to the ["GitHub Actions" section](https://github.com/KILTprotocol/mashnet-node/actions?query=workflow%3A%22Deploy+to+AWS+EKS%22), find the "Deploy to AWS EKS" workflow on the develop branch with your merged PR.
+2. In the `build` job, you can find the new runtime under "Artifacts" in the top right. Download it.
+3. Go to the chain-explorer and setup the sudo account (which is ALICE in the case of then devnet)
+4. In "Extrinsics" use `sudo.sudoUncheckedWeight` with a low weight (needed because runtime is too big for a normal block) and `system.setCodeWithoutChecks`
```
weichweich, line 251:
```suggestion
4. In "Extrinsics" use `sudo.sudoUncheckedWeight` with the maximum weight (should be 3000000000000, can be found in runtime/src/lib.rs, needed because runtime is too big for a normal block) and `system.setCodeWithoutChecks`
```

might be nitpicking, but AFAIK setting the block weight to the maximum ensures that the runtime upgrade is the only thing that gets included in the block. Since we don't have to many TXs at the moment that shouldn't be a problem...

weichweich (approved):

PR #103, opened 2020-11-12
feat: unit denomination

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/714

Updating the unit denomination and name
Changing the properties doesn't affect the genesis block.

## How to test:

Spun up two nodes, one with the changes and the other without. They were able to connect without conflicts.

## Checklist:

### DIFF
--- a/src/chain_spec.rs
+++ b/src/chain_spec.rs
@@ -23,7 +23,7 @@ use mashnet_node_runtime::{
 	WASM_BINARY,
 };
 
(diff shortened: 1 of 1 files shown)


Dudleyneedham:
Question is what value do we want to put the decimal place too.

weichweich (approved):

wischli (approved):

PR #105, opened 2020-12-03
feat: added unit denomination to chain spec properties

LeonFLK:
## fixes KILTProtocol/ticket#714

Adds dudleys changes to the chain unit properties to the actual spec files.

## Checklist:


wischli (approved):
LGTM

PR #109, opened 2021-02-01
Refactor Delegations

weichweich:
## Checklist:


pallets/delegation/src/lib.rs
```diff
@@ -323,25 +326,30 @@ impl<T: Trait> Module<T> {
 	pub fn is_delegating(
 		account: &T::AccountId,
 		delegation: &T::DelegationNodeId,
+		max_depth: u64,
 	) -> result::Result<bool, &'static str> {
+		if max_depth == 0 {
+			return Err("Reached max delegation search depth");
```
wischli, line 332: 1. This error is never checked in a test or am I missing something?
2. Do we want this style or rather open an else closure?

pallets/delegation/src/lib.rs
```diff
@@ -323,25 +326,30 @@ impl<T: Trait> Module<T> {
...
+		}
+
 		// check if delegation exists
 		let d = <error::Module<T>>::ok_or_deposit_err(
```
wischli, line 336: Still wondering whether we should give a more distinct name?

pallets/delegation/src/lib.rs
```diff
@@ -389,12 +397,80 @@ impl<T: Trait> Module<T> {
...
+			permissions,
+			revoked: false,
+		}
+	}
+}
+
+#[derive(Encode, Decode)]
+pub struct DelegationProperties<T: Trait> {
+	pub ctype_hash: T::Hash,
+	pub account: T::AccountId,
```
wischli, line 449: Maybe we should add a comment which displays what this field stands for? `account` is very undescriptive.

wischli (commented):
Three minor things we might want to improve. However, it can be merged as is.

pallets/delegation/src/lib.rs
weichweich: 1. I hid it well. 😄 https://github.com/KILTprotocol/mashnet-node/pull/109/files#diff-030cb87e9d69994cc367bbd6579734da355d6901ba5791f2c8997235a668c092R410-R411
2. In general I'm all for for doing one return in the end of the function and not do things like `if ... {return}`. For me, doing this in the very beginning of a function is an exception. IMHO this doesn't hurt readability like a `if ... {return}` in the middle of the function would do. When you start reading the function, it's very clear that it's a No-op in case `max_depth` is 0. Putting an else there would indent the rest of the function which I also don't like.

pallets/delegation/src/lib.rs
weichweich: `ok_or_deposit_err`? I think this will be touched when we refactor errors?

pallets/delegation/src/lib.rs
weichweich: yes. It's the owner right? Wasn't sure because the delegation root also has an owner.

wischli: I am not a 100% sure either, but that's what I assumed.

wischli: Checked the whitepaper to be sure: It is indeed the owner.

pallets/delegation/src/lib.rs
weichweich: ahhhhh you meant the `d`... fixed in 951c5da

pallets/delegation/src/lib.rs
weichweich: fixed in 951c5da

wischli (approved):
LGTM!

PR #110, opened 2021-02-03
Refactor Attestations

wischli:
## fixes KILTProtocol/ticket#943
* Adds custom type `Attestation` to `Attestations` storage instead of a tuple
* Fixes clippy warnings and fmt warnings
* Adds clippy, fmt check to CLI
* Adds checking & testing of all features to CLI

## How to test:
```
```

## Checklist:


wischli:
Changing to compare with develop to check CLI.

pallets/attestation/src/lib.rs
```diff
@@ -130,15 +128,15 @@ decl_module! {
...
 			)?;
 			// if the sender of the revocation transaction is not the attester, check delegation tree
-			if !existing_attestation.1.eq(&sender) {
-				match existing_attestation.2 {
-					Some(d) => {
-						if !<delegation::Module<T>>::is_delegating(&sender, &d, max_depth)? {
+			if !owner.eq(&sender) {
+				match delegation_id {
+					Some(delegation_id) => {
+						if !<delegation::Module<T>>::is_delegating(&sender, &delegation_id, max_depth)? {
```
weichweich, line 139: maybe we could use an `if let some` here to reduce the indentation? Another option would be to do

```rust
let del_id = delegation_id.ok_or_else(ERROR)?;

if !is_delegating(...) {
    ERROR
}
```

I'm a fan of the later because the errors occur in code in the order they are checked

weichweich (commented):
maybe we could replace the 'match on an option' while we are at a style refactor? 🙂

pallets/attestation/src/lib.rs
wischli: This will be part of the error removal ticket #940!

weichweich: Ok than let's meeeeeerge! 😁

weichweich (approved):

pallets/attestation/src/lib.rs
```diff
@@ -191,10 +193,22 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+#[derive(Encode, Decode)]
+pub struct Attestation<T: Trait> {
+	// hash of the CTYPE used for this attestation
+	ctype_hash: T::Hash,
+	// the account which executed the attestation
+	owner: T::AccountId,
```
weichweich, line 201: I just noticed in the sdk that owner is very misleading. Who owns an attestation? That's a philosophical question that we all might answer differently. The owner in our case is the attester?
 
```suggestion
	// the account which executed the attestation
	attester: T::AccountId,
```

wischli: That is true. Will change.

wischli: Fixed in 9b0fe51

PR #111, opened 2021-02-03
fix readme after merge

weichweich:
updated how to build the standalone node

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ You can build it by executing these commands:
 #### Building in dev mode
 
 ```
-cargo build
+cargo build -p mashnet-node
 ```
 
 start, by running:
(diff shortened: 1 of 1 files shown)


wischli:
Do we want to add how to build and run a collator in a separate PR? I would be interested in that. I suppose its just
```
cargo build --release -p kilt-parachain
``` 
?

weichweich:
> Do we want to add how to build and run a collator in a separate PR? I would be interested in that. I suppose its just
> 
> ```
> cargo build --release -p kilt-parachain
> ```
> 
> ?

Yes i would do that in a separate PR. It's not that easy to run parachain. I think if we drop how to launch a parachain there it will only confuse people.

wischli (approved):

PR #113, opened 2021-02-05
Chain-Spec included

weichweich:
## fixes KILTProtocol/ticket#1052

- added a script for updating the spec
- chain spec is now included in the executable
- one dockerfile for standalone and parachain
  - the project is only compiled once after which the binaries just need to be copied
  - saves us 11 min on the rust server 😬 

@tjwelde i removed the `start-node.sh` hope that's fine? The boot nodes and telemetry will be added to the next chain spec we generate. Starting a testnet node would then just be `sudo docker run -it mashnet ./target/release/mashnet-node --chain kilt-testnet`.


wischli (approved):
Loving it!

wischli:
Maybe we should add something "how to update chain spec" to the project readme and not keep it secret in the python script?

weichweich:
> Maybe we should add something "how to update chain spec" to the project readme and not keep it secret in the python script?

You are right. We need to update the README anyways because i removed the start-node.sh. I will do that as part of this PR.

weichweich:
@wischli I updated the README and merged the dockerfile, i will wait until @tjwelde had a look because of the start-node.sh

weichweich:
After playing around with the setup i noticed that we now have it worse than it was in the beginning! Since the chain spec is in side the binary we have to recompile if the spec changes! 😢

SO i moved all the specs that might change our of the binary and kept specs that won't change inside the binary. This way frequent changes to the dev specs are possible without recompiling.

scripts/update-spec.py
```diff
@@ -0,0 +1,224 @@
...
+import json
+import uuid
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def build_spec(node, chain, parachain_id=None, raw=False):
+    # cmd = ["cargo", "run", "--release", "-p",
```
wischli, line 28: Is this comment kept intentionally?

wischli (commented):
We might want to add to the readme how to use the python script for our testnet?

weichweich:
> We might want to add to the readme how to use the python script for our testnet?

we only use the script if we reset a chain and start is from block 0. this will only happen for our internal chains. I would try to make a split between internal documentation and external documentation.

README.md
```diff
@@ -51,13 +51,13 @@ Substrate Documentation:
 Start chain and connect to alice bootnode:
 
 ```
-docker run -p 9944:9944 kiltprotocol/mashnet-node ./start-node.sh --connect-to alice
+docker run -p 9944:9944 kiltprotocol/mashnet-node --chain kilt-testnet --bootnodes /dns4/bootnode-alice.kilt-prototype.tk/tcp/30333/p2p/12D3KooWPuXafPUY8E8zo7m4GuWkgj9ByfsanrNUznZShBgJrW4A
```
tjwelde, line 54: Can we add the bootnodes into the chainspec, so that the user does not have to specify them himself?

README.md
```diff
@@ -273,7 +167,7 @@ docker pull 348099934012.dkr.ecr.eu-central-1.amazonaws.com/kilt/prototype-chain
 To run a node and connect it to the KILT testnet: Run the image and pass the command to start a node:
 
 ```
-docker run 348099934012.dkr.ecr.eu-central-1.amazonaws.com/kilt/prototype-chain ./start-node.sh --connect-to alice
+docker run 348099934012.dkr.ecr.eu-central-1.amazonaws.com/kilt/prototype-chain --chain kilt-testnet --bootnodes /dns4/bootnode-alice.kilt-prototype.tk/tcp/30333/p2p/12D3KooWPuXafPUY8E8zo7m4GuWkgj9ByfsanrNUznZShBgJrW4A
```
tjwelde, line 170: Same as above

tjwelde (changes requested):

README.md
weichweich: fixed in 76700c1

README.md
weichweich: fixed in 76700c1

tjwelde (approved):
LGTM

PR #114, opened 2021-02-08
bound on recursion

weichweich:
## Checklist:


pallets/delegation/src/lib.rs
```diff
@@ -329,35 +332,57 @@ impl<T: Trait> Module<T> {
...
 		// check if already revoked
 		if !delegation_node.revoked {
-			// set revoked flag and store delegation node
-			delegation_node.revoked = true;
-			<Delegations<T>>::insert(*delegation, delegation_node);
-			// deposit event that the delegation has been revoked
-			Self::deposit_event(RawEvent::DelegationRevoked(sender.clone(), *delegation));
-			// revoke all children recursively
-			Self::revoke_children(delegation, sender)?;
+			// fist revoke all children recursively
```
wischli, line 347: Minor typo here
```suggestion
			// first revoke all children recursively
```

pallets/delegation/src/lib.rs
```diff
@@ -329,35 +332,57 @@ impl<T: Trait> Module<T> {
...
+			revocations += Self::revoke_children(delegation, sender, max_revocations)?;
+
+			// if we run out of revocation gas, we only revoke children. The tree will be changed but still valid.
```
wischli, line 350: Missing a word here
```suggestion
			// if we run out of revocation gas, we only revoke children. The tree will be changed but is still valid.
```

pallets/delegation/src/lib.rs
```diff
@@ -130,6 +130,7 @@ decl_error! {
 		ParentNotFound,
 		UnauthorizedRevocation,
 		UnauthorizedDelegation,
+		ExeededBounds,
```
wischli, line 133: Typo here. I'd also suggest to get a more descriptive name like
```suggestion
		ExceededRevocationBounds,
```

wischli (changes requested):
Three typos and I think we should also add a unit test for `ExceedRevocationBounds`.

pallets/delegation/src/lib.rs
weichweich: fixed in 4966e4b

weichweich:
> Three typos and I think we should also add a unit test for `ExceedRevocationBounds`.

good catch! Totally forgot that! There was even a bug there! The root was always revoked even if there where no revocations remaining, leading to an invalid delegation tree! 😨 Fixed in 4966e4b...

pallets/delegation/src/tests.rs
```diff
@@ -442,17 +445,26 @@ fn check_add_and_revoke_delegations() {
...
 			),
 			Error::<Test>::RootNotFound
 		);
 		assert_noop!(
-			Delegation::revoke_root(Origin::signed(account_hash_bob), id_level_0),
-			Error::<Test>::UnauthorizedRevocation
+			Delegation::revoke_root(Origin::signed(account_hash_bob), id_level_0, 1),
+			Error::<Test>::UnauthorizedRevocation,
+		);
+		let res =
```
wischli, line 457: Why not the usual check?

```suggestion
		assert_noop!(Delegation::revoke_root(Origin::signed(account_hash_alice.clone()), id_level_0, 0), Error::<Test>::ExceededRevocationBounds);
```

weichweich: forgot that this exists. f8f7e8c

wischli (approved):
LGTM!

PR #115, opened 2021-02-09
chore: add benchmarks

wischli:
## Adds benchmarks

fixes KILTProtocol/ticket#964
fixes KILTProtocol/ticket#966
fixes KILTProtocol/ticket#965
fixes KILTProtocol/ticket#967

## How to test:

## Checklist:


pallets/delegation/src/benchmarking.rs
```diff
@@ -0,0 +1,373 @@
...
+		// only put in a leaf in the the first iteration
+		leaf = leaf.or(Some((
+			delegation_acc_public,
+			delegation_acc_id,
+			delegation_id,
+		)));
+	}
+	// if we didn't add children, return the parent
+	let (leaf_acc_public, leaf_acc_id, leaf_id) =
+		leaf.unwrap_or((parent_acc_public, parent_acc_id, parent_id));
```
weichweich, line 133: This should never happen and we probably should make sure it doesn't. If we have level = 10 and `children_per_level` = 0 the delegation tree will have no levels at all...

maybe `children_per_level` should be a [NonZeroInt](https://doc.rust-lang.org/std/num/struct.NonZeroU64.html)? But that's maybe to fancy...

weichweich: I know i wrote that. 😅

pallets/delegation/src/lib.rs
```diff
@@ -334,58 +352,69 @@ impl<T: Trait> Module<T> {
...
+			consumed_weight += T::DbWeight::get().reads(1);
+
 			for child in children {
 				let remaining_revocations = max_revocations.saturating_sub(revocations);
 				if remaining_revocations > 0 {
-					revocations += Self::revoke(&child, sender, remaining_revocations)?;
+					Self::revoke(&child, sender, remaining_revocations).map(|(r, w)| {
+						revocations += r;
+						consumed_weight += w;
+					})?;
```
weichweich, line 411: wouldn't that work too? 😁 But the map is also a very cool solution.

```suggestion
					let (r, w) = Self::revoke(&child, sender, remaining_revocations)?;
					revocations += r;
					consumed_weight += w;
```

runtimes/parachain/Cargo.toml
```diff
@@ -16,10 +16,10 @@ sp-arithmetic = {git = 'https://github.com/paritytech/substrate', default-featur
...
 
 # KILT pallets & primitives
-attestation = {default-features = false, path = "../../pallets/attestation"}
-ctype = {default-features = false, path = "../../pallets/ctype"}
-delegation = {default-features = false, path = "../../pallets/delegation"}
-did = {default-features = false, path = "../../pallets/did"}
+attestation = {default-features = false, path = '../../pallets/attestation'}
+ctype = {default-features = false, path = '../../pallets/ctype'}
+delegation = {default-features = false, path = '../../pallets/delegation'}
+did = {default-features = false, path = '../../pallets/did'}
```
weichweich, line 22: why single tics?

runtimes/standalone/Cargo.toml
```diff
@@ -208,6 +213,21 @@ version = '3.0.0'
 
 [features]
 default = ['std']
+runtime-benchmarks = [
+  'frame-benchmarking',
+  'frame-system-benchmarking',
+  'hex-literal',
```
weichweich, line 219: we should replace them everywhere

runtimes/standalone/src/lib.rs
```diff
@@ -400,29 +404,29 @@ construct_runtime!(
 		NodeBlock = opaque::Block,
 		UncheckedExtrinsic = UncheckedExtrinsic
 	{
-		System: frame_system::{Module, Call, Config, Storage, Event<T>},
-		RandomnessCollectiveFlip: randomness_collective_flip::{Module, Call, Storage},
+		System: frame_system::{Module, Call, Config, Storage, Event<T>} = 0,
+		RandomnessCollectiveFlip: randomness_collective_flip::{Module, Call, Storage} = 1,
```
weichweich, line 408: Is there a reason for the explicit indices?

runtimes/standalone/src/weights/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod attestation;
+pub mod ctype;
+pub mod delegation;
+pub mod did;
```
weichweich, line 4: this is probably a leftover? But we should have separate `WeightInfo` implementations for our runtimes?

weichweich (changes requested):
`runtimes/parachain/src/weights/mod.rs` Looks like it's not needed.

Some tomls still use `'` while others use `"`.

I think we need to focus on a single tasks in our PRs. This change of toml format added a lot of noise. I know I started that. But I probably shouldn't have... Also there is an dependency update in this PR which should have been it's own PR.

runtimes/standalone/src/lib.rs
wischli: I saw this in the [Polkadot runtimes](https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L983) and liked it because we are actually using the indices for the extrinsic errors in the SDK. It. should have definitely been a separate PR so I am happy to revert it and propose it later.

runtimes/standalone/Cargo.toml
wischli: Fixed in e421d63

runtimes/parachain/Cargo.toml
wischli: Fixed in e421d63

pallets/delegation/src/lib.rs
wischli: That would work too. I would like to keep the map though, no need to instantiate two new vars right?

wischli:
> I think we need to focus on a single tasks in our PRs. This change of toml format added a lot of noise. I know I started that. But I probably shouldn't have... Also there is an dependency update in this PR which should have been it's own PR.

Totally agree. We both have to stick to this in the future 🥲

runtimes/standalone/src/lib.rs
weichweich: na let's keep it. I reviewed it already! 😁 Another PR means that I have to review it AGAIN!!! 😨

runtimes/standalone/src/weights/mod.rs
wischli: Yes, this is a leftover. I suppose eventually we would need this for our separate `WeightInfo` implementations as you said. I could append this to this PR?!

pallets/delegation/src/benchmarking.rs
wischli: I actually went with that so it's more clear. Unfortunately, the constructor wraps it into an option. See e91528a + f0679d1

runtimes/standalone/src/weights/mod.rs
weichweich: > Yes, this is a leftover. I suppose eventually we would need this for our separate `WeightInfo` implementations as you said. I could append this to this PR?!

let's do that in a separate PR

weichweich: When we do the "real" benchmarks. We don't want to run the benchmarks on our Workstation

wischli: Fixed in 42820cd

wischli:
The current CI failure will be fixed in a separate PR.

weichweich (approved):
LGTM

PR #118, opened 2021-02-24
refactor: convert did storage return type

wischli:
## fixes KILTProtocol/ticket#1116
* converts the `DIDs` storage return type from tuple to distinct struct `Did`
* fixes clippy warnings which broke the CI
* breaks SDK --> have to merge [this related PR](https://github.com/KILTprotocol/sdk-js/pull/348)
## How to test:
```
```
## Checklist:


runtimes/parachain/src/lib.rs
```diff
@@ -278,6 +278,7 @@ impl did::Config for Runtime {
 	type WeightInfo = ();
 }
 
+#[allow(clippy::from_over_into)]
```
weichweich, line 281: is this doing something? Otherwise i would remove it?

pallets/did/src/tests.rs
```diff
@@ -118,7 +119,7 @@ impl Config for Test {
 	type WeightInfo = ();
 }
 
-type DID = Module<Test>;
+type DidModule = Module<Test>;
```
weichweich, line 122: The `DID` Module should come out of the `construct_runtime!`-Thingy. THis should go and get replaced with `Did`? Didn't try that tho.

pallets/did/src/tests.rs
```diff
@@ -134,24 +135,24 @@ fn check_add_did() {
 		let signing_key = H256::from_low_u64_be(1);
 		let box_key = H256::from_low_u64_be(2);
 		let account = MultiSigner::from(pair.public()).into_account();
-		assert_ok!(DID::add(
+		assert_ok!(DidModule::add(
```
weichweich, line 138: That should work?

```suggestion
		assert_ok!(Did::add(
```

pallets/did/src/lib.rs
```diff
@@ -100,9 +101,19 @@ decl_module! {
 	}
 }
 
+#[derive(Encode, Decode)]
+pub struct Did<T: Config> {
```
weichweich, line 105: Ah now I see. The stuff i suggest below won't work. This thing is not the DID but something different. I think it's the did document https://www.w3.org/TR/did-core/#dfn-did-documents

```suggestion
pub struct DidDocument<T: Config> {
```

weichweich (changes requested):
only a few minor things.

- Do the `#[allow(clippy::from_over_into)]` work? otherwise i would remove them.
- The line `type DidModule = Module<Test>;` should not be necessary. This should come out of `construct_runtime`
- I would maybe call the `Did` Struct `DidDocument`. Maybe consult Raphael?

runtimes/parachain/src/lib.rs
wischli: That is right, I hoped it would solve the clippy issue, but I will remove it.

.github/workflows/tests.yml
```diff
@@ -34,23 +33,74 @@ jobs:
...
+        toolchain: nightly-2021-02-22
         target: wasm32-unknown-unknown
         override: true
         default: true
 
-    # Enable this for clippy linting.
-    # - name: Check and Lint Code
-    #   run: cargo +nightly-2021-02-14 clippy -- -D warnings
     - name: Check Code
       run: cargo check --all --all-features
```
weichweich, line 42: maybe we can just run the tests? Could be faster? Not sure but maybe worth a try?

wischli: Definitely, fixed in fd30b5a

runtimes/parachain/src/lib.rs
wischli: Fixed in 51333a7

pallets/did/src/lib.rs
wischli: Oh thanks, I was unhappy with the naming. Fixed in 79bd337

weichweich: Ähm I just talked to raphael and it's not the document but more like the `DidRecord`. Sorry 😭 !!

weichweich: There is no 'DidRecord' AFAIK but that name is better suited...

weichweich (approved):
LGTM

PR #123, opened 2021-03-04
chore: adjusted max_depth parameter in revoke_delegation

rflechtner:
## fixes KILTProtocol/ticket#1139
Adjusts the max_depth parameter in revoke_delegation so that it is equal to the number of levels you need to walk up to find the node owned by the sender of the tx.

**Examples:**
- when sender is owner of the current node, `max_depth=0`
- when sender is owner of the parent node, `max_depth=1`

Also renamed the param in is_delegating to `max_lookups`, as that's what it counts - not depth, which may have led to the issue in the first place.

## How to test:
?

## Checklist:

### DIFF
--- a/pallets/delegation/src/lib.rs
+++ b/pallets/delegation/src/lib.rs
@@ -284,7 +284,7 @@ decl_module! {
 		/// Revoke a delegation node and all its children, where
 		/// origin - the origin of the transaction
 		/// delegation_id - id of the delegation node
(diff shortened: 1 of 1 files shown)


rflechtner:
probs not the most elegant solution, but that should work. If you have a suggestion on how to improve it, tell me - but let's make it quick.

weichweich (approved):
LGTM! 😄

weichweich:
@rflechtner auto merge enabled! 😁

PR #124, opened 2021-03-05
xtoken

weichweich:
## Checklist:


wischli (approved):
LGTM!

As discussed, let's not merge this PR until we've updated xtokens, Substrate and Cumulus after which we hopefully can replace our local xtokens pallet with the original one.

wischli (approved):
LGTM!

PR #128, opened 2021-03-22
Add license check

weichweich:
## No Ticket

- check that license is in all rust files
- wrap comments after 120 columns

## Checklist:


pallets/delegation/src/tests.rs
```diff
@@ -1,4 +1,4 @@
-// KILT Blockchain – https://botlabs.o index: (), error: (), message: ()rg
+// KILT Blockchain – https://botlabs.org
```
weichweich, line 1: 😂

wischli (approved):
Love it!

PR #130, opened 2021-03-23
feat: add democracy

weichweich:
## fixes KILTProtocol/ticket#963

- [x] add rpc for querying tx info
- [x] fix the origin problem
- [x] fix the type problem

```
RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1002: Verification Error: Runtime error: Execution failed: 
ApiError(FailedToConvertParameter {
    function: "validate_transaction",
    parameter: "tx",
    error: Error {
        cause: Some(Error {
            cause: Some(Error {
                cause: Some(Error {
                    cause: Some(Error {
                        cause: None,
                        desc: "Not enough data to fill buffer"
                    }), 
                    desc: "Could not decode `Call::fast_track.2`"
                }),
                desc: "Could not decode `Call::Democracy.0`"
            }),
            desc: "Could not decode `Call::propose.1`"
        }),
        desc: "Could not decode `Call::TechnicalCommittee.0`"
    } 
})
```

- [x] change back duration of things

## Changed JS types

delete:

- ReferendumInfo

added:

I think we need to add `AmountOf`.

```json
{
  "BlockNumber": "u64",
  "AmountOf": "i128"
}
```

## Checklist:


wischli (approved):
Looks good to me and the council!

PR #133, opened 2021-03-29
feat: add support for DID updates

Diiaablo95:
## fixes KILTProtocol/ticket#1196

Adds support for DID updates.

### With cargo
Run the tests with `cargo test`.

> Not possible at the moment to test using Polkadot-js apps web UI as `BTreeSet` inputs are not interpreted correctly by the web interface.

## Checklist:


pallets/did/src/lib.rs
```diff
@@ -184,6 +186,12 @@ pub enum StorageError {
 	/// The given DID does not contain the right key to verify the signature of
 	/// a DID operation.
 	DidKeyNotPresent(DidVerificationKeyType),
+	/// One or more verification keys referenced are not stored in the set of
+	/// verification keys.
+	VerificationKeyNotPresent(PublicVerificationKey),
+	/// Duplicate key specified in the vector of keys to delete from the
+	/// verification keys
+	DuplicateVerificationKey(PublicVerificationKey),
```
Diiaablo95, line 194: Since I did not get a `BTreeMap` to work properly as input parameter inside an update operation, I had to switch to lists. And so we need another error in case the user is trying to delete the same key twice in the same operation.

pallets/did/src/lib.rs
```diff
@@ -240,6 +255,43 @@ where
...
+pub struct DidUpdateOperation<DidIdentifier>
+where
+	DidIdentifier: Parameter + Encode + Decode + Debug,
+{
+	did: DidIdentifier,
+	new_auth_key: Option<PublicVerificationKey>,
+	new_key_agreement_key: Option<PublicEncryptionKey>,
+	new_attestation_key: Option<PublicVerificationKey>,
+	new_delegation_key: Option<PublicVerificationKey>,
+	verification_keys_to_remove: Option<Vec<PublicVerificationKey>>,
```
Diiaablo95, line 277: Same as above. Initially this was an `Option<BTreeSet<PublicVerificationKey>>`, but polkadot-js was only allowing me to input it as a HEX-encoded byte array, so there had to be something wrong in the types. Tests were running ok, and I would to revert to using a `BTreeSet` tho.

Diiaablo95:
Apparently support for `BTreeSet` was added in https://github.com/polkadot-js/api/pull/1639, but apparently there is no support for it in apps (https://github.com/polkadot-js/apps/issues/2059).

weichweich (approved):
LGTM just a few style things and I would prefer the `BTreeSet`. 😇

pallets/did/src/lib.rs
```diff
@@ -292,6 +344,92 @@ impl DidDetails {
...
+		// Start from all the current verification keys...
+		let mut remaining_verification_keys = new_details.verification_keys;
+
+		// If there are some keys to remove...
+		if let Some(verification_keys_to_remove) = update_operation.verification_keys_to_remove {
+			// Keep track of the keys already seen so far (as we are using a list instead of
+			// a set for the keys)
+			let mut seen_keys = BTreeSet::<PublicVerificationKey>::new();
+
+			verification_keys_to_remove.iter().try_for_each(|key| {
```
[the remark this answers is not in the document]
Diiaablo95: I would also prefer to use the `BTreeSet`! So I guess I'll just revert back to that version, if you also agree that is a better way to go.

Regarding the `for .. in` stuff, I used the `try_for_each` to allow for exit returns in case of errors. Is that also possible with the classical `foreach` construct?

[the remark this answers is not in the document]
weichweich: Yes that's possible. It's a bit overloaded. [If you think of an if-block that returns a value](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=04931d20ad8f0c641ae893131f7b1c54):

The if block returns something into x and can also bail early and return something for the whole function. That's a bit strange.

pallets/did/src/lib.rs
```diff
@@ -433,16 +629,38 @@ impl<T: Config> Pallet<T> {
 		let did_entry: DidDetails =
 			<Did<T>>::get(op.get_did()).ok_or(DidError::StorageError(StorageError::DidNotPresent))?;
 
+		Self::verify_operation_signature_for_entry(op, signature, &did_entry)?;
+
+		Ok(())
+	}
+
+	/// Verify the signature of a generic DidOperation, and
```
wischli, line 629: The first sentence of this comment seems off. I would remove the part about the return.

```suggestion
	/// Verify the signature of a generic DidOperation.
```

pallets/did/src/tests.rs
```diff
@@ -210,16 +214,313 @@ fn check_invalid_signature_did_creation() {
...
+
+	ext.execute_with(|| {
+		assert_ok!(Did::submit_did_update_operation(
+			Origin::signed(DEFAULT_ACCOUNT),
+			did_update_operation.clone(),
+			did::DidSignature::from(signature),
+		));
+	});
+	let new_did_details = ext.execute_with(|| Did::get_did(ALICE_DID).expect("ALICE_DID should be present on chain."));
+	assert_eq!(new_did_details.auth_key, did_update_operation.new_auth_key.unwrap());
```
wischli, line 259:
```suggestion
	assert_eq!(new_did_details.auth_key, did_update_operation.new_auth_key.expect("Missing new auth key"));
```

pallets/did/src/tests.rs
```diff
@@ -210,16 +214,313 @@ fn check_invalid_signature_did_creation() {
...
+	assert_eq!(
+		new_did_details.key_agreement_key,
+		did_update_operation.new_key_agreement_key.unwrap()
```
wischli, line 262:
```suggestion
		did_update_operation.new_key_agreement_key.expect("Missing new key agreement key")
```

wischli (changes requested):
Some minor remarks.

pallets/did/src/lib.rs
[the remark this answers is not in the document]
Diiaablo95: Reverting back to sets solves the issue altogether since we use set operations 👍

Diiaablo95:
@weichweich @wischli please take another look when you can. I reverted back to accepting a `BTreeSet` as input parameter for the set of verification keys to delete. Makes it a bit more cumbersome to test as it is not understood correctly by the apps web UI, but it should not cause any issues with the polkadot api types.

pallets/did/src/lib.rs
```diff
@@ -292,6 +343,74 @@ impl DidDetails {
 	}
 }
 
+/// Generates a new DID entry starting from the current one stored in the
+/// storage and by applying the changes in the DIDUpdateOperation. The operation
+/// fails with a DidError if the update operation instructs to delete a
+/// verification key that is not associated with the DID or if the operation
+/// counter is not larger than the one stored on chain. !!! Please note that
```
wischli, line 350: Generally I would like to move notes like this to a new line and not use exclamation marks as suggested previously. This is handled similarly in Substrate.

```suggestion
/// counter is not larger than the one stored on chain. 
/// 
/// Please note that
```

wischli (approved):
LGTM!

pallets/did/src/lib.rs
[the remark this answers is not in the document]
wischli: > Yes that's possible. It's a bit overloaded. [If you think of an if-block that returns a value](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=04931d20ad8f0c641ae893131f7b1c54):
> 
> The if block returns something into x and can also bail early and return something for the whole function. That's a bit strange.
> 
> Also: [exit on error in a for loop](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e20888ea698133435acfd54a327b4b07).

I was dealing with a similar issue recently and was also wondering whether `for ... in` was the way to go. I tried to get `map` working but failed.

PR #136, opened 2021-04-06
feat: add support for DID deletion

Diiaablo95:
## fixes KILTProtocol/ticket#1182

Add support to delete all the information associated with a DID from the chain.

### With Cargo

Run `cargo test`.

### With Polkadot-JS apps

1. Follow the steps in https://github.com/KILTprotocol/mashnet-node/pull/127 to create a new test identity and test DID
2. Open the UI to invoke a `submitDidDeleteOperation` extrinsic
3. `did: DIDIdentifier` -> select the test account created
4. `tx_counter: U64` -> insert `1`
5. `signature: DidSignature` -> select `Ed25519` and value `0x1901e750f8fb18486858ac59136e6448974afb1b3da4adb184e7453aa2fe4352f210a1e57a12c536d88ccbd0b0569c9cbf99fb6d4623e1071c0fb98087d6dc07` (this is the signature over the encoded delete operation)

Now, if you try to re-execute the same operation, the extrinsic will fail with a `did.DidNotPresent` error.

## Checklist:

### DIFF
--- a/pallets/did/src/lib.rs
+++ b/pallets/did/src/lib.rs
@@ -291,6 +291,31 @@ where
 	}
 }
 
+/// A DID deletion request. It contains the following values:
+/// * The DID identifier being deleted
+/// * A counter used to protect against replay attacks
+#[derive(Clone, Decode, Debug, Encode, PartialEq)]
+pub struct DidDeletionOperation<DidIdentifier>
+where
+	DidIdentifier: Parameter + Encode + Decode + Debug,
+{
+	did: DidIdentifier,
+	tx_counter: u64,
+}
+
+impl<DidIdentifier> DidOperation<DidIdentifier> for DidDeletionOperation<DidIdentifier>
+where
+	DidIdentifier: Parameter + Encode + Decode + Debug,
+{
+	fn get_verification_key_type(&self) -> DidVerificationKeyType {
+		DidVerificationKeyType::Authentication
+	}
+
+	fn get_did(&self) -> &DidIdentifier {
+		&self.did
+	}
+}
+
 /// The details associated to a DID identity. Specifically:
 /// * The authentication key, used to authenticate DID-related operations
 /// * The key agreement key, used to encrypt data addressed to the DID subject
@@ -302,7 +327,7 @@ where
 ///   exposes
 /// * A counter used to avoid replay attacks, which is checked and updated upon
 ///   each DID-related operation
-#[derive(Clone, Decode, Encode, PartialEq)]
+#[derive(Clone, Debug, Decode, Encode, PartialEq)]
 pub struct DidDetails {
 	auth_key: PublicVerificationKey,
 	key_agreement_key: PublicEncryptionKey,
@@ -378,7 +403,7 @@ where
 			);
 		};
 
-		// Verify that the operation counter is greater than the stored
+		// Verify that the operation counter is greater than the stored one
 		ensure!(
 			update_operation.tx_counter > new_details.last_tx_counter,
 			DidError::OperationError(OperationError::InvalidNonce)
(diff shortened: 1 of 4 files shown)


weichweich:
The commit history is a little bit messy 😂 😅

wischli:
> The commit history is a little bit messy 😂 😅

Yeah, I was also about to request a rebase to have 2a5e8a3 as the first commit of this PR.

pallets/did/src/lib.rs
```diff
@@ -291,6 +291,31 @@ where
 	}
 }
 
+/// A DID deletion request. It contains the following values:
+/// * The DID identifier being deleted
+/// As replaying a deletion request does not change the state of the system,
+/// the deletion request does not have a tx counter.
```
weichweich, line 297: Can I recreate a did with the same identifier?

pallets/did/src/lib.rs
```diff
@@ -592,6 +618,41 @@ pub mod pallet {
...
+
+			// Verify the signature of the delete operation.
+			Self::verify_operation_signature_for_entry(&did_deletion_operation, &signature, &did_details)
+				.map_err(<Error<T>>::from)?;
+
+			log::debug!("Deleting DID {:?}", did_identifier);
+			<Did<T>>::take(&did_identifier);
+
+			Self::deposit_event(Event::DidDeleted(sender, did_identifier.to_owned()));
+			//TODO: Return the real weight used
```
weichweich, line 653: No need to adjust the weight after execution since the number of read/writes is constant.

pallets/did/src/tests.rs
```diff
@@ -483,6 +483,113 @@ fn check_equal_tx_counter_did_update() {
 	});
 }
 
+// submit_did_delete_operation
+
+#[test]
+fn check_successful_deletion() {
```
weichweich, line 489: maybe a test for adding a deleted DID would be nice?

weichweich (commented):
Looks good. I have two questions tho.

pallets/did/src/lib.rs
Diiaablo95: Potentially, yes. But it will likely have a different set of keys.

pallets/did/src/lib.rs
Diiaablo95: Does this apply also to the other extrinsics in the DID pallet?

pallets/did/src/lib.rs
weichweich: If I accidentally deleted my DID and recreate it with the same keys because I'm lazy, everyone can delete it again?

If it's possible, I would add a nonce to prevent it. Just to make it waterproof.

pallets/did/src/lib.rs
weichweich: yes. You only need to adjust the weight if you have DB read/writes in a if/foreach-branch.

pallets/did/src/lib.rs
Diiaablo95: In this case the counter will be reset to 0, and anyone could still re-delete the DID. To prevent that, we should allow users to specify a custom tx counter upon DID creation, or otherwise we could not completely delete the DID from the chain but just mark it as unusable, but then we are adding more complex logic there.

weichweich: He could do a batch TX that contains a DID-Create and DID-Update operation. In the update operation he then can update the nonce to a custom value. I think this is fine. 👍🏻

Diiaablo95: Yeah, I guess having a tx counter makes sense then ✌🏻

Diiaablo95:
Rebasing seems to give many issues. I'll open a new PR with a new branch.

PR #139, opened 2021-04-12
feat: add basic checks for encoded URL in DID creations and updates

Diiaablo95:
## fixes KILTProtocol/ticket#1128

A new enum `Url` for the different types of URLs supported has been added. Each variant of the enum has an associated Struct value of type `HttpUrl`, `FttpUrl`, or `IpfsUrl`. Each of these structs has logic to validate an input URL encoded as a `Vec<u8>`.

At the moment, the checks are basic. For all of the different URLs, it is checked whether they start with the correct scheme, i.e., `http://` or `https://` for HTTP, `ftp://` or `ftps://` for FTP, and `ipfs://` for IPFS. Furthermore, for HTTP and FTP URLs, it is also checked whether all characters are valid URL ASCII characters according to [RFC 3986](https://tools.ietf.org/html/rfc3986). For IPFS, it is checked whether the CID following the `ipfs://` scheme is either Base58-encoded (v0 of the protocol), or Base32-encoded (v1).

## How to test:

Run `cargo test --all --all-features`.

## Checklist:


Diiaablo95:
The alternative approach of using regex was not working properly and would "pollute" the codebase by adding a lot of pre-compiled byte arrays. So I guess this is a good compromise.

pallets/did/src/tests.rs
```diff
@@ -225,19 +230,22 @@ fn check_successful_complete_update() {
...
-	let new_url: UrlEncoding = "https://new_kilt.io".into();
+	let new_url = did::Url::from(
+		did::HttpUrl::try_from("https://new_kilt.io".as_bytes())
+			.expect("https://new_kilt.io should not be considered an invalid HTTP URL."),
+	);
 
 	let mut old_did_details =
 		generate_mock_did_details(did::PublicVerificationKey::from(old_auth_key.public()), old_enc_key);
-	old_did_details.attestation_key = Some(PublicVerificationKey::from(old_att_key.public()));
+	old_did_details.attestation_key = Some(did::PublicVerificationKey::from(old_att_key.public()));
```
wischli, line 240: Is there any reason you removed `PublicVerificationKey` from the imports?

pallets/did/src/utils.rs
```diff
@@ -0,0 +1,60 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+/// Verifies that an input string contains only URL-allowed ASCII characters.
+/// For more info about what those characters are, please visit the official RFC
+/// 3986.
+pub fn is_valid_ascii_url(input: &str) -> bool {
```
wischli, line 22: We should also add unit tests for these utility functions.

wischli (commented):
IMO we should add unit tests for the three utility functions. Apart from that, LGTM.

pallets/did/src/tests.rs
Diiaablo95: No. It was just a matter of consistency, as I was accessing the did pallet using the full path, and in some places I was using `PublicVerificationKey` after importing it directly.

pallets/did/src/utils.rs
Diiaablo95: Done that! ✌🏻

wischli (approved):
LGTM!

pallets/did/src/lib.rs
```diff
@@ -211,6 +211,15 @@ pub enum OperationError {
 	InvalidNonce,
 }
 
+// Error generated when invalid URLs are used as encoded URLs.
```
weichweich, line 214: Should probably be a doc comment?

```suggestion
/// Error generated when invalid URLs are used as encoded URLs.
```

pallets/did/src/lib.rs
```diff
@@ -316,6 +325,118 @@ where
...
+
+	// It fails if the URL is not ASCII-encoded or does not start with the expected
+	// URL scheme.
+	fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+		let str_url = str::from_utf8(value).map_err(|_| UrlError::InvalidUrlEncoding)?;
+
+		ensure!(str_url.starts_with("ipfs://"), UrlError::InvalidUrlScheme);
+
+		// Remove the first 7 characters (the URL scheme)
+		let slice_to_verify = &str_url[7..];
```
weichweich, line 399: This panics if `str_url` is to short. I would rather use [get](https://doc.rust-lang.org/std/primitive.slice.html#method.get) in combination with `.unwrap().expect("The length was ensured by calling starts_with")` or something similar.

pallets/did/src/utils.rs
```diff
@@ -0,0 +1,161 @@
...
+	for c in input.chars() {
+		// Matches [0-9], [a-z], [A-Z], plus the symbols as in the RFC.
```
weichweich, line 24: You could probably also use `input.chars().find(|c| !match!(c, ...)).is_none()`.

weichweich (changes requested):
I would like to have a more explicit handling of the panic. Looks good otherwise. 🙂

pallets/did/src/lib.rs
weichweich: And maybe even ... have `"ipfs://"` as a const and use the length of that const instead of 7? I think this would make it clear where the 7 comes from even for me! 😅

weichweich (approved):
LGTM! :+1:

pallets/did/src/utils.rs
```diff
@@ -0,0 +1,156 @@
...
+/// 3986.
+pub fn is_valid_ascii_url(input: &str) -> bool {
+	// Matches [0-9], [a-z], [A-Z], plus the symbols as in the RFC.
+	input
+		.chars()
+		.find(|c| {
+			!matches!(c, ':' | '/' | '?' | '#' | '[' | ']' | '@' | '!' | '$' | '&' | '\'' | '(' | ')' | '*' | '+' | ',' | ';'
+	| '=' | '-' | '.' | '_' | '~' | '%' | '0'..='9' | 'a'..='z' | 'A'..='Z')
+		})
+		.is_none()
```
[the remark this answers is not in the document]
Diiaablo95: They both short-circuit, but perhaps it would make it more readable to not have a negation in there. I agree 🙂

PR #141, opened 2021-04-15
feat: refactor remaining pallet to integrate the new DID pallet

Diiaablo95:
## fixes KILTProtocol/ticket#1205, KILTProtocol/ticket#1206 and KILTProtocol/ticket#1207

TBD.

## How to test:

Run `cargo test --all --all-features`.

## Checklist:


pallets/ctype/Cargo.toml
```diff
@@ -12,41 +12,59 @@ targets = ["x86_64-unknown-linux-gnu"]
 [build-dependencies.substrate-wasm-builder-runner]
```
Diiaablo95, line 12: The dependencies have just, for some reason, been ordered alphabetically, nothing more.

pallets/ctype/src/benchmarking.rs
```diff
@@ -16,36 +16,36 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
-#![cfg(feature = "runtime-benchmarks")]
```
Diiaablo95, line 19: Commented out this whole file.

pallets/ctype/src/lib.rs
```diff
@@ -26,73 +26,148 @@ pub mod benchmarking;
 #[cfg(test)]
```
Diiaablo95, line 26: Update to new FRAME version and update for new DID pallet.

pallets/did/Cargo.toml
```diff
@@ -9,28 +9,43 @@ version = "0.24.0"
 [package.metadata.docs.rs]
```
Diiaablo95, line 9: Same here. Mainly just re-ordered alphabetically.

Diiaablo95:
As CTYPE refactor is complete, I think it would be a nice time to have a first round of reviews, that is nevertheless not blocking the rest of the development @weichweich @wischli

wischli:
> As CTYPE refactor is complete, I think it would be a nice time to have a first round of reviews, that is nevertheless not blocking the rest of the development @weichweich @wischli

Can't review before tomorrow unfortunately.

pallets/ctype/Cargo.toml
weichweich: I like the sorting! 👍🏻

pallets/ctype/src/lib.rs
```diff
@@ -26,73 +26,148 @@ pub mod benchmarking;
...
+/// information about the creator's DID, the type of DID key required to
+/// verify the operation signature, and the tx counter to protect against replay
+/// attacks. The struct has the following fields:
+/// * creator_did: the DID of the CTYPE creator
+/// * hash: the CTYPE hash
+/// * tx_counter: the DID tx counter to mitigate replay attacks
+#[derive(Clone, Decode, Encode, PartialEq)]
+pub struct CtypeCreationOperation<T: Config> {
+	creator_did: <T as did::Config>::DidIdentifier,
+	hash: <T as frame_system::Config>::Hash,
```
weichweich, line 53: If you pass both of them separately you can derive debug. Makes it also more generic since the requirements are not as tight as requiring the whole `Config`.

```suggestion
pub struct CtypeCreationOperation<DidIdentifier, Hash> {
	creator_did: DidIdentifier,
	hash: Hash,
```

pallets/did/Cargo.toml
weichweich: and the mock feature ;)

pallets/did/src/mock.rs
```diff
@@ -81,7 +80,8 @@ impl did::Config for Test {
 
 pub type TestDidIdentifier = <Test as did::Config>::DidIdentifier;
 
-pub const DEFAULT_ACCOUNT: AccountId = AccountId::new([0u8; 32]);
+#[allow(dead_code)]
```
weichweich, line 83: this is only used in the tests? Then i would use

```suggestion
#[cfg(test)]
```

pallets/did/src/lib.rs
```diff
@@ -248,30 +248,44 @@ pub trait DidOperation<DidIdentifier>: Encode {
...
 /// * The optional delegation key to use
 /// * The optional endpoint URL pointing to the DID service endpoints
-#[derive(Clone, Decode, Debug, Encode, PartialEq)]
-pub struct DidCreationOperation<DidIdentifier>
-where
-	DidIdentifier: Parameter + Encode + Decode + Debug,
-{
-	did: DidIdentifier,
+#[derive(Clone, Decode, Encode, PartialEq)]
+pub struct DidCreationOperation<T: Config> {
```
weichweich, line 252: i would rather keep the `DidIdentifier` but I'm open to talk about it. 🙂

pallets/did/src/mock.rs
```diff
@@ -239,7 +242,7 @@ impl ExtBuilder {
 		let storage = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
 		let mut ext = sp_io::TestExternalities::new(storage);
 
-		if self.dids_stored.len() > 0 {
+		if !self.dids_stored.is_empty() {
```
weichweich, line 245: hello clippy! 😁

weichweich (commented):
looks good! I'm not 100% happy with the debug implementation. Only 99% happy!

pallets/ctype/src/lib.rs
Diiaablo95: This is the way I was going initially. But then when I wanted to use the DID pallet structs in the CTYPE pallet, I was having weird type casting issues that I did not manage to solve in an afternoon (something like `type annotations needed: cannot satisfy <_ as did::Config>::DidIdentifier == <T as did::Config>::DidIdentifier).

Plus, I think it would be quite repetitive to have the same type constraints, which are basically a logic of the DID pallet, in all the other pallets that want to implement the `DidOperation` trait. In this way, a pallet just must require the implementation of the did Config Trait as well, and can be readily used.

pallets/did/src/lib.rs
Diiaablo95: Same as above.

pallets/ctype/src/lib.rs
weichweich: Since we have the did pallet everywhere you can just use these associated types from there. No need to repeat them anywhere.

Diiaablo95: Yes but to use the pallet associated type, we need a reference to the pallet's Config trait, since we cannot simply reference it from everywhere. That is why the type parametrisation over the pallet's Config trait, which in turn extends the did pallet's Config trait. Are there other ways to access the did pallet's `DidIdentifier` type directly that I am not aware of?

pallets/did/src/lib.rs
```diff
@@ -537,12 +588,6 @@ where
 			);
 		};
 
-		// Verify that the operation counter is greater than the stored one
```
wischli, line 540: Could you explain the why this does not hold true anymore? Sorry if I has been explained already.

pallets/ctype/src/mock.rs
```diff
@@ -0,0 +1,141 @@
...
+
+	pub fn build(self) -> sp_io::TestExternalities {
+		let mut ext = if let Some(did_builder) = self.did_builder.clone() {
+			did_builder.build()
+		} else {
+			let storage = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
+			sp_io::TestExternalities::new(storage)
+		};
+
+		if self.ctypes_stored.len() > 0 {
```
wischli, line 131: Same as below
```suggestion
		if !self.ctypes_stored.is_empty() {
```

wischli (commented):
One minor remark and a question additionally to @weichweich's remarks which I share.

pallets/ctype/src/lib.rs
weichweich: Doesn't `<Config as did::Config>::DidIdentifier` work?

pallets/did/src/lib.rs
Diiaablo95: Yeah we basically briefly discussed with @weichweich that if we allow users to put ANY value greater than the current one, we might face the situation in which we reach the maximum counter value. And as we do not have at the moment a strategy to wrap those around, we are just making it harder to happen by forcing users to only nonces that are 1 greater than the last one used.

pallets/ctype/src/lib.rs
weichweich: got the problem now! as we discussed, keep it like it is for now.

Diiaablo95:
Closing in favor of https://github.com/KILTprotocol/mashnet-node/pull/146.

PR #144, opened 2021-04-22
fix: clippy

wischli:
## no ticket
* Add clippy allowance to all pallets and runtimes to temporarily fix clippy issues from Substrate macros
  * Note: It's not required for all pallets but DID in the current state of develop but will be required once we merge #141 in which we upgrade the pallets to frame v2
* Fix other clippy warnings
* Move constants from parachain runtime to primitives as also done in #129 

## How to test:
CI pipeline for clippy should go green

## Checklist:

### DIFF
--- a/runtimes/parachain/src/lib.rs
+++ b/runtimes/parachain/src/lib.rs
@@ -20,6 +20,8 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 // `construct_runtime!` does a lot of recursion and requires us to increase the limit to 256.
 #![recursion_limit = "256"]
+// The `from_over_into` warning originates from `construct_runtime` macro.
+#![allow(clippy::from_over_into)]
 
 // Make the WASM binary available.
 #[cfg(feature = "std")]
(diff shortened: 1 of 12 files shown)


runtimes/parachain/src/lib.rs
```diff
@@ -492,7 +478,7 @@ parameter_types! {
 	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
 	pub const VotingPeriod: BlockNumber = 7 * DAYS;
 	pub const FastTrackVotingPeriod: BlockNumber =  3 * HOURS;
-	pub const MinimumDeposit: Balance = 1 * DOLLARS;
+	pub const MinimumDeposit: Balance = 2 * DOLLARS;
```
weichweich, line 481: nice fix 😂

weichweich (approved):
LGTM

runtimes/parachain/src/lib.rs
wischli: Yeah... I was not sure but since this will be changed before the launch I thought it would be fine 😅

PR #145, opened 2021-04-22
feat: staking

wischli:
## fixes KILTProtocol/ticket#1253

* Adds parachain staking
* adds shell runtime execution
* spiritnet runtime
* specify which runtime should be used with `--runtime {spiritnet,mashnet}`

## How to test:
```
cargo test -p parachain-staking
```

### custom types

This PR introduces new custom JS-types which are required for compatibility with [our SDK](https://github.com/KILTprotocol/sdk-js) and the [Polkadot Apps](https://polkadot.js.org/apps/#/extrinsics). 

[Companion Types PR](https://github.com/KILTprotocol/type-definitions/pull/6)

Please use the following types to test the code with the Polkadot Apps:

<details>
  <summary>JS-Types</summary>
  
  ```json
{
  "AccountInfo": "AccountInfoWithTripleRefCount"
  "Address": "MultiAddress",
  "AmountOf": "i128",
  "Balance": "u128",
  "BlockNumber": "u64",
  "Index": "u64",
  "LookupSource": "MultiAddress",
  "CtypeCreatorOf": "DidIdentifierOf",
  "CtypeHashOf": "Hash",
  "ClaimHashOf": "Hash",
  "AttesterOf": "DidIdentifierOf",
  "AttestationDetails": {
    "ctypeHash": "CtypeHashOf",
    "attester": "AttesterOf",
    "delegationId": "Option<DelegationNodeIdOf>",
    "revoked": "bool"
  },
  "Permissions": "u32",
  "DelegationNodeIdOf": "Hash",
  "DelegatorIdOf": "DidIdentifierOf",
  "DelegationSignature": "DidSignature",
  "DelegationRoot": {
    "ctypeHash": "CtypeHashOf",
    "owner": "DelegatorIdOf",
    "revoked": "bool"
  },
  "DelegationNode": {
    "rootId": "DelegationNodeIdOf",
    "parent": "Option<DelegationNodeIdOf>",
    "owner": "DelegatorIdOf",
    "permissions": "Permissions",
    "revoked": "bool"
  },
  "KeyIdOf": "Hash",
  "DidIdentifierOf": "AccountId",
  "AccountIdentifierOf": "AccountId",
  "BlockNumberOf": "BlockNumber",
  "DidCallableOf": "Call",
  "DidVerificationKey": {
    "_enum": {
      "Ed25519": "[u8; 32]",
      "Sr25519": "[u8; 32]"
    }
  },
  "DidEncryptionKey": {
    "_enum": {
      "X25519": "[u8; 32]"
    }
  },
  "DidPublicKey": {
    "_enum": {
      "PublicVerificationKey": "DidVerificationKey",
      "PublicEncryptionKey": "DidEncryptionKey"
    }
  },
  "DidVerificationKeyRelationship": {
    "_enum": [
      "Authentication",
      "CapabilityDelegation",
      "CapabilityInvocation",
      "AssertionMethod"
    ]
  },
  "DidSignature": {
    "_enum": {
      "Ed25519": "Ed25519Signature",
      "Sr25519": "Sr25519Signature"
    }
  },
  "DidError": {
    "_enum": {
      "StorageError": "StorageError",
      "SignatureError": "SignatureError",
      "UrlError": "UrlError",
      "InternalError": "Null"
    }
  },
  "StorageError": {
    "_enum": {
      "DidAlreadyPresent": "Null",
      "DidNotPresent": "Null",
      "DidKeyNotPresent": "DidVerificationKeyRelationship",
      "VerificationKeyNotPresent": "Null",
      "CurrentlyActiveKey": "Null",
      "MaxTxCounterValue": "Null"
    }
  },
  "SignatureError": {
    "_enum": [
      "InvalidSignatureFormat",
      "InvalidSignature",
      "InvalidNonce"
    ]
  },
  "KeyError": {
    "_enum": [
      "InvalidVerificationKeyFormat",
      "InvalidEncryptionKeyFormat"
    ]
  },
  "UrlError": {
    "_enum": [
      "InvalidUrlEncoding",
      "InvalidUrlScheme"
    ]
  },
  "DidPublicKeyDetails": {
    "key": "DidPublicKey",
    "blockNumber": "BlockNumberOf"
  },
  "DidDetails": {
    "authenticationKey": "KeyIdOf",
    "keyAgreementKeys": "BTreeSet<KeyIdOf>",
    "delegationKey": "Option<KeyIdOf>",
    "attestationKey": "Option<KeyIdOf>",
    "publicKeys": "BTreeMap<KeyIdOf, DidPublicKeyDetails>",
    "endpointUrl": "Option<Url>",
    "lastTxCounter": "u64"
  },
  "DidCreationOperation": {
    "did": "DidIdentifierOf",
    "newAuthenticationKey": "DidVerificationKey",
    "newKeyAgreementKeys": "BTreeSet<DidEncryptionKey>",
    "newAttestationKey": "Option<DidVerificationKey>",
    "newDelegationKey": "Option<DidVerificationKey>",
    "newEndpointUrl": "Option<Url>"
  },
  "DidUpdateOperation": {
    "did": "DidIdentifierOf",
    "newAuthenticationKey": "Option<DidVerificationKey>",
    "newKeyAgreementKeys": "BTreeSet<DidEncryptionKey>",
    "attestationKeyUpdate": "DidVerificationKeyUpdateAction",
    "delegationKeyUpdate": "DidVerificationKeyUpdateAction",
    "publicKeysToRemove": "BTreeSet<KeyIdOf>",
    "newEndpointUrl": "Option<Url>",
    "txCounter": "u64"
  },
  "DidVerificationKeyUpdateAction": {
    "_enum": {
      "Ignore": "Null",
      "Change": "DidVerificationKey",
      "Delete": "Null"
    }
  },
  "DidDeletionOperation": {
    "did": "DidIdentifierOf",
    "txCounter": "u64"
  },
  "DidAuthorizedCallOperation": {
    "did": "DidIdentifierOf",
    "txCounter": "u64",
    "call": "DidCallableOf"
  },
  "HttpUrl": {
    "payload": "Text"
  },
  "FtpUrl": {
    "payload": "Text"
  },
  "IpfsUrl": {
    "payload": "Text"
  },
  "Url": {
    "_enum": {
      "Http": "HttpUrl",
      "Ftp": "FtpUrl",
      "Ipfs": "IpfsUrl"
    }
  },
  "LockedBalance": {
    "block": "BlockNumber",
    "amount": "Balance"
  },
  "RoundIndex": "u32",
  "RoundInfo": {
    "current": "RoundIndex",
    "first": "BlockNumber",
    "length": "BlockNumber"
  },
  "OrderedSet": "Vec<Stake>",
  "Stake": {
    "owner": "AccountId",
    "amount": "Balance"
  },
  "TotalStake": {
    "collators": "Balance",
    "delegators": "Balance"
  },
  "InflationInfo": {
    "collator": "StakingInfo",
    "delegator": "StakingInfo"
  },
  "StakingInfo": {
    "maxRate": "Perquintill",
    "rewardRate": "RewardRate"
  },
  "RewardRate": {
    "annual": "Perquintill",
    "perBlock": "Perquintill"
  },
  "Delegator": {
    "delegations": "Vec<Stake>",
    "total": "Balance"
  },
  "CollatorSnapshot": {
    "stake": "Balance",
    "delegators": "Vec<Stake>",
    "total": "Balance"
  },
  "Collator": {
    "id": "AccountId",
    "stake": "Balance",
    "delegators": "Vec<Stake>",
    "total": "Balance",
    "state": "CollatorStatus"
  },
  "CollatorStatus": {
    "_enum": {
      "Active": "Null",
      "Idle": "Null",
      "Leaving": "SessionIndex"
    }
  }
}
  ```
</details>
## Checklist:


pallets/parachain-staking/src/inflation.rs
```diff
@@ -0,0 +1,427 @@
...
+	// TODO: Remove and replace usage with `compute_block_rewards`
+	pub fn compute_rewards<T: Config>(&self, stake: BalanceOf<T>, total_issuance: BalanceOf<T>) -> BalanceOf<T> {
+		let staking_rate = Perbill::from_rational(stake, total_issuance).min(self.max_rate);
+		let rewards = self.reward_rate.round * total_issuance;
+		staking_rate * rewards
+	}
+
+	pub fn compute_block_rewards<T: Config>(&self, stake: BalanceOf<T>, total_issuance: BalanceOf<T>) -> BalanceOf<T> {
+		// TODO: Remove shifting after applying Perquintill to InflationInfo
+		let max_rate: u64 = (self.max_rate.deconstruct() as u64).saturating_mul(1000000000u64);
```
shawntabrizi, line 106: Where does ```1000000000u64``` come from?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+	impl<AccountId: Ord, Balance: PartialEq> PartialOrd for Bond<AccountId, Balance> {
+		fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+			Some(self.cmp(other))
+		}
+	}
+
+	impl<AccountId: Ord, Balance: PartialEq> PartialEq for Bond<AccountId, Balance> {
+		fn eq(&self, other: &Self) -> bool {
+			self.owner == other.owner && self.amount == other.amount
+		}
```
shawntabrizi, line 128: doesn't this automatically derive from `#[derive(PartialEq)`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+	/// selected
+	pub struct CollatorSnapshot<AccountId, Balance> {
+		pub bond: Balance,
+		pub delegators: Vec<Bond<AccountId, Balance>>,
+		pub total: Balance,
+	}
+
+	impl<AccountId: Ord, Balance: PartialEq> PartialEq for CollatorSnapshot<AccountId, Balance> {
+		fn eq(&self, other: &Self) -> bool {
+			self.bond == other.bond && self.total == other.total && self.delegators.eq(&other.delegators)
```
shawntabrizi, line 159: same

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+		}
+	}
+
+	impl<AccountId: Ord, Balance: PartialEq> Eq for CollatorSnapshot<AccountId, Balance> {}
+
+	#[derive(Encode, Decode, RuntimeDebug)]
+	/// Global collator state with commission fee, bonded stake, and delegations
+	pub struct Collator<AccountId, Balance> {
+		pub id: AccountId,
```
shawntabrizi, line 168: with this much type definition code, might be cool to refactor into a `types.rs`

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+		pub fn is_active(&self) -> bool {
+			self.state == CollatorStatus::Active
+		}
+
+		pub fn is_leaving(&self) -> bool {
+			matches!(self.state, CollatorStatus::Leaving(_))
+		}
+
+		pub fn bond_more(&mut self, more: B) {
+			// TODO: use saturaring_add instead?
```
shawntabrizi, line 198: yes please

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+		// Returns Some(remaining balance), must be more than MinDelegatorStk
+		// Returns None if delegation not found
+		pub fn rm_delegation(&mut self, collator: AccountId) -> Option<Balance> {
+			let mut amt: Option<Balance> = None;
+			let delegations = self
+				.delegations
+				.0
+				.iter()
+				.filter_map(|x| {
```
shawntabrizi, line 297: should be ordered and be able to do a binary search.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+						amt = Some(x.amount);
+						None
+					} else {
+						Some(x.clone())
+					}
+				})
+				.collect();
+			if let Some(balance) = amt {
+				self.delegations = OrderedSet::from(delegations);
+				self.total -= balance;
```
shawntabrizi, line 308: def need saturation here

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				Some(self.total)
+			} else {
+				None
+			}
+		}
+
+		// Returns None if delegation not found
+		pub fn inc_delegation(&mut self, collator: AccountId, more: Balance) -> Option<Balance> {
+			for x in &mut self.delegations.0 {
+				if x.owner == collator {
```
shawntabrizi, line 318: binary search again

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				bond: other.bond,
+				delegators: other.delegators.0,
+				total: other.total,
+			}
+		}
+	}
+
+	#[derive(Encode, Decode, RuntimeDebug)]
+	pub struct Delegator<AccountId, Balance> {
+		pub delegations: OrderedSet<Bond<AccountId, Balance>>,
```
shawntabrizi, line 260: you will need to use a `Bounded` set in the near future to work with PoV stuff

shawntabrizi: https://github.com/paritytech/substrate/issues/8629

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+		BlocksPerRoundSet(RoundIndex, T::BlockNumber, u32, u32),
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(n: T::BlockNumber) {
+			let mut round = <Round<T>>::get();
+			if round.should_update(n) {
+				// kill snapshot of current round
+				<AtStake<T>>::remove_prefix(round.current);
```
shawntabrizi, line 512: I don't know approximately how big this iteration will be, but you def can NOT do this on a on_finalize.

Should be moved to `on_initialize`, or better cleaned up via transactions.

Remember if for whatever reason this ends up too big, and it pushes the PoV time or computation time over the limit, you will have bricked your chain.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+		RoundInflationSet(Perbill, Perbill, Perbill, Perbill),
+		/// Set total selected candidates to this value [old, new]
+		TotalSelectedSet(u32, u32),
+		/// Set blocks per round [current_round, first_block, old, new]
+		BlocksPerRoundSet(RoundIndex, T::BlockNumber, u32, u32),
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(n: T::BlockNumber) {
```
shawntabrizi, line 508: Basically you should never use `on_finalize` except to delete like 1 or 2 transient storage items

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+	#[pallet::storage]
+	#[pallet::getter(fn selected_candidates)]
+	/// The collator candidates selected for the current round
+	type SelectedCandidates<T: Config> = StorageValue<_, Vec<T::AccountId>, ValueQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn total)]
+	/// Total capital locked by this staking pallet
+	// TODO: Might want to use Struct instead of Tuple
```
shawntabrizi, line 565: yes i agree with this. zero overhead cost, but better ergonomics when writing code.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+					)
+				} else {
+					<Pallet<T>>::join_candidates(T::Origin::from(Some(actor.clone()).into()), balance)
+				};
+			}
+			// Set total selected candidates to minimum config
+			<TotalSelected<T>>::put(T::MinSelectedCandidates::get());
+			// Choose top TotalSelected collator candidates
+			let (v_count, collator_staked, delegator_staked) = <Pallet<T>>::select_top_candidates(1u32);
+			// Start Round 1 at Block 0
```
shawntabrizi, line 645:
```suggestion
			// Start Round 1 at Block 0
```

a nit preference is that things should be zero based. Should be no problem to start at "round 0"

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				delegator_staked,
+			));
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Set the annual inflation rate to derive per-round inflation
+		#[pallet::weight(0)]
+		pub fn set_inflation(origin: OriginFor<T>, inflation: InflationInfo) -> DispatchResultWithPostInfo {
```
shawntabrizi, line 663:
```suggestion
		pub fn set_inflation(origin: OriginFor<T>, inflation: InflationInfo) -> DispatchResult {
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+			frame_system::ensure_root(origin)?;
+
+			Self::update_inflation(inflation)?;
+			Ok(().into())
```
shawntabrizi, line 667:
```suggestion
			Ok(())
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+			ensure!(!Self::is_delegator(&acc), Error::<T>::DelegatorExists);
+			ensure!(bond >= T::MinCollatorCandidateStk::get(), Error::<T>::ValBondBelowMin);
+			ensure!(bond <= T::MaxCollatorCandidateStk::get(), Error::<T>::ValBondAboveMax);
+
+			let mut candidates = <CandidatePool<T>>::get();
+			ensure!(
+				candidates.insert(Bond {
+					owner: acc.clone(),
+					amount: bond
+				}),
```
shawntabrizi, line 719: there should be some max number of candidates?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+						owner: collator.clone(),
+						amount
+					}),
+					Error::<T>::AlreadyDelegatedCollator
+				);
+				let delegation = Bond {
+					owner: acc.clone(),
+					amount,
+				};
+				// TODO: How to handle: Kick lowest bonded delegators or keep as is?
```
shawntabrizi, line 874: should kick or else you are open to a simple attack on candidates

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				ensure!(
+					(state.delegators.0.len() as u32) < T::MaxDelegatorsPerCollator::get(),
+					Error::<T>::TooManyDelegators
+				);
+				ensure!(state.delegators.insert(delegation), Error::<T>::DelegatorExists);
+				T::Currency::reserve(&acc, amount)?;
+				let new_total = state.total + amount;
```
shawntabrizi, line 881: lots of unsafe math where i dunno its guaranteed not to fail.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+					owner: acc.clone(),
+					amount,
+				};
+				// TODO: How to handle: Kick lowest bonded delegators or keep as is?
+				ensure!(
+					(state.delegators.0.len() as u32) < T::MaxDelegatorsPerCollator::get(),
+					Error::<T>::TooManyDelegators
+				);
+				ensure!(state.delegators.insert(delegation), Error::<T>::DelegatorExists);
+				T::Currency::reserve(&acc, amount)?;
```
shawntabrizi, line 880: If you use reserve, the users cannot use this same stake for voting or other things. normally you would use a lock.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+				// Update states
+				let (total_collators, total_delegators) = <Total<T>>::get();
+				<Total<T>>::put((total_collators, total_delegators.saturating_add(amount)));
+				state.total = new_total;
+				<CollatorState<T>>::insert(&collator, state);
+				<DelegatorState<T>>::insert(&acc, delegator);
+				Self::deposit_event(Event::Delegation(acc, amount, collator, new_total));
+			} else {
+				// first delegation
```
shawntabrizi, line 894: such a big `if`/`else` statement will be hard to weigh. Might be worth just making a separate `first_delegate` extrinsic or similar.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				.ok_or(Error::<T>::DelegationDNE)?
+				.ok_or(Error::<T>::Underflow)?;
+
+			ensure!(remaining >= T::MinDelegation::get(), Error::<T>::DelegationBelowMin);
+			ensure!(
+				delegations.total >= T::MinDelegatorStk::get(),
+				Error::<T>::NomBondBelowMin
+			);
+
+			T::Currency::unreserve(&delegator, less);
```
shawntabrizi, line 992:
```suggestion
			let err_amt = T::Currency::unreserve(&delegator, less);
			debug_assert!(err_amt.is_zero());
```

shawntabrizi: You will be surprised when having such a pattern can catch errors which would not be found normally in tests.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+		/// NOTE: Unlocking automatically occurs in `note_author` after a block
+		/// author has produced a block. If you assume that a specific collator
+		/// candidate (and their  corresponding delegators) is an active
+		/// collator for every round, it would be unnecessary to ever call
+		/// `unlock_rewards` for such collator and their delegators.
+		#[pallet::weight(0)]
+		pub fn unlock_rewards(
+			origin: OriginFor<T>,
+			// TODO: Better to use Lookup?
+			// target: <T::Lookup as StaticLookup>::Source
```
shawntabrizi, line 1018: yeah in general it is.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+			// mint
+			if reward > T::Currency::minimum_balance() {
+				if let Ok(imb) = T::Currency::deposit_into_existing(who, reward) {
+					// TODO: Remove?
+					Self::deposit_event(Event::Rewarded(who.clone(), imb.peek()));
+				}
+			}
+
+			// set & update lock
+			let mut locks = <RewardLocks<T>>::get(who);
```
shawntabrizi, line 1238: Bug.

This gets called even if the above if statement would not deposit a reward.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1322 @@
...
+				inflation.delegator.reward_rate.round,
+			));
+			<InflationConfig<T>>::put(inflation);
+			Ok(())
+		}
+
+		// Weight: writes(2)
+		fn do_reward(who: &T::AccountId, reward: BalanceOf<T>, now: T::BlockNumber) {
+			// mint
+			if reward > T::Currency::minimum_balance() {
```
shawntabrizi, line 1230: Youve checked this like 3 times to get this to code, and I dont think you need to check it at all. In order to be a delegator or collator, you must already have an account right?

So why are we worried about minimum balance stuff?

wischli: I agree. Honestly, I thought we might not want to mint lower amounts but it's true that they must have an account. Thus, I remove this check everywhere.

pallets/parachain-staking/src/lib.rs
wischli: Moved to `on_initialize` in 63fa2e4.

I don't know how I we could solve this in transactions: Both the collator exits as well as setting the new candidates should be applied automatically every new round.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+				}),
+				Error::<T>::CandidateExists
+			);
+
+			// Post-launch TODO: Replace with `check_collator_candidate_inclusion`.
+			ensure!(
+				(candidates.len() as u32) <= T::MaxCollatorCandidates::get(),
+				Error::<T>::TooManyCollatorCandidates
+			);
+			T::Currency::reserve(&acc, bond)?;
```
Diiaablo95, line 458: What happens if this call fails but the storage has been updated with the previous `candidates.insert()`? Should this also be transactional?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+			Self::deposit_event(Event::CollatorScheduledExit(now, collator, when));
+			Ok(().into())
+		}
+
+		/// Temporarily leave the set of collator candidates without unbonding
+		#[pallet::weight(0)]
+		pub fn go_offline(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+			let collator = ensure_signed(origin)?;
+			let mut state = <CollatorState<T>>::get(&collator).ok_or(Error::<T>::CandidateDNE)?;
+			ensure!(state.is_active(), Error::<T>::AlreadyOffline);
```
Diiaablo95, line 512: What if state is `is_leaving`? Should that trigger an AlreadyOffline error? Might be better something like NotActive, or something similar to what is done in `go_online()`

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+				Self::update_active(collator.clone(), state.total);
+			}
+			<CollatorState<T>>::insert(&collator, state);
+			Self::deposit_event(Event::CollatorBondedLess(collator, before, after));
+			Ok(().into())
+		}
+
+		/// Join the set of delegators by delegating to a collator candidate.
+		///
+		/// NOTE: Expects the caller to have delegated before. Otherwise, they
```
Diiaablo95, line 593: to NOT have delegated before probably?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+				});
+
+				// TODO: Add update of select_top_candidates
+				<CollatorState<T>>::insert(&collator, state);
+				<DelegatorState<T>>::insert(&acc, delegator);
+				Self::deposit_event(Event::Delegation(acc, amount, collator, new_total));
+
+				Ok(().into())
+			} else {
+				Err(Error::<T>::NotYetDelegating.into())
```
Diiaablo95, line 718: Maybe move this at the beginning with an `ensure!`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+				.ok_or(Error::<T>::DelegationDNE)?;
+			// edge case; if no delegations remaining, leave set of delegators
+			if delegator.delegations.is_empty() {
+				// leave the set of delegators because no delegations left
+				Self::delegator_leaves_collator(acc.clone(), collator)?;
+				<DelegatorState<T>>::remove(&acc);
+				Self::deposit_event(Event::DelegatorLeft(acc, old_total));
+				return Ok(().into());
+			}
+			ensure!(remaining >= T::MinDelegatorStk::get(), Error::<T>::NomBondBelowMin);
```
Diiaablo95, line 878: This `ensure!` should never fail, right? Better to highlight that. If it does, what happens to the partially updated storage?

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1163 @@
...
+				.collect::<Vec<T::AccountId>>();
+
+			// Snapshot exposure for round for weighting reward distribution
+			for account in collators.iter() {
+				let state = <CollatorState<T>>::get(&account).expect("all members of CandidateQ must be candidates");
+				let amount_collator = state.bond;
+				let amount_delegators = state.total.saturating_sub(amount_collator);
+				let exposure: CollatorSnapshot<T::AccountId, BalanceOf<T>> = state.into();
+				<AtStake<T>>::insert(next, account, exposure);
+				all_collators += 1u32;
```
Diiaablo95, line 997: `saturating_add()`?

pallets/parachain-staking/src/types.rs
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	pub fn bond_more(&mut self, more: B) {
+		self.bond = self.bond.saturating_add(more);
+		self.total = self.total.saturating_add(more);
+	}
+
+	// Returns None if underflow or less == self.bond (in which case collator should
+	// leave)
+	pub fn bond_less(&mut self, less: B) -> Option<B> {
```
Diiaablo95, line 145: Use `checked_sub()` instead?

pallets/parachain-staking/src/set.rs
```diff
@@ -76,10 +105,37 @@ impl<T: Ord> OrderedSet<T> {
 		self.0.binary_search(&value).is_ok()
 	}
 
+	/// Return if the set contains `value`
```
Diiaablo95, line 108: Return... what? 🤣

pallets/parachain-staking/src/set.rs
```diff
@@ -76,10 +105,37 @@ impl<T: Ord> OrderedSet<T> {
...
+	pub fn binary_search(&self, value: &T) -> Result<usize, usize> {
+		self.0.binary_search(&value)
+	}
+
+	/// Return if the set contains `value`
+	pub fn binary_search_by<'a, F>(&'a self, f: F) -> Result<usize, usize>
```
Diiaablo95, line 114: Same here

pallets/parachain-staking/src/set.rs
weichweich: someone™ might have just copied the doc string from another function. Happens further below again a couple of times.

pallets/parachain-staking/src/set.rs
```diff
@@ -76,10 +105,37 @@ impl<T: Ord> OrderedSet<T> {
...
+	pub fn len(&self) -> usize {
+		self.0.len()
+	}
+
+	/// Clear the set
+	pub fn is_empty(&self) -> bool {
+		self.0.is_empty()
+	}
+
+	pub fn to_vec(self) -> Vec<T> {
```
Diiaablo95, line 136: Should we just implement `From<Vec<T>>`?

weichweich: We actually have both.

* `impl<T: Ord> From<Vec<T>> for OrderedSet<T>` To get from a `Vec` to a `Set`
* `impl<T: Ord> From<OrderedSet<T>> for Vec<T>` to get from the `Set` to a `Vec`

When you try the to use into, the compiler sometimes is not able to derive `T` (e.g. where I used to_vec()). That's why I added `to_vec`. Since `T` is already set here, we don't need to specify it when we would call `into`.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1160,4 +1158,26 @@ pub mod pallet {
...
+			log::info!(
+				"assembling new collators for new session {} at #{:?}",
+				new_index,
+				<frame_system::Pallet<T>>::block_number(),
+			);
+
+			Self::select_top_candidates(new_index);
+
+			Some(<SelectedCandidates<T>>::get())
+		}
```
weichweich: @wischli we need to select the new collators here. Is this correct!?

runtimes/parachain/src/lib.rs
```diff
@@ -269,6 +287,40 @@ impl parachain_info::Config for Runtime {}
...
+	pub const Period: u64 = 0xFFFF_FFFF_FFFF_FFFF;
+	pub const Offset: u64 = 0xFFFF_FFFF_FFFF_FFFF;
+}
+
+impl pallet_session::Config for Runtime {
+	type Event = Event;
+	type ValidatorId = AccountId;
+	type ValidatorIdOf = ();
+	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
+	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
```
weichweich, line 316: @wischli We specify the session duration here. This should be controlled by the `ParachainStaking` too?

pallets/parachain-staking/src/lib.rs
wischli: Yes, looks good to me.

runtimes/parachain/src/lib.rs
wischli: I suppose. Or to put it into other words: The session duration should equal the duration of a round in which we select the top candidates to become active collators for this round.

pallets/parachain-staking/src/set.rs
Diiaablo95: But that's what turbofish is for, right? Something like `obj.into::<Vec<T>>()`?

weichweich: Yep. But I'm too lazy to explicitly state the type that I want to convert into. 😁 

It's probably named wrong. Should be `into_vec`. You can compare that with the [BinaryHeap::into_vec](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#method.into_vec). They even use `into` in `into_vec` 😂 . It's just a short hand, so that you don't have to specify the type again.

pallets/parachain-staking/src/set.rs
weichweich: We should probably just copy the docs from the rust doc


>Binary searches this ~~ordered VecDeque~~OrderedSet for a given element.
>
>If the value is found then Result::Ok is returned, containing the index of the matching element. ~~If there are multiple matches, then any one of the matches could be returned.~~ If the value is not found then Result::Err is returned, containing the index where a matching element could be inserted while maintaining sorted order.

pallets/parachain-staking/src/lib.rs
wischli: Yeah I would like to have all extrinsics here be transactional. Unfortunately, I ran into compilation issues. If those cannot be resolved, we have to refactor the extrinsics to not make storage changes before potential errors!

wischli: Tracking issue: https://github.com/kiltprotocol/ticket/issues/1335

pallets/parachain-staking/src/lib.rs
wischli: Good catch! Fixed in 58dcb24ffb425040c499fb2c2e292287a2d9105c

pallets/parachain-staking/src/lib.rs
wischli: Yes, thanks!

pallets/parachain-staking/src/lib.rs
wischli: Definitely!
```rust
let mut delegator = <DelegatorState<T>>::get(&acc).ok_or(Error::<T>::NotYetDelegating)?;
```

pallets/parachain-staking/src/lib.rs
wischli: It cannot fail if `MinDelegatorStk == MinDelegation`. I will still mention this in a comment. Good catch, had to look it up to be sure.
```rust
/// Minimum stake for any registered on-chain account to delegate
type MinDelegation: Get<BalanceOf<Self>>;
/// Minimum stake for any registered on-chain account to become a delegator
type MinDelegatorStk: Get<BalanceOf<Self>>;
```

pallets/parachain-staking/src/set.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/set.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/lib.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/lib.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/lib.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/lib.rs
wischli: Fixed in af20eb0

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1295 @@
...
+
+		// TODO: Replace with function without delay
+		fn execute_delayed_collator_exits(next: RoundIndex) {
+			let remain_exits = <ExitQueue<T>>::get()
+				.into_iter()
+				.filter_map(|x| {
+					if x.amount > next {
+						Some(x)
+					} else {
+						if let Some(state) = <CollatorState<T>>::get(&x.owner) {
```
Diiaablo95, line 950: Perhaps replace with `let state = <CollatorState<T>>::get(&x.owner)?`?

Diiaablo95: So no need to have a big if branch and no need to return `None` otherwise.

pallets/parachain-staking/src/lib.rs
```diff
@@ -0,0 +1,1295 @@
...
+			// Order candidates by their total stake
+			candidates.sort_unstable_by(|a, b| a.amount.cmp(&b.amount));
+
+			// Choose the top TotalSelected qualified candidates, ordered by stake (least to
+			// greatest, thus requires `rev()`)
+			let mut collators = candidates
+				.into_iter()
+				.rev()
+				.take(top_n)
+				.filter(|x| x.amount >= T::MinCollatorStk::get())
```
Diiaablo95, line 1018: What happens if any of the `top_n` has less than the minimum amount? I guess this is also impossible if the other functions are correctly implemented, but would be good to have this highlighted here.

pallets/parachain-staking/src/inflation.rs
```diff
@@ -0,0 +1,253 @@
...
+
+		let collator_rewards = self.collator.compute_block_rewards::<T>(collator_stake, circulating);
+		let delegator_rewards = self.delegator.compute_block_rewards::<T>(delegator_stake, circulating);
+
+		(collator_rewards, delegator_rewards)
+	}
+
+	/// Check whether the annual reward rate is approx. the per_block reward
+	/// rate multiplied with the number of blocks per year
+	pub fn is_valid(&self) -> bool {
```
Diiaablo95, line 115: Should we maybe have this checked every time an `InflationInfo` instance is created, rather than having to check each time that the instance is valid? So we would be sure that each time `InflationInfo` is used, it is valid.

pallets/parachain-staking/src/lib.rs
Diiaablo95: Ok I had a different thing in mind.

PR #154, opened 2021-05-07
feat: move version to primitives

weichweich:
* Add companion PR for types to template
* use the same version in the parachain and standalone
  * this way we can use the same types and have the same version bounds
  * the two versions are very close (and should probably be the same anyways)
* also removes old currency types

### custom types

types are unaffected.

## Checklist:


weichweich:
doesn't work because of `RUNTIME_API_VERSIONS` coming from the runtime... :)

PR #155, opened 2021-05-10
feat: remove society

weichweich:
## No Ticket

remove society since we might not need it and it adds a lot of complexity.

### custom types

No custom types

## Checklist:


rflechtner:
lol I think you just summed up anarchism's political agenda in a sentence

runtimes/parachain/src/lib.rs
```diff
@@ -419,22 +408,8 @@ parameter_types! {
...
-	type PeriodSpend = PeriodSpend;
-	type MembershipChanged = ();
-	type RotationPeriod = RotationPeriod;
-	type MaxLockDuration = MaxLockDuration;
-	type FounderSetOrigin = pallet_collective::EnsureProportionMoreThan<_1, _2, AccountId, CouncilCollective>;
-	type SuspensionJudgementOrigin = pallet_society::EnsureFounder<Runtime>;
-	type ChallengePeriod = ChallengePeriod;
-	type MaxCandidateIntake = MaxCandidateIntake;
+pub const fn deposit(items: u32, bytes: u32) -> Balance {
+	items as Balance * 20 * DOLLARS + (bytes as Balance) * 100 * MILLICENTS
```
Diiaablo95, line 412: Where are these values coming from?

Diiaablo95 (approved):
I guess it looks good 😄 just a minor question.

runtimes/parachain/src/lib.rs
weichweich: It's copy pasted from the governance config from kusama. It doesn't really matter since it's for the election phragmen pallet which we only have in here as a place holder at the moment.

weichweich: We didn't really put much thought into any of these values yet. I think the values parity uses are sensible defaults and we will check and reevaluate these values when we tackle this: https://github.com/KILTprotocol/ticket/issues/1194

PR #160, opened 2021-05-17
feat: peregrine spec

weichweich:
## related to KILTProtocol/ticket#1227

* adds a feature that reduces the duration of all governance related eras to a few minutes
* change specs with JQ

## Checklist:


nodes/parachain/src/chain_spec.rs
```diff
@@ -146,6 +146,36 @@ pub fn staging_test_net(id: ParaId) -> Result<ChainSpec, String> {
 	))
 }
 
+pub fn peregrine_test_net(id: ParaId) -> Result<ChainSpec, String> {
+	let properties = get_properties("PKILT", 15, 38);
```
Diiaablo95, line 150: What are these 15 and 38?

Diiaablo95: And also PKILT?

weichweich: PKILT = peregrine-kilt

15 = 10^15 units are one PKILT

38 = The SS58 identifier

Yeah it's not that readable. But i would hope that's something for a different PR? 😅  We do it for all specs like that.

scripts/peregrine-relay.jq
```diff
@@ -0,0 +1,49 @@
+.bootNodes = []
```
Diiaablo95, line 1: Love this approach! 😎

scripts/spec.sh
```diff
@@ -1,17 +1,33 @@
...
 
+ROCOCO_PLAIN=/tmp/parachain/rococo.plain.json
+ROCOCO=/tmp/parachain/rococo.json
+
+# ##############################################################################
+# #                                                                            #
+# #                                  PEREGRINE                                 #
+# #                                                                            #
+# ##############################################################################
+docker run parity/rococo:rococo-v1-0.8.30-943038a8-f14fa75f build-spec --chain rococo-local --disable-default-bootnode >$ROCOCO_PLAIN
```
Diiaablo95, line 12: I would leave a space before and after the redirect symbol ">".

scripts/spec.sh
```diff
@@ -1,17 +1,33 @@
...
+cargo run --release -p kilt-parachain --features fast-gov -- build-spec --chain peregrine --disable-default-bootnode >peregrine-kilt.plain.spec
```
Diiaablo95, line 13: Ditto

Diiaablo95 (approved):
Few minor comments, but I guess LVVGTM 😁

scripts/spec.sh
weichweich: I would too, but i have a weird bash formatter that removes it. 🤔 i added it. lets hope it survives the formatter for a while 😅

PR #163, opened 2021-05-21
fix: don't transfer expired locks

weichweich:
## fixes KILTProtocol/ticket#1341

>  When I migrate an account with a locked amount that unlocks at a block that is already in the past (block the amount is unlocked: 250 and the current block is 500), the amount is not unlocked.

## Checklist:

### DIFF
--- a/pallets/kilt-launch/src/lib.rs
+++ b/pallets/kilt-launch/src/lib.rs
@@ -721,6 +721,23 @@ pub mod pallet {
 			// Otherwise, this will always be the source's locked amount
 			let max_add_amount = source_amount.min(max_amount.unwrap_or(source_amount));
 
+			// We don't need to transfer any locks if the lock already expired. So we bail early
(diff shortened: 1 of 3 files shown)


pallets/kilt-launch/src/lib.rs
```diff
@@ -721,6 +721,25 @@ pub mod pallet {
 			// Otherwise, this will always be the source's locked amount
 			let max_add_amount = source_amount.min(max_amount.unwrap_or(source_amount));
 
+			// We don't need to transfer any locks if the lock already expired. So we bail
+			// early
+			if unlock_block <= frame_system::Pallet::<T>::block_number() {
+				// But we still need to reduce the old lock or remove it, if it's consumed
+				// completly.
```
wischli, line 728:
```suggestion
				// completely.
```

wischli (approved):
One minor comment improvement, otherwise LGTM!

PR #168, opened 2021-05-25
feat: add utility pallet

weichweich:
### custom types

no types

## Checklist:


wischli (approved):
LGTM!

runtimes/standalone/Cargo.toml
```diff
@@ -216,6 +216,11 @@ git = "https://github.com/paritytech/substrate"
 package = "pallet-timestamp"
 version = "3.0.0"
 
+[dependencies.pallet-utility]
```
Diiaablo95, line 219: Why do we have mixed dependency styles here? Should we make it consistent?

weichweich: this toml uses the other stile. So in the toml it's not mixed. 😅

Diiaablo95: I see a `[dependencies]` section tho...

weichweich: Ah yes. didn't see that. So IMHO we should use the short version for simple dependencies and the expended version for dependencies that use multiple options.

weichweich: on the other hand it becomes quite verbose and long. So I'm undecided.

Diiaablo95: In my opinion it would be more chaotic, and also I think we have been converging to using the short version in the other pallets. So we should probably use it everywhere, that's my 2 cents here.

PR #171, opened 2021-05-27
chore: update to 0.9.3

weichweich:
## fixes KILTprotocol/ticket/issues/1358

- fixes KILTprotocol/ticket/issues/1357

## Checklist:


nodes/standalone/src/service.rs
```diff
@@ -43,27 +43,22 @@ native_executor_instance!(
...
 	sp_consensus::DefaultImportQueue<Block, FullClient>,
 	sc_transaction_pool::FullPool<Block, FullClient>,
 	(
-		sc_consensus_aura::AuraBlockImport<
-			Block,
-			FullClient,
-			sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
-			AuraPair,
-		>,
+		sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
```
wischli, line 53: Any reason why we switch to GrandPa from AuRa in the standalone runtime? Shouldn't we stick with the same consensus?

wischli (commented):
I have one question regarding switching the consensus for standalone.

nodes/standalone/src/service.rs
weichweich: We don't switch the consensus. We still use Aura+Grandpa. [Aura is just not needed for the block import anymore](https://github.com/paritytech/substrate/pull/8881).

wischli (approved):
LGTM! Thanks for clarifying my question.

PR #173, opened 2021-05-28
feat: refactor collator exit from intrinsic to extrinsic

wischli:
## fixes KILTProtocol/ticket#1361
* Removes the automatic execution of requested collator candidate exits from `on_initialize`
* Candidates which requested to leave in `init_leave_candidates` now also have to execute their request by calling `execute_leave_candidates`
* Adds an extrinsic to revoke a candidate's exit request by calling `cancel_leave_candidates`
* Disables collators to `init_leave_candidates` if the current size of the CandidatePool is below `MinSelectedCandidates`
* fixes KILTprotocol/ticket/issues/1364
* fixes KILTprotocol/ticket/issues/1323

A collator candidate which wants to stop collating and withdraw their total stake now has to call 4 extrinsics which include two separate waiting times before being able to be called in total:
1. Call `init_leave_candidates` which removes the collator from the `CandidatePool` and updates the `SelectedCandidates`
2. Wait at least for `ExitQueueDelay` many rounds
3. Call `execute_leave_candidates` which moves the stake to the `Unstaking` storage for the candidate and all their delegators
4. Wait at least for `StakeDuration` many blocks (same holds true for their delegators)
5. Call `withdraw_unstaked` to remove the lock on the staked balance and make it usable (same holds true for their delegators)

## Todo

- [x] Add test(s) for `cancel_leavel_candidates`
- [x] Add benchmark for `execute_leave_candidates`
- [x] Add post weight correction to `execute_leave_candidates`
- [x] Add benchmark for `cancel_leave_candidates`
- [x] Add post weight correction to `cancel_leave_candidates`

## How to test:
```
```
## Checklist:

(https://github.com/KILTprotocol/type-definitions/pulls)


wischli:
@weichweich This PR is not completely finished yet but since I will be on vacation after Monday lunch time it would be good to have it reviewed at this stage already. The open TODOs are only tests and benchmarks.

pallets/parachain-staking/src/lib.rs
```diff
@@ -834,34 +838,163 @@ pub mod pallet {
...
+		/// candidate bounded by `MaxDelegatorsPerCollator`
+		/// - Reads: CollatorState, Round, D * DelegatorState, D
+		///   * BlockNumber, D * Unstaking
+		/// - Writes: D * Unstaking, D * DelegatorState, Total
+		/// - Kills: CollatorState, DelegatorState
+		/// # </weight>
+		// TODO: Add benchmark
+		#[pallet::weight(100_000_000_000)]
+		pub fn execute_leave_candidates(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+			let collator = ensure_signed(origin)?;
```
shawntabrizi, line 891: I think you would be fine allowing anyone to call this function on behalf of the collator right?

If so, such a design is more ergonomic, since the collator can perform actions with a hot key

pallets/parachain-staking/src/lib.rs
```diff
@@ -834,34 +838,163 @@ pub mod pallet {
...
+			let state = <CollatorState<T>>::get(&collator).ok_or(Error::<T>::CandidateNotFound)?;
+			ensure!(state.is_leaving(), Error::<T>::NotLeaving);
+			ensure!(state.can_exit(<Round<T>>::get().current), Error::<T>::CannotLeaveYet);
+
+			// iterate over delegators
+			let _num_delegators = state.delegators.len() as u32;
```
shawntabrizi, line 897: for weight? or is it unused?

shawntabrizi:
makes sense to me and looks better than before

pallets/parachain-staking/src/lib.rs
wischli: Yes, this will be used for post weight reduction.

pallets/parachain-staking/src/lib.rs
wischli: I wasn't sure because of `cancel_leave_candidates` but my initial hunch was also to allow anyone to call it.

shawntabrizi: they have the entire "leave period" to cancel. If they get the point where the leave period has passed, and they did not cancel, it is totally fine to trigger it on their behalf imo

pallets/parachain-staking/src/benchmarking.rs
```diff
@@ -176,14 +168,67 @@ benchmarks! {
 		}
 
 		let now = <Round<T>>::get().current;
-		let old_candidate = candidates[0].clone();
+		let candidate = candidates[0].clone();
```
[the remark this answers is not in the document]
weichweich: A panic during benchmarking is fine IMO. We have a lot of asserts that can panic anyway.

pallets/parachain-staking/src/lib.rs
```diff
@@ -834,34 +841,182 @@ pub mod pallet {
 		/// - Writes: CollatorState, CandidatePool, ExitQueue, N * AtStake,
 		///   SelectedCandidates
 		/// # </weight>
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::leave_candidates(T::MaxCollatorCandidates::get(), T::MaxCollatorCandidates::get() * T::MaxDelegatorsPerCollator::get()))]
-		pub fn leave_candidates(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::init_leave_candidates(
+			T::MaxCollatorCandidates::get(),
```
Diiaablo95, line 845: Why don't we use the actual number of collators currently active in the set?

Diiaablo95: I mean, here we are just always considering the worst case?

pallets/parachain-staking/src/lib.rs
```diff
@@ -834,34 +841,182 @@ pub mod pallet {
...
 			}
-			<ExitQueue<T>>::put(exits);
 			<CollatorState<T>>::insert(&collator, state);
 
 			// update candidates for next round
 			let (num_collators, num_delegators, _, _) = Self::select_top_candidates();
 
 			Self::deposit_event(Event::CollatorScheduledExit(now, collator, when));
-			Ok(Some(<T as pallet::Config>::WeightInfo::leave_candidates(
+			Ok(Some(<T as pallet::Config>::WeightInfo::init_leave_candidates(
```
Diiaablo95, line 870: Especially if we rare returning it here anyway, no?

Diiaablo95: Ok I guess it depends on `select_top_candidates` being executed, alright. 😄

pallets/parachain-staking/src/lib.rs
weichweich: The weight calculation shouldn't include any storage calls, because calculating the weight needs to be super lightweight and fast. So there are multiple different approaches to calculate the weight.

* the preferred way is to calculate the exact weight from the call parameters
* if the parameters don't provide enough info for the weight, assume the worst case and pay fees back after we know the exact weight
* if the worst case is too heavy and extreme (uses the whole block or similar), the user provides the values for calculating the weight and those values are checked first in the call.

pallets/parachain-staking/src/types.rs
```diff
@@ -219,8 +246,8 @@ where
 		}
 	}
 
-	// Returns Some(remaining balance), must be more than MinDelegatorStk
-	// Returns None if delegation not found
+	/// Returns Some(remaining stake for delecator), must be more than
+	/// MinDelegatorStk Returns None if delegation not found
 	pub fn rm_delegation(&mut self, collator: AccountId) -> Option<Balance> {
```
Diiaablo95, line 251: Maybe collator could be directly a `&AccountId`, so we don't have to clone everytime just to use it as a reference in the search?

pallets/parachain-staking/src/lib.rs
weichweich: > Ok I guess it depends on `select_top_candidates` being executed, alright. 😄

exactly 💯

pallets/parachain-staking/src/types.rs
weichweich: very good point!

pallets/parachain-staking/src/lib.rs
```diff
@@ -1946,7 +2005,10 @@ pub mod pallet {
 			let mut unstaking = <Unstaking<T>>::get(who);
 
 			ensure!(
-				unstaking.len() <= T::MaxUnstakeRequests::get(),
+				u32::try_from(unstaking.len())
```
Diiaablo95, line 2008: All the other parts `usize` is simply used `as u32`, and here it is not. Perhaps we should decide on which one to use?

pallets/parachain-staking/src/types.rs
```diff
@@ -107,10 +110,22 @@ where
 	AccountId: Eq + Ord,
 	Balance: Eq + Ord,
 {
+	/// the collators account id
```
Diiaablo95, line 113: Should they all start with a capital letter, and perhaps a dot at the end?

pallets/parachain-staking/src/types.rs
```diff
@@ -219,8 +246,8 @@ where
 		}
 	}
 
-	// Returns Some(remaining balance), must be more than MinDelegatorStk
-	// Returns None if delegation not found
+	/// Returns Some(remaining stake for delecator), must be more than
```
Diiaablo95, line 249: We should either use the infinite everywhere (e.g. Return) or the third person everywhere (e.g., Returns),

Diiaablo95:
Mostly some style points about comments, which seem to be a bit inconsistent across files, which I guess is the result of more than one person working on it. Plus, there are places where `usize` are casted to `u32`, and one place where the whole `TryInto` flow is called, which is fine, but perhaps should be applied to all the places?
Otherwise, no objection 😎

pallets/parachain-staking/src/types.rs
weichweich: I didn't find where we use only `return` and i think `returns` sounds better. But not sure... 😅 Added a dot in the end anyways. 2e35cbe8838996f08cbc4020913233440c965b2e

pallets/parachain-staking/src/lib.rs
weichweich: yeah converting with as is not that nice in case it's bigger than a u32. But since wasm is u32 only it should be fine. Felt fancy, removed it... f566feaf5fcbe8d4c32e3bd80ff735e22d9fcd7d

pallets/parachain-staking/src/types.rs
Diiaablo95: Sorry I did not mean that there are other `Return`, but that there are other places that do not use the third person, like some `Check`, for instance.

weichweich: Ah i think i got it! Fixed here -> 90cfd09 <- maybe?

Diiaablo95 (approved):
Good to go!

weichweich (approved):

PR #182, opened 2021-06-09
feat: set existential deposit to 0.1 KILT

weichweich:
Sorry @wischli you need to review again! 😂 Should have been 0.1 KILT = 100 milliKILT and not  0.01

## Checklist:


wischli (approved):
I also want to get rid of `DOLLARS` but can do this in the PR in which we finalize the parameters.

PR #186, opened 2021-06-10
feat: move westend spec into rust code

weichweich:
Using the python script is not nice for wilt and spiritnet since we load the genesis-accounts in the rust code and they get mixed up with the dev accounts.

## Checklist:


wischli (approved):
LGTM

PR #189, opened 2021-06-10
feat: update KILT pallets benchmarks that do not use DidOrigin

Diiaablo95:
## fixes KILTProtocol/ticket#1411
This PR basically reverts the old benchmarks for the ctype, delegation, and attestation pallets, at least for as long as they do not require DID authentication. Almost all of the benchmark code is taken from https://github.com/KILTprotocol/mashnet-node/pull/115 with the needed changes to make it work with the current version of the node.

## How to test:

Run `cargo test --all --all-features` to test the benchmarks.


pallets/delegation/src/benchmarking.rs
```diff
@@ -0,0 +1,287 @@
...
+use frame_system::RawOrigin;
+use sp_core::{offchain::KeyTypeId, sr25519};
+use sp_io::crypto::sr25519_generate;
+use sp_runtime::MultiSignature;
+use sp_std::{num::NonZeroU32, vec::Vec};
+
+use crate::*;
+
+const SEED: u32 = 0;
+const MAX_REVOCATIONS: u32 = 5;
```
wischli, line 31: I suppose we should add a parameter to the delegation config which sets the global max revocations in the future. We wouldn't even have to enforce them, e.g., by throwing if exceeding. Doesn't have to be part of this PR but I still wanted to raise this up to discussion.

PS: I know this code is pretty much what Albi and I wrote before.

wischli (approved):
LGTM. See question below for a potential improvement - now or in the future.

pallets/attestation/src/benchmarking.rs
```diff
@@ -0,0 +1,79 @@
...
+use delegation::{benchmarking::setup_delegations, Permissions};
+use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
+use frame_system::RawOrigin;
+use sp_core::sr25519;
+use sp_runtime::traits::Hash;
+use sp_std::num::NonZeroU32;
+
+use crate::*;
+
+const MAX_DEPTH: u32 = 10;
```
weichweich, line 28: We should have a max depth config in the Config trait and use that here

weichweich (commented):
I'm not 100% sure whether we need to set an upper limit here. Normally i would say yes.

pallets/attestation/src/benchmarking.rs
Diiaablo95: Added in https://github.com/KILTprotocol/mashnet-node/pull/189/commits/fbb964e3c59d5d67066fd66039854b463a3454f6

pallets/delegation/src/benchmarking.rs
Diiaablo95: Added in https://github.com/KILTprotocol/mashnet-node/pull/189/commits/fbb964e3c59d5d67066fd66039854b463a3454f6

pallets/attestation/src/lib.rs
```diff
@@ -230,6 +239,7 @@ pub mod pallet {
 			// the original attester
 			if attestation.attester != revoker {
 				let delegation_id = attestation.delegation_id.ok_or(Error::<T>::UnauthorizedRevocation)?;
+				let max_parent_checks = max_parent_checks.min(T::MaxParentChecks::get());
```
weichweich, line 242: maybe we should ensure the maximum in the beginning and fail?

pallets/delegation/src/lib.rs
```diff
@@ -321,6 +332,8 @@ pub mod pallet {
 
 			ensure!(root.owner == invoker, Error::<T>::UnauthorizedRevocation);
 
+			let max_children = max_children.min(T::MaxRevocations::get());
```
weichweich, line 335: same here

pallets/delegation/src/lib.rs
```diff
@@ -374,11 +389,15 @@ pub mod pallet {
 				Error::<T>::DelegationNotFound
 			);
 
+			let max_parent_checks = max_parent_checks.min(T::MaxParentChecks::get());
```
weichweich, line 392: here

pallets/delegation/src/lib.rs
```diff
@@ -374,11 +389,15 @@ pub mod pallet {
...
+
 			ensure!(
 				Self::is_delegating(&invoker, &delegation_id, max_parent_checks)?,
 				Error::<T>::UnauthorizedRevocation
 			);
 
+			let max_revocations = max_revocations.min(T::MaxRevocations::get());
```
weichweich, line 399: and here

weichweich (commented):
I'm not sure how to handle the limit. But i think we should rather fail then silently use the upper bound instead.

pallets/attestation/src/lib.rs
Diiaablo95: So we just remove the user-provided parameter altogether? It would make sense in that case.

weichweich: Yes we could do that. 🤔 It's probably the correct thing to do.

It's just cheaper for the user, because he doesn't have to pay the worst case. I think the extrinsic is quite expensive so maybe that makes a difference.

pallets/delegation/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/189/commits/0ddb538a40f37949049b516c3fa96c4f938e533b

pallets/attestation/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/189/commits/0ddb538a40f37949049b516c3fa96c4f938e533b

PR #193, opened 2021-06-14
feat: parachain staking 2.0

wischli:
### Short description
* changes reward calculation
* fixes reward-delegation attack vector
* adds automatically decreasing reward rates
* adds extrinsic to increase max collator stake
* adds extrinsic to decrease max collator stake
* remove unused `AtStake` storage
* add benchmarks for all variants of `on_initialize`
* fix missing storage updates after a delegator was replaced with another one with more delegated stake (KILTProtocol/ticket#1420)
* adds sudo extrinsic to remove a collator candidate
* includes runtime migration (thanks @Diiaablo95 for adding the `try-runtime` feature)

### Long description
**1. Fix attack vector of changing delegation after receiving reward**
A delegator can change their delegation after their collator has authored a block to a collator which has not authored a block yet in this round. Since Aura choses authorities on a round robin basis, it is easy to know the set of collators which hasn't authored a block in the current round. You could simply write a script to change the delegation. This only makes sense, if the fees paid for re-delegating are less than the expected rewards. This should only occur for high amounts of delegations. We haven't calculated this because our fees are still not final.
As of now, we are planning to allow a delegator to change their delegation only `MaxCollatorsPerDelegators` many times per round by adding a storage which stores the last round index in which a change was made. If the current round index equals the index in which the last change was made, we will throw an error. This still allows the delegator to theoretically snipe one more reward per round but on average, it shouldn't yield advantage. Consider an attacker to have delegated to Alice and after Alice authors a block, the attacked delegates to Bob which hasn't authored a block yet. In the next round, both Alice and Bob have equal probability to be block author before the other one (if at all). Thus, the attacker should not gain an advantage with less than random likelihood.

**2. Change reward calculation for collators and delegators**
Both collators and delegators should receive the reward rate set in the InflationInfo until the max staking rates are met:
* `collator reward per block = collator_stake * collator_reward_rate / blocks_per_year * number_of_active_candidates`
* `delegator reward per block = delegator_stake * delegator_reward_rate / blocks_per_year * number_of_active_candidate`

This differs from the current implementation where a collator's reward is independent of their own stake. Their stake was just used to get into the top n candidates. Moreover, a delegator could receive the full block reward of delegators if they are the only delegators for this collator. This enabled a delegator to receive the full delegator block reward by just staking a single KILT coin.

**3. Automatically decrease reward rate annually**
Add a check to on_initialize which compares the current block with `BLOCKS_PER_YEAR`. If a new year has started, both reward rates should decrease by a fixed percentage automatically. This requires adding an additional StorageValue which stores an integer in which the last change was made to disable multiple changes within the same year.

**4. Add extrinsic to increase/decrease max collator stake (sudo call)**
When decreasing, the difference of the current stake and the new max will automatically be made available without waiting a single block.

**5. See KILTProtocol/ticket#1420**

## Types companion

https://github.com/KILTprotocol/type-definitions/pull/8

## How to test
```
RUST_LOG=info cargo run -p kilt-parachain --features try-runtime -- try-runtime --chain staging --execution wasm --wasm-execution compiled on-runtime-upgrade live wss://kilt-peregrine-k8s.kilt.io
```

## TODO

- [x] Open companion PR with newly added types
- [x] Re-run benchmarks
- [x] Disallow `stake_{more, less}` with 0 value
- [x] Disallow `stake_{more, less}` when collator state is `IsLeaving`

## Checklist:


pallets/parachain-staking/src/inflation.rs
```diff
@@ -34,15 +33,18 @@ pub struct RewardRate {
 }
 
 /// Convert annual reward rate to per_block.
-fn annual_to_per_block(rate: Perquintill) -> Perquintill {
+fn annual_to_per_block<T: Config>(rate: Perquintill) -> Perquintill {
+	// FIXME: Should use T::BlocksPerYear::get() instead of YEARS
+	// let per_year: u64 = T::BlocksPerYear::get().into();
+	// rate / per_year
 	rate / YEARS
```
shawntabrizi, line 40:
```suggestion
fn annual_to_per_block<T: Config>(rate: Perquintill) -> Perquintill {
	// Need to make sure that `BlocksPerYear` is not more than `u64::max_value()`
	let per_year: u64 = T::BlocksPerYear::get().saturated_into::<u64>();
	rate / per_year
```

pallets/parachain-staking/src/inflation.rs
```diff
@@ -58,21 +60,30 @@ pub struct StakingInfo {
...
+		stake: BalanceOf<T>,
+		current_staking_rate: Perquintill,
+		authors_per_round: BalanceOf<T>,
+	) -> BalanceOf<T> {
+		// Perquintill automatically bounds to [0, 100]% in case staking_rate >
+		// self.max_rate
+		let reduction = Perquintill::from_rational(self.max_rate.deconstruct(), current_staking_rate.deconstruct());
 		// multiplication with perbill cannot overflow
-		staking_rate * rewards
+		let reward = self.reward_rate.per_block * stake * authors_per_round;
```
shawntabrizi, line 85: `stake * authors per round` could overflow though?

sounds unlikely, but not free

pallets/parachain-staking/src/inflation.rs
```diff
@@ -145,7 +141,7 @@ mod tests {
 		let rate = Perquintill::one();
 		assert!(almost_equal(
 			rate * 10_000_000_000u128,
```
Diiaablo95, line 143: Can we make this `10_000_000_000u128` a constant or a variable somehow instead of a magic constant?

Diiaablo95: Ah ok these are tests, nevermind. 😄

pallets/parachain-staking/src/lib.rs
```diff
@@ -432,29 +441,26 @@ pub mod pallet {
...
-					round.current,
-					collator_staked,
-					delegator_staked,
-				));
-				T::DbWeight::get().reads_writes(2, 1)
+				Self::deposit_event(Event::NewRound(round.first, round.current));
+				post_weight = post_weight.saturating_add(T::DbWeight::get().writes(1));
+			}
+			// check for InflationInfo update
+			if now > T::BlocksPerYear::get() {
```
Diiaablo95, line 459: We can maybe assign the result of the if-else expression to `post_weight` and return that?

pallets/parachain-staking/src/lib.rs
```diff
@@ -585,19 +598,14 @@ pub mod pallet {
...
 			<Round<T>>::put(round);
 			// Snapshot total stake
 
-			Pallet::<T>::deposit_event(Event::NewRound(
-				T::BlockNumber::zero(),
-				0u32,
-				collator_staked,
-				delegator_staked,
-			));
+			Pallet::<T>::deposit_event(Event::NewRound(T::BlockNumber::zero(), 0u32));
```
Diiaablo95, line 608: What happens if the staking pallet is added to the runtime at a later time, when the current block number is > 0? Is that even possible?

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +721,98 @@ pub mod pallet {
...
+			ensure!(new >= T::MinCollatorCandidateStk::get(), Error::<T>::CannotSetBelowMin);
+			let old = <MaxCollatorCandidateStk<T>>::get();
+			ensure!(new > old, Error::<T>::NotIncreasing);
+			<MaxCollatorCandidateStk<T>>::put(new);
+
+			Self::deposit_event(Event::MaxCandidateStakeChanged(old, new));
+			Ok(())
+		}
+
+		/// Increases the maximum amount a collator candidate can stake. The new
```
weichweich, line 754:
```suggestion
		/// Decreases the maximum amount a collator candidate can stake. The new
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +721,98 @@ pub mod pallet {
...
+		/// ---------
+		/// Weight: O(N) where N is `MaxSelectedCandidates` bounded by
+		/// `MaxCollatorCandidates`
+		/// - Reads: MaxCollatorCandidateStk, 2 * N * CollatorState,
+		///   CandidatePool
+		/// - Writes: MaxCollatorCandidateStk, N * CollatorState,
+		///   SelectedCandidates
+		/// # </weight>
+		#[pallet::weight(100_000)]
+		pub fn decrease_max_candidate_stake(origin: OriginFor<T>, new: BalanceOf<T>) -> DispatchResultWithPostInfo {
```
weichweich, line 775: same here, "decrease by" instead of "set to"?

pallets/parachain-staking/src/lib.rs
wischli: The genesis builder function is only invoked when we build the chain spec. Thus, when the parachain-staking pallet is added to the runtime later, the initial setup needs to be taken care of in the runtime migration.

pallets/parachain-staking/src/lib.rs
wischli: Should be fixed in 309251c.

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +721,98 @@ pub mod pallet {
...
+								<CollatorState<T>>::insert(&candidate.owner, state);
+
+								// update consumed weight
+								post_weight = post_weight.saturating_add(T::DbWeight::get().writes(2));
+							}
+						}
+						candidate
+					})
+					.collect();
+				*pool = OrderedSet::from(candidates);
```
weichweich, line 809: The pool is mutable and we are not planning on removing items? why not simply go over it in for loop and mutate the items?

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +727,98 @@ pub mod pallet {
...
+			});
+
+			// update candidates for next round
+			let (_num_collators, _num_delegators, _, _) = Self::select_top_candidates();
```
Diiaablo95, line 813: Do we even need to capture the return values here?

pallets/parachain-staking/src/lib.rs
weichweich: add these two traits to the set

```
impl<T: Ord> IndexMut<Range<usize>> for OrderedSet<T> {
	fn index_mut(&mut self, range: Range<usize>) -> &mut Self::Output {
		&mut self.0[range]
	}
}

impl<T: Ord> IndexMut<RangeFull> for OrderedSet<T> {
	fn index_mut(&mut self, range: RangeFull) -> &mut Self::Output {
		&mut self.0[range]
	}
}
```

than you can use `for mut candidate in pool[..].iter_mut() {...}`

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +721,98 @@ pub mod pallet {
...
+			// post_weight = post_weight.saturating_add()
```
weichweich, line 814: should this stay or should this go

runtimes/spiritnet/src/lib.rs
```diff
@@ -342,20 +342,22 @@ parameter_types! {
...
-	/// Max stake possible to be reserved to be collator candidate is 200_000
-	pub const MaxCollatorCandidateStk: Balance = 200_000 * KILT;
 	/// Minimum stake required to be reserved to be a delegator is 1000
 	pub const MinDelegatorStk: Balance = 1000 * KILT;
 	/// Maximum number of collator candidates
 	pub const MaxCollatorCandidates: u32 = 75;
 	/// Maximum number of concurrent requests to unlock unstaked balance
 	pub const MaxUnstakeRequests: u32 = 10;
+	/// Number of blocks per year based on 12s block time
+	pub const BlocksPerYear: BlockNumber = YEARS;
```
weichweich, line 360: is this really necessary? why not use `YEARS`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -1865,13 +1909,18 @@ pub mod pallet {
 			}
 			collators.sort();
 
+			<Total<T>>::mutate(|old| {
```
Diiaablo95, line 1912: I would rename `old` to something else, as we are using the parameter to update the total stake, and it could be misleading.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1803,6 +1831,21 @@ pub mod pallet {
 			Ok(())
 		}
 
+		fn delegator_is_kicked(delegation: StakeOf<T>, collator: &T::AccountId) -> DispatchResult {
```
Diiaablo95, line 1834: I think `delegation` can be a `&StakeOf<T>`, so that we avoid cloning when we want to use it.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1903,6 +1952,10 @@ pub mod pallet {
 						.saturating_sub(stake_to_remove.amount)
 						.saturating_add(stake.amount);
 					state.delegators = OrderedSet::from_sorted_set(delegators);
+
+					// update storage of kicked delegator
+					Self::delegator_is_kicked(stake_to_remove.clone(), &state.id)?;
```
Diiaablo95, line 1957: Should be possible to use `&stake_to_remove` without cloning.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2097,6 +2150,75 @@ pub mod pallet {
...
 
+		/// Annually reduce the reward rates for collators and delegators.
+		///
+		/// # <weight>
+		/// Weight: O(1)
+		/// - Reads: LastRewardReduction, InflationConfig
+		/// - Writes: LastRewardReduction, InflationConfig
+		/// # </weight>
+		fn adjust_reward_rates(now: T::BlockNumber) -> Weight {
+			let year = now / T::BlocksPerYear::get();
```
Diiaablo95, line 2161: Although unlikely, nothing prevents `BlocksPerYear` to be equal to 0.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2097,6 +2150,75 @@ pub mod pallet {
...
+
+		/// Checks whether a delegator can still delegate in this round, e.g.,
+		/// if they have not delegated MaxDelegationsPerRound many times
+		/// already in this round.
+		///
+		/// # <weight>
+		/// Weight: O(1)
+		/// - Reads: LastDelegation, Round
+		/// # </weight>
+		fn update_delegation_counter(delegator: &T::AccountId) -> Result<DelegationCounter, DispatchError> {
```
Diiaablo95, line 2201: The name `update` is not really correct here as all this function does is to check whether the delegator can still delegate.

```suggestion
		fn get_delegation_counter(delegator: &T::AccountId) -> Result<DelegationCounter, DispatchError> {
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -2116,40 +2238,65 @@ pub mod pallet {
...
-					if amount >= T::MinDelegatorStk::get() {
-						// Compare this delegator's stake with the total amount of
-						// delegated stake for this collator
-						// multiplication with perquintill cannot overflow
-						let percent = Perquintill::from_rational(amount, delegator_stake);
-						let due = percent * amt_due_delegators;
-						Self::do_reward(&owner, due);
+			// should always include state
+			if let Some(state) = <CollatorState<T>>::get(author.clone()) {
+				if state.stake >= T::MinCollatorStk::get() && state.total >= T::MinCollatorStk::get() {
```
weichweich, line 2272: do we need to check this here? This should be ensured in the extrinsics? If these preconditions are not met, what is the outcome here? At the moment the block will be produced without rewards. if we remove the check, the rewards are just less? Or can the function go to infinity if the total stake is to low?

weichweich (commented):
should we maybe replace all `YEARS` with `T::BlocksPerYear::get()`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -2116,40 +2238,65 @@ pub mod pallet {
...
+		/// blocks_per_year`. Since a collator can only author blocks every
+		/// `MaxSelectedCandidates` many rounds, we multiply the reward with
+		/// this number. As a result, a collator who has been in the set of
+		/// selected candidates, eventually receives `reward_rate * stake` after
+		/// one year.
+		///
+		/// However, if the current staking rate exceeds the max staking rate,
+		/// the reward will be reduced by `max_rate / current_rate`. E.g., if
+		/// the current rate is at 50% and the max rate at 40%, the reward is
+		/// reduced by 20% by multiplying the reward with 80%.
```
Diiaablo95, line 2260: I would just say that the reward is reduced by 20%.

```suggestion
		/// reduced by 20%.
```

Diiaablo95 (commented):
Sorry for the noise generated with single comments, I was not very familiar with this review feature 😄

runtimes/spiritnet/src/lib.rs
wischli: Turns out it can be removed when using `saturated_into` 🥳

pallets/parachain-staking/src/lib.rs
wischli: Yeah, the check can be removed.

pallets/parachain-staking/src/inflation.rs
```diff
@@ -195,63 +191,171 @@ mod tests {
...
+						current_staking_rate,
+						authors_per_round
+					) * years_u128,
+					Perquintill::from_percent(15) * 10_000 * DECIMALS,
+				);
+
+				// // Check delegator reward computation
+				// assert_eq!(inflation.delegator.compute_block_rewards::
+				// <Test>(0, 100_000), 0); assert!(
+				// 	almost_equal(
```
shawntabrizi, line 265: looks like this should be a test in the file

pallets/parachain-staking/src/lib.rs
```diff
@@ -110,12 +110,12 @@
 //!   rounds enable more frequent changes of the selected candidates, earlier
 //!   withdrawal from unstaking and earlier collator leaving. Requires sudo.
 //! - `join_candidates` - Join the set of collator candidates by staking at
-//!   least `MinCandidateStk` and at most `MaxCandidateStk`.
+//!   least `MinCandidateStk` and at most `MaxCollatorCandidateStk`.
```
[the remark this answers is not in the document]
shawntabrizi: agreed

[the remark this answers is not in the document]
wischli: Right now we abbreviated all usages of _Stake_ to _Stk_. I agree, we should probably switch to `Stake`.

pallets/parachain-staking/src/lib.rs
shawntabrizi: Events arent emitted in the genesis. There is code which explicitly does not add events on block 0

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +727,98 @@ pub mod pallet {
...
+		/// The dispatch origin must be Root.
+		///
+		/// Emits `MaxCandidateStakeChanged`.
+		///
+		/// # <weight>
+		/// Weight: O(1)
+		/// - Reads: [Origin Account], MaxCollatorCandidateStk
+		/// - Writes: Round
+		/// # </weight>
+		#[pallet::weight(100_000)]
```
shawntabrizi, line 742: needs weight

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +721,98 @@ pub mod pallet {
...
+		pub fn increase_max_candidate_stake(origin: OriginFor<T>, new: BalanceOf<T>) -> DispatchResult {
```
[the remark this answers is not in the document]
shawntabrizi: +1

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +727,98 @@ pub mod pallet {
...
+		/// - For each N, we read `CollatorState` from the storage.
+		/// ---------
+		/// Weight: O(N) where N is `MaxSelectedCandidates` bounded by
+		/// `MaxCollatorCandidates`
+		/// - Reads: MaxCollatorCandidateStk, 2 * N * CollatorState,
+		///   CandidatePool
+		/// - Writes: MaxCollatorCandidateStk, N * CollatorState,
+		///   SelectedCandidates
+		/// # </weight>
+		#[pallet::weight(100_000)]
```
shawntabrizi, line 774: weight

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +727,98 @@ pub mod pallet {
...
+		pub fn decrease_max_candidate_stake(origin: OriginFor<T>, new: BalanceOf<T>) -> DispatchResultWithPostInfo {
+			frame_system::ensure_root(origin)?;
+			ensure!(new >= T::MinCollatorCandidateStk::get(), Error::<T>::CannotSetBelowMin);
+
+			let old = <MaxCollatorCandidateStk<T>>::get();
+			ensure!(old > new, Error::<T>::NotDecreasing);
+
+			// iterate candidate pool and update all candidates with stake > max
+			let mut post_weight = T::DbWeight::get().reads(2);
+			<CandidatePool<T>>::mutate(|pool| {
```
shawntabrizi, line 784: You should start using bounded vec stuff here or you will need to migrate later

pallets/parachain-staking/src/lib.rs
shawntabrizi: T::BlocksPerYear::get().max(1)

pallets/parachain-staking/src/lib.rs
wischli: Yes, for post weight reduction.

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: Fixed in 11ecc03

pallets/parachain-staking/src/lib.rs
wischli: Thanks, fixed in 11ecc03

runtimes/spiritnet/src/lib.rs
wischli: Fixed in aa558d4

pallets/parachain-staking/src/lib.rs
wischli: Will do in a separate PR. Thanks for pointing that out!

pallets/parachain-staking/src/inflation.rs
```diff
@@ -34,15 +33,15 @@ pub struct RewardRate {
 }
 
 /// Convert annual reward rate to per_block.
-fn annual_to_per_block(rate: Perquintill) -> Perquintill {
-	rate / YEARS
+fn annual_to_per_block<T: Config>(rate: Perquintill) -> Perquintill {
```
Diiaablo95, line 36: I guess we don't need the generic `T` anymore in this function.

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +760,150 @@ pub mod pallet {
...
+			collator: <T::Lookup as StaticLookup>::Source,
+		) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
+			let collator = T::Lookup::lookup(collator)?;
+			let state = <CollatorState<T>>::get(&collator).ok_or(Error::<T>::CandidateNotFound)?;
+			let total_amount = state.total;
+
+			let mut candidates = <CandidatePool<T>>::get();
+			ensure!(
+				candidates.len().saturating_sub(1) as u32 >= T::MinSelectedCandidates::get(),
```
Diiaablo95, line 890: Maybe we can use `try_into()` to catch cases in which `usize` is `u64`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -1803,6 +1917,21 @@ pub mod pallet {
 			Ok(())
 		}
 
+		fn delegator_is_kicked(delegation: &StakeOf<T>, collator: &T::AccountId) -> DispatchResult {
```
Diiaablo95, line 1924: It would be clearer to name it something like `kick_delegator`.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2097,6 +2272,75 @@ pub mod pallet {
...
+
+			ensure!(
+				T::MaxDelegationsPerRound::get() > last_delegation.counter,
+				Error::<T>::ExceededDelegationsPerRound
+			);
+
+			Ok(DelegationCounter {
+				round: round.current,
+				counter: counter + 1,
+			})
```
Diiaablo95, line 2345: Saturating?

Diiaablo95 (approved):
Looks good! Just a few minor points 😄

pallets/parachain-staking/src/lib.rs
wischli: fixed in b03f3c6

pallets/parachain-staking/src/lib.rs
wischli: fixed in b03f3c6

pallets/parachain-staking/src/lib.rs
wischli: fixed in b03f3c6

pallets/parachain-staking/src/inflation.rs
wischli: fixed in b03f3c6

pallets/parachain-staking/src/lib.rs
```diff
@@ -432,33 +458,53 @@ pub mod pallet {
...
+			migration_consumed_weight
 		}
 	}
 
+	/// True if network has been upgraded to this version.
+	/// Storage version of the pallet.
+	///
+	/// This is set to v2.0.0 for new networks.
+	#[pallet::storage]
+	pub(crate) type StorageVersion<T: Config> = StorageValue<_, Releases, ValueQuery>;
```
weichweich, line 506: isn't there a build in pallet version?

weichweich (approved):
lgtm I thought there was a build in pallet version. But i couldn't find anything about it. so i guess no

PR #197, opened 2021-06-17
refactor: rename runtime to peregrine

weichweich:
## Checklist:


Diiaablo95 (approved):
LGTM! Finally, I was wondering when the name chain would happen! 🤌🤌🤌

weichweich:
> LGTM! Finally, I was wondering when the name chain would happen! 🤌🤌🤌

i also thought about rename the standalone to `mashnet-runtime`... 🤔

Diiaablo95:
... and it'll come the time for the repo main as well!

nodes/parachain/src/chain_spec/peregrine.rs
```diff
@@ -39,12 +39,12 @@ use crate::chain_spec::{get_account_id_from_seed, get_from_seed, get_properties,
...
 
 pub fn make_dev_spec(id: ParaId) -> Result<ChainSpec, String> {
-	let properties = get_properties("KILT", 15, 38);
+	let properties = get_properties("PILT", 15, 38);
 	let wasm = WASM_BINARY.ok_or("No WASM")?;
 
 	Ok(ChainSpec::from_genesis(
-		"KILT Local",
-		"kilt_parachain_local_testnet",
+		"KILT Peregrine Local",
```
wischli, line 46: This change has to be applied to the Polkadot Apps in order to connect it with the styling.

runtimes/peregrine/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["KILT <info@kilt.io>"]
 edition = "2018"
-name = "kilt-parachain-runtime"
+name = "peregrine-runtime"
```
wischli, line 4: This change might also effect the Polkadot Apps in regards to the types.

wischli (approved):
LGTM! But please test with the Polkadot Apps, if the styling and types are still matched properly. I could do this on Monday, if we can postpone the merge until then.

nodes/parachain/src/chain_spec/peregrine.rs
weichweich: This is only for local testing. So it's no problem if we break it

runtimes/peregrine/Cargo.toml
weichweich: just checked. Spiritnet runtime still works

PR #204, opened 2021-06-22
fix: power of TransferAccount

wischli:
## fixes KILTProtocol/ticket#1429
Right now, the `TransferAccount` can migrate balance of any user which does not have any balance locks (e.g. vesting or LockedUp balance). It should only be possible to do this for unowned accounts which are set up during the genesis build.

* Adds `UnownedAccount` storage during genesis build
* Once a user is migrated, their entry in the Storage Map is removed
* Disables migration for accounts who do not have an entry in the Storage Map

## Checklist:


weichweich (approved):
LGTM, but we need to update the spec before we merge

PR #207, opened 2021-06-23
feat: new did creation logic

Diiaablo95:
## fixes KILTProtocol/ticket#1368

This PR introduces mainly two new features:

- DID creation now requires a valid signature from the account whose DID is being created, meaning that if Alice is generating and signing a new DID creation operation, the signature over the operation must be generated with Alice's private key. Specifying an authentication key upon creation is not possible anymore, albeit the key can still be updated with update operations. In this way, we are binding DID identifiers to KILT accounts.
- DIDs now support ECDSA keys and signature which use the secp256k1 curve, adding compatibility with Ethereum accounts/signatures.

Minor changes include:

- Updates to the DID benchmarks that include the ECDSA signatures, which are anyway definitely slower and less efficient than Ed25519 and Sr25519 signatures.
- A new trait called `DidVerifiableIdentifier` that can be implemented from types that both identify a Kilt account and whose ownership can be verified via digital signatures.
- Changes to the `DidCallDispatched` event payload to match what is done in the [sudo pallet](https://github.com/paritytech/substrate/blob/master/frame/sudo/src/lib.rs#L156). The previous version was not showing up in the PolkadotJS apps in case the dispatch failed.

## How to test

Run `cargo test --all --all-features --all-targets`.

### Custom types

Companion PR: https://github.com/KILTprotocol/type-definitions/pull/9.

## Checklist:


pallets/did/src/lib.rs
```diff
@@ -126,12 +125,9 @@ pub mod pallet {
...
 		/// \[transaction signer, DID identifier\]
 		DidDeleted(AccountIdentifierOf<T>, DidIdentifierOf<T>),
-		/// A DID-authorized call has been successfully executed.
-		/// \[DID caller]
-		DidCallSuccess(DidIdentifierOf<T>),
-		/// A DID-authorized call has failed to execute.
-		/// \[DID caller, error]
-		DidCallFailure(DidIdentifierOf<T>, DispatchError),
+		/// A DID-authorized call has been executed.
+		/// \[DID caller, dispatch result]
```
wischli, line 129:
```suggestion
		/// \[DID caller, dispatch result\]
```

wischli (approved):
One tiny remark, LGTM!

pallets/did/src/lib.rs
Diiaablo95: Fixed in https://github.com/KILTprotocol/mashnet-node/pull/207/commits/2155d5e2496391fa181646e89343cee9d4f1c69e

weichweich (approved):

pallets/did/src/did_details.rs
```diff
@@ -17,49 +17,60 @@
...
 			}
 			// Follows same process as above, but using a Sr25519 instead
 			DidVerificationKey::Sr25519(public_key) => {
 				if let DidSignature::Sr25519(sig) = signature {
-					Ok(sig.verify(payload, public_key))
+					ensure!(sig.verify(payload, public_key), SignatureError::InvalidSignature);
+					Ok(())
+				} else {
+					Err(SignatureError::InvalidSignatureFormat)
+				}
```
weichweich, line 62: That looks weird!

pallets/did/src/did_details.rs
```diff
@@ -17,49 +17,60 @@
...
 }
 
 impl DidVerificationKey {
 	/// Verify a DID signature using one of the DID keys.
-	pub fn verify_signature(&self, payload: &Payload, signature: &DidSignature) -> Result<bool, SignatureError> {
+	pub fn verify_signature(&self, payload: &Payload, signature: &DidSignature) -> Result<(), SignatureError> {
 		match self {
 			DidVerificationKey::Ed25519(public_key) => {
 				// Try to re-create a Signature value or throw an error if raw value is invalid
 				if let DidSignature::Ed25519(sig) = signature {
```
weichweich, line 46: why not match over both? self and signature??

weichweich (changes requested):

pallets/did/src/did_details.rs
Diiaablo95: Nice catch! There you go sir! 👴🏻  -> https://github.com/KILTprotocol/mashnet-node/pull/207/commits/bbb0f21acc7fffabb28b7848d891b015a3668f64

weichweich (approved):
LGTM

PR #208, opened 2021-06-23
chore: upgrade to 0.9.7

weichweich:
## Checklist:


wischli (changes requested):
Should use the v0.9.7 polkadot image which is already available and used on Westend and Kusama.

scripts/spec.sh
```diff
@@ -13,7 +13,7 @@ cp target/release/kilt-parachain $TMP_DIR/kilt-parachain
 # cargo build --release -p kilt-parachain --features fast-gov
 # cp target/release/kilt-parachain $TMP_DIR/kilt-parachain-fast-gov
 
-RELAY_CHAIN_IMG=parity/polkadot:v0.9.5
+RELAY_CHAIN_IMG=parity/polkadot:v0.9.6
```
wischli, line 16: See Polkadot [v0.9.7](https://github.com/paritytech/polkadot/releases/tag/v0.9.7) release and the [Westend](https://github.com/paritytech/polkadot/blob/master/runtime/westend/src/lib.rs#L123) as well as [Kusama](https://github.com/paritytech/polkadot/blob/master/runtime/kusama/src/lib.rs#L124) runtimes.

```suggestion
RELAY_CHAIN_IMG=parity/polkadot:v0.9.7
```

wischli (approved):

PR #211, opened 2021-06-28
fix: replace cast with saturated_into

Diiaablo95:
Some places have a `usize as u32` casts which, technically, might fail on 64-bit architectures (e.g., native runtime execution). This PR tries to address all the places where `usize` is either cast to `u32` or `u64` and where it could potentially fail, leading to runtime panics.
It most likely won't be merged before the staking PR is merged.


wischli (approved):
LGTM. As mentioned, let's not merge this PR before #206 (which will happen once the benchmarks are done).

PR #216, opened 2021-06-30
fix: delegation counter

wischli:
* fixes incorrect check for delegations per round
* replaces some unchecked conversions

## Checklist:


Diiaablo95 (approved):
LGTM! As I said, I am not 100% sure we NEED to saturate into `usize` from `u32`, as it would make sense only if the target architecture is, e.g., 16-bit. But I guess it doesn't hurt anyways to saturate.

wischli:
> As I said, I am not 100% sure we NEED to saturate into `usize` from `u32`, as it would make sense only if the target architecture is, e.g., 16-bit. But I guess it doesn't hurt anyways to saturate.

Completely agree.

PR #219, opened 2021-07-05
fix: sort candidates and delegations by staked amount

wischli:
## fixes KILTProtocol/ticket#1443
* Replaces sorting of CandidatePool and delegations by account id with sorting by (total) stake from greatest to lowest
  * Prioritizes existing entries of the OrderedSet over new entries with the same stake, e.g.

```
[{ owner: 3, amount: 100}, {owner: 2, amount: 99}] + {owner: 1, amount: 100}
= [{ owner: 3, amount: 100}, {owner: 1, amount: 100}], {owner: 2, amount: 99}]
```

  * Downside: Increases execution duration for all {insert | update | delete} operations on the `OrderedSet`, e.g. if a candidate or collator {joins | leaves, stakes more|less}
  * Upside: Does not introduce new storage, e.g., we could have added another vector which just stores account ids to check for uniqueness. However, then we still would have to increase execution time on the mentioned operations to keep the vector with stake amounts always sorted from greatest to lowest

## How to test:
* Add candidates or delegations from an account with lower account id and the same amount of stake as other candidates/delegations with higher account id(s)

### ToDo

- [x] Please test this locally @rflechtner
- [x] Write migration for Peregrine networks (e.g. just re-execute `select_top_candidates()`)

## Checklist:


pallets/parachain-staking/src/benchmarking.rs
```diff
@@ -44,7 +44,7 @@ fn setup_collator_candidates<T: Config>(
 	let collators: Vec<T::AccountId> = (current_collator_count..num_candidates)
 		.map(|i| account("collator", i.saturated_into::<u32>(), COLLATOR_ACCOUNT_SEED))
 		.collect();
-	let amount: T::CurrencyBalance = default_amount.unwrap_or_else(T::MinCollatorCandidateStake::get);
```
wischli, line 47: I wonder why this line (`T::MinCollatorCandidateStake::get`) was not caught 🥲

weichweich: should be the same? Actually clippy should complain now?

pallets/parachain-staking/src/lib.rs
```diff
@@ -2042,53 +2058,25 @@ pub mod pallet {
 			log::trace!("{} Candidates for {} Collator seats", candidates.len(), top_n);
 
 			// Order candidates by their total stake (greatest to least)
-			candidates.sort_by(|a, b| b.amount.cmp(&a.amount));
+			//
+			// NOTE: Resorting is required if a collator's total stake just changed before
+			// updating the selected candidates.
+			candidates.sort_by(|a, b| b.cmp(&a));
```
weichweich, line 2064: Ought to work:

```suggestion
			candidates.sort();
```

pallets/parachain-staking/src/set.rs
```diff
@@ -124,6 +126,28 @@ impl<T: Ord> OrderedSet<T> {
 		self.0.binary_search(value)
 	}
 
+	// iterative search from left to right which is assumed to be sorted from
+	// greatest to least
+	pub fn linear_search(&self, value: &T) -> Result<usize, usize> {
```
weichweich, line 131: why not binary search if it's sorted?

pallets/parachain-staking/src/set.rs
```diff
@@ -42,7 +43,7 @@ impl<T: Ord> OrderedSet<T> {
 	///
 	/// `v` will be sorted and dedup first.
 	pub fn from(mut v: Vec<T>) -> Self {
-		v.sort();
+		v.sort_by(|a, b| b.cmp(a));
```
weichweich, line 46: Would be nice to have a comment that explicitly states that reverse ordering is wanted here. Also the OrderedSet gets less generic by that.

pallets/parachain-staking/src/set.rs
```diff
@@ -124,6 +126,28 @@ impl<T: Ord> OrderedSet<T> {
...
+			if v.cmp(value) == Ordering::Equal {
+				return Ok(i);
+			// store index if v is bigger (since we assume order from max to
+			// min) but keep checking for Ordering::Equal
+			} else if v.cmp(value) == Ordering::Less && loc == size {
+				// insert after current element
+				loc = i;
+			}
+			i = i.saturating_add(1);
+		}
```
weichweich, line 147: No need to keep track of i manually! You could use [enumerate](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate).

Something like that should work too?

```suggestion
		self.0.iter().enumerate().find_map(|(i, elem)| {
			// prevent to have same items
			match v.cmp(value) {
			Ordering::Equal => Some(Ok(i))
			// store index if v is bigger (since we assume order from max to
			// min) but keep checking for Ordering::Equal
			Ordering::Less => {
				// insert after current element
				Some(Err(i))
			},
			_ => None,
		})
```

I don't really get the `else if` part. Wouldn't we want to return there?

pallets/parachain-staking/src/set.rs
```diff
@@ -160,45 +184,50 @@ impl<T: Ord> OrderedSet<T> {
...
 	}
+
+	/// Sort from greatest to lowest
+	pub fn sort_greatest_to_lowest(&mut self) {
+		self.0.sort_by(|a, b| b.cmp(a));
+	}
 }
 
-impl<T: Ord> From<Vec<T>> for OrderedSet<T> {
+impl<T: Ord + Debug> From<Vec<T>> for OrderedSet<T> {
```
weichweich, line 194: why do we need Debug here?

pallets/parachain-staking/src/set.rs
```diff
@@ -297,4 +328,57 @@ mod tests {
 		set.clear();
 		assert_eq!(set, OrderedSet::new());
 	}
+
+	#[test]
+	fn linear_search() {
```
weichweich, line 333: I would rather test `linear_search` directly. Looks odd to test `linear_search` and the test doesn't contain a single reference to `linear_search` 😜 .

rflechtner:
As far as I can tell, this works great! When filling the candidate pool with multiple candidates with the same stake, parachainStaking.candidatePool() remains in the order in which they joined. The active candidates are selected from the first 16 candidates. The order is maintained even when individual candidates stake more or less (with the exception of the candidate who's rank has changed based on total stake). This is also true for delegations.
You can tunnel port 2001 from the rust server and have a look yourself:
- https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A2001#/explorer/query/0x9946bb04e55659c0d8a5fd9446612923edc97358488e5de3595353c5d157e3ec
- https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A2001#/explorer/query/0xeadae4c834d9284767c982e202c8eb7d69eff82078237f3eef88054c32a64ee9
- https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A2001#/explorer/query/0x868d927a37db050fd21d1d7244d9b911eba8e903d4b08b2968dd0f69f7647ba5
- https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A2001#/explorer/query/0xd0fc193e9841fa5eff3394be027181b821ffc7b50947ec0e192b710a7832ce7d
- https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A2001#/explorer/query/0xe0d1c229edd83c5f05e4b1085c25ce003b7000379023e30ba057a699a1509f7e

pallets/parachain-staking/src/set.rs
wischli: Have you checked the comparator? When you do binary search, it searches by amount first. However, we search for the account id half of the time, e.g. when changing the amount at stake or just checking for membership.

Thus, it does not work with binary search as we have 2 search parameters. We need to iterate over the whole list.

pallets/parachain-staking/src/lib.rs
wischli: This does not yield the reverse sorting that we require, e.g., it sorts from lowest to greatest.

pallets/parachain-staking/src/set.rs
wischli: Agree on the comment. I also agree with the OrderedSet getting less generic in general but not for this specific line of code.

pallets/parachain-staking/src/set.rs
wischli: > I don't really get the `else if` part. Wouldn't we want to return there?

I had that in the beginning but this does not work unfortunately because then we do not check whether the account id is already in the set. Thus, we end up with account ids existing more than once in the set.

Check the following example from the OrderedSet test `linear_search` with your suggestion, e.g., returning in the `else if` part:

```
OrderedSet::from_sorted(vec![
 { owner: 1, amount: 100 },
 { owner: 3, amount: 90 },
 { owner: 5, amount: 80 },
 { owner: 2, amount: 75 },
 { owner: 7, amount: 70 },
 { owner: 9, amount: 60 }
]).linear_search({ owner: 2, amount: 90}) = Some(Err(2))
```

However, it should return `Some(Ok(3))`, change the value and then resort.

pallets/parachain-staking/src/set.rs
wischli: It can be removed again, I needed it for debugging bugs.

pallets/parachain-staking/src/set.rs
wischli: Should be fixed in 211fabd

pallets/parachain-staking/src/set.rs
wischli: Should be fixed in 211fabd.

pallets/parachain-staking/src/benchmarking.rs
```diff
@@ -192,7 +192,7 @@ benchmarks! {
 	}: _(RawOrigin::Root,  T::Lookup::unlookup(candidate.clone()))
 	verify {
 		let candidates = CandidatePool::<T>::get();
-		assert!(candidates.binary_search_by(|other| other.owner.cmp(&candidate)).is_err())
+		assert!(!candidates.into_vec().iter().any(|other| other.owner == candidate));
```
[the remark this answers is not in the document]
wischli: All occurences fixed in 795fbdc.

pallets/parachain-staking/src/benchmarking.rs
wischli: > should be the same? Actually clippy should complain now?

Why should clippy complain now? And shouldn't `T:MinCollatorCandidateStake:get` without the parenthesis NOT work?

pallets/parachain-staking/src/set.rs
wischli: True. I renamed it and added specific tests for `linear_search` in e164d7d.

pallets/parachain-staking/src/benchmarking.rs
wischli: As discussed: Reverted in 0cb2746

pallets/parachain-staking/src/set.rs
wischli: Fixed in c11ad27

pallets/parachain-staking/src/set.rs
```diff
@@ -38,18 +38,20 @@ impl<T: Ord> OrderedSet<T> {
...
 	}
 
-	/// Create a set from a `Vec`.
+	/// Create an ordered set from a `Vec`.
 	///
-	/// `v` will be sorted and dedup first.
+	/// The vector will be sorted reversily (from greatest to lowest) and
+	/// deduped first.
 	pub fn from(mut v: Vec<T>) -> Self {
-		v.sort();
```
Diiaablo95, line 45: I would perhaps rename this to `from_sorted_vec`, as it might not be clear whether the vec is assumed to be sorted or if it is sorted internally, even though there are comments. But I don't have a strong argument against keeping it like this.

pallets/parachain-staking/src/lib.rs
Diiaablo95, line 2064: Suggesting changes from the only clippy warning:

```suggestion
			candidates.sort_by(|a, b| b.cmp(a));
```

Diiaablo95 (approved):
No major objections, as I found most of the things also Albi pointed out already. Overall, I like the new comparator 👍

pallets/parachain-staking/src/set.rs
wischli: The input vector does not need to be sorted but will be sorted when creating the set?!

pallets/parachain-staking/src/lib.rs
weichweich: Didn't see that it's a reverse sort.

pallets/parachain-staking/src/set.rs
Diiaablo95: Right... 😅

weichweich (approved):

PR #222, opened 2021-07-08
feat: add force_new_round

wischli:
## fixes KILTProtocol/ticket#1446
* Adds sudo extrinsic to force a new round
  * Added boolean StorageValue `ForceNewRound` which will automatically trigger a new round + session in the next block
  * Increases weight of `should_end_session` by 1 read per call and if `ForceNewRound` was set, 2 more writes and 1 more read are executed
* Improves `force_remove_new_candidate` by adding the candidate to `Session::DisabledValidators`
* Fixes minor issue for checking `MaxUnstakeRequests` (was off by 1)
* Fixes minor issue when a candidate which is leaving calls `join_candidates`: This threw `CandidateExists` when it should throw `AlreadyLeaving`

## TODO

- [x] Add benchmark for force_new_round

## Checklist:


pallets/parachain-staking/src/lib.rs
```diff
@@ -2284,6 +2324,20 @@ pub mod pallet {
...
+				.into_iter()
+				.enumerate()
+				.find_map(|(i, id): (usize, T::ValidatorId)| {
+					if <T as pallet_session::Config>::ValidatorIdOf::convert(collator.clone()) == Some(id) {
+						Some(i)
+					} else {
+						None
+					}
+				}) {
+				pallet_session::Pallet::<T>::disable_index(index);
```
wischli, line 2338: Unfortunately this does not seem to prevent this collator from being able to produce a block. I tried it out locally twice, and the kicked collator still produced blocks. 

I asked in Element and did not receive any answer yet. There is an [open issue for this with Babe instead of AuRa](https://github.com/paritytech/substrate/issues/8004).

pallets/parachain-staking/src/lib.rs
```diff
@@ -2550,7 +2606,20 @@ pub mod pallet {
 
 	impl<T: Config> ShouldEndSession<T::BlockNumber> for Pallet<T> {
 		fn should_end_session(now: T::BlockNumber) -> bool {
-			<Round<T>>::get().should_update(now)
+			let mut round = <Round<T>>::get();
+			// always update when a new round should start
+			if round.should_update(now) {
+				true
+			} else if <ForceNewRound<T>>::get() {
```
wischli, line 2613: Maybe we should register extra weight unchecked here?
```rust
frame_system::Pallet::<T>::register_extra_weight_unchecked(
  T::DbWeight::get().reads_writes(1, 2),
  DispatchClass::Mandatory,
);
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -1844,14 +1879,19 @@ pub mod pallet {
...
+		/// Check whether an account is currently a collator candidate and
+		/// whether their state is CollatorStatus::Active.
 		///
 		/// # <weight>
 		/// Weight: O(1)
 		/// - Reads: CollatorState
 		/// # </weight>
-		pub fn is_candidate(acc: &T::AccountId) -> bool {
-			<CollatorState<T>>::get(acc).is_some()
+		pub fn is_candidate_and_active(acc: &T::AccountId) -> (bool, bool) {
```
weichweich, line 1889: could also be `is_active_candidate` and return an `Option<bool>`. I guess this would represent the exists or not and if it exists is active or not better.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2284,6 +2324,20 @@ pub mod pallet {
 			// prepare unstaking of collator candidate
 			Self::prep_unstake_exit_queue(&state.id, state.stake);
 
+			// disable validator for next session if they were in the set of validators
+			if let Some(index) = pallet_session::Pallet::<T>::validators()
+				.into_iter()
+				.enumerate()
+				.find_map(|(i, id): (usize, T::ValidatorId)| {
```
weichweich, line 2331: Why do we need types here?

```suggestion
				.find_map(|(i, id)| {
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -2284,6 +2324,20 @@ pub mod pallet {
 			// prepare unstaking of collator candidate
 			Self::prep_unstake_exit_queue(&state.id, state.stake);
 
+			// disable validator for next session if they were in the set of validators
+			if let Some(index) = pallet_session::Pallet::<T>::validators()
```
weichweich, line 2328: I would maybe not use the `let if` here but rather add `.map(pallet_session::Pallet::<T>::disable_index)` to the end of the chain

pallets/parachain-staking/src/lib.rs
weichweich: maybe add a `FIXME: ...` in the code?

pallets/parachain-staking/src/lib.rs
```diff
@@ -2468,11 +2522,13 @@ pub mod pallet {
...
 		/// block author bounded by `MaxDelegatorsPerCollator`.
 		/// - Reads: CollatorState, Total, Balance, InflationConfig,
-		///   MaxSelectedCandidates
-		/// - Writes: D * Balance
+		///   MaxSelectedCandidates, Validators, DisabledValidators
+		/// - Writes: (D + 1) * Balance
 		/// # </weight>
 		fn note_author(author: T::AccountId) {
-			// should always include state
+			log::info!("current author {:?}", pallet_authorship::Pallet::<T>::author());
```
weichweich, line 2529: we log on each block? Not sure if we want to have that as a `debug`. 🤔 Normally we log very few things...

pallets/parachain-staking/src/lib.rs
weichweich: yes we should!

pallets/parachain-staking/src/lib.rs
wischli: Good catch, this was used to debug but the debug logs spammed my terminal too much.

pallets/parachain-staking/src/lib.rs
```diff
@@ -672,13 +676,42 @@ pub mod pallet {
...
+		/// - Writes: ForceNewRound
+		/// # </weight>
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_inflation())]
+		pub fn force_new_round(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+
+			frame_system::Pallet::<T>::register_extra_weight_unchecked(
+				T::DbWeight::get().reads_writes(1, 1),
+				DispatchClass::Mandatory,
+			);
```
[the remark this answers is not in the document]
wischli: This is a deprecate remainder from an earlier version, thanks.

pallets/parachain-staking/src/lib.rs
Diiaablo95: Why is it 1 read only? Don't we also always read from `Round<T>`? Does that come "for free"?

pallets/parachain-staking/src/lib.rs
```diff
@@ -672,13 +676,42 @@ pub mod pallet {
...
+		/// # <weight>
+		/// Weight: O(1)
+		/// - Reads: [Origin Account]
+		/// - Writes: ForceNewRound
+		/// # </weight>
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_inflation())]
+		pub fn force_new_round(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+
+			frame_system::Pallet::<T>::register_extra_weight_unchecked(
```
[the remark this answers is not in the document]
Diiaablo95: Mhh, just saw Albi's comment on this one as well. You can ignore this.

pallets/parachain-staking/src/lib.rs
wischli: Much more elegant, thanks. See e547d10.

pallets/parachain-staking/src/lib.rs
wischli: Fixed in e547d10

pallets/parachain-staking/src/lib.rs
wischli: Thanks, much more elegant. Fixed in e547d10

pallets/parachain-staking/src/lib.rs
wischli: This was exemplary. See ac6a011.

weichweich (approved):

PR #224, opened 2021-07-09
docs: make pallet description consistent

wischli:
## fixes KILTProtocol/ticket#1448
* Makes pallet documentation consistent based on the `kilt-launch` and `parachain-staking` pallets.

## ToDo

- [x] CTypes
- [x] Attestations
- [x] Delegations
- [x] DIDs

## Checklist:


Diiaablo95:
@wischli I cannot add you to the reviewers for some reasons, probably because you opened the PR.

Diiaablo95:
P.S. I also noticed that something was wrong in the weight calculation for DID-proxied calls, so that might be worth looking at as well.

wischli:
> P.S. I also noticed that something was wrong in the weight calculation for DID-proxied calls, so that might be worth looking at as well.

Can you please open an issue for that so we have it in the backlog?

Diiaablo95:
I addressed it in this PR actually. Should I still make a ticket for that?

wischli:
> I addressed it in this PR actually. Should I still make a ticket for that?

Naw, it's fine then.

Diiaablo95 (approved):

Diiaablo95:
@weichweich I guess your review would be somewhat valuable, as you are the one external to the documentation improvement process.

pallets/did/src/lib.rs
```diff
@@ -285,34 +392,86 @@ pub mod pallet {
...
+		/// operation has been signed by the DID subject using the
+		/// authentication key currently stored on chain.
+		///
+		/// The referenced DID identifier must be present on chain before the
+		/// update operation is evaluated.
+		///
+		/// Any new key specified is added to the set of public keys, and a
+		/// reference to it is updated in the relative field, i.e.,
+		/// authentication, key agreement, attestation or delegation.
+		/// As there can always at most one active authentication,
```
wischli, line 405:
```suggestion
		/// As there can always be at most one active authentication,
```

pallets/did/src/lib.rs
```diff
@@ -285,34 +392,86 @@ pub mod pallet {
...
+		/// removed from the set of public keys. The only exception is
+		/// represented by *key agreement keys* and *attestation keys*. For the
+		/// former, new keys added via update operations are always added to the
+		/// set of public keys, and their removal happens by including the ID of
+		/// the keys to remove in the `public_keys_to_remove` set in subsequent
+		/// update operations. In the latter case, the old key is replaced from
+		/// the `attestation_key` field by the new one but it is not removed
+		/// from the set of public keys. This means that the key can still be
+		/// retrieved (and used) by other parties to validate the signature of
+		/// attestations issued with thay key, but the issuance of new
```
wischli, line 417:
```suggestion
		/// attestations issued with that key, but the issuance of new
```

wischli (commented):
Two minor typos.

PR #225, opened 2021-07-12
feat: build wasm using SRTool

weichweich:
## Checklist:


.gitlab-ci.yml
```diff
@@ -64,3 +64,23 @@ build:
...
+build-wasm:
+  image:
+    name: paritytech/srtool:1.53.0
+    entrypoint: [""]
+  stage: build
+  only:
+    - develop
+    - master
+  script:
+    - export PACKAGE=peregrine-runtime
```
wischli, line 77: I suppose we also want the `spiritnet-runtime`?

.maintain/reset-spec/app.py
```diff
@@ -108,44 +122,46 @@ def make_native(docker_img, out_file, chain, runtime):
     tmp_dir = os.path.join("/tmp/", str(uuid.uuid1()))
     os.mkdir(tmp_dir)
 
-    if args.westend:
```
Diiaablo95, line 111: We still have the `--westend` option but we don't check anymore for it?

weichweich: good catch! Fixed in 31fa889

.gitlab-ci.yml
weichweich: fixed in d104b68

PR #226, opened 2021-07-13
feat: delegation pallet storage refactor

Diiaablo95:
## fixes KILTProtocol/ticket#1478

This PR improves the efficiency of the delegation pallet's storage so that for most of the operations less DB reads and writes are performed. The base concept is the following:

- All nodes (root and not) are considered the same, as opposite to before where a root node was not really a node. This means that they are all under the same `DelegationNodes` storage entry.
- A second storage entry, `DelegationHierarchies`, which maps from a root node ID to the details of a hierarchy, which at the moment only includes the CTYPE hash.
- The extrinsics `create_root` has been renamed to `create_hierarchy`.
- Unit tests and benchmarks have been updated accordingly. Same goes for the parts of the attestation's pallet  that needed to be updated.
- Added storage migrations and a potentially new approach to better handle consecutive and sequential storage migrations.

> 21.07.2021: A bug was discovered that revocation status of the parent node is not checked upon delegation creation. This PR is updated to mitigate such bug -> https://github.com/KILTprotocol/ticket/issues/1476 and fixed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/1e6b4361d5c22400c2424cf02e5c1d8f48221d41.
> 21.07.2021: After a more careful analysis, we found out that by now `revoke_root` and `revoke_delegation` are basically equivalent. So we have decided to get rid of `revoke_root` -> https://github.com/KILTprotocol/ticket/issues/1478 and addressed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/11bcc2d8cda5c55dce3f38e43d5f25392b886286.

## How to test:

Run `cargo test --all --all-features --all-targets`. Weights for the delegation pallets can also be compared with the current `develop` version to get an idea of the changed efficiency of the pallet.

For a dry run of migration on the live peregrine parachain, run

```
cargo run --release -p kilt-parachain --features try-runtime -- try-runtime --url wss://kilt-peregrine-k8s.kilt.io --block-at 0xe8be760e8b01b585c5aa858ec3a0a406d152c961d70a1b5c99b2ba21237cadda --dev --wasm-execution compiled --execution wasm on-runtime-upgrade live
```

### Note

**We decided to halt merging this PR until we are sure we can merge the updated SDK as well, so we avoid getting the new node version in the security audit but the SDK does not include the changes.**

### Custom types

https://github.com/KILTprotocol/type-definitions/pull/10.

## Checklist:

- [] This PR does not introduce new custom types


weichweich:
pls merge develop! 😁

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum DelegationMigrationError {
+	AlreadyLatest,
+	MigrationResultInconsistent,
+}
+
+// The "manager" of the pallet's storage migrations. It contains a vector of
+// version migrations each of which corresponds to a new runtime upgrade to run,
```
wischli, line 47: Shouldn't it be `of which each`?
```suggestion
// version migrations of which each corresponds to a new runtime upgrade to run,
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+// sequencially. It interacts with the `LastUpgradeVersion` of the pallet's
```
wischli, line 48: Typo
```suggestion
// sequentially. It interacts with the `LastUpgradeVersion` of the pallet's
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	}
+}
+
+mod v0 {
+	use super::*;
+	// Migrator for the first actual pallet's migration. It migrates the Roots,
+	// Delegations, and Children storage entries to a simpler HierarchyInfos,
+	// DelegationNodes, while maintaining the original information an all
+	// parent-child links.
+	pub struct V0Migrator();
```
wischli, line 134: Can't we just call this `Migrator`? It's already versioned by the module.
```suggestion
	pub struct Migrator();
```

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -52,87 +57,158 @@ impl Default for Permissions {
...
-	/// The hash of the CTYPE that delegated attesters within this trust
-	/// hierarchy can attest.
-	pub ctype_hash: CtypeHashOf<T>,
-	/// The identifier of the root owner.
-	pub owner: DelegatorIdOf<T>,
-	/// The flag indicating whether the root has been revoked or not.
-	pub revoked: bool,
-}
+// The old types.
+pub(crate) mod v0 {
```
weichweich, line 61: maybe we could move those definitions into their own file? Looks also better when there will be more versions

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -52,87 +57,158 @@ impl Default for Permissions {
...
-	///   allowed to perform
-	pub fn new_root_child(root_id: DelegationNodeIdOf<T>, owner: DelegatorIdOf<T>, permissions: Permissions) -> Self {
-		DelegationNode {
-			root_id,
-			owner,
-			permissions,
-			revoked: false,
-			parent: None,
+// The new types.
+pub(crate) mod v1 {
```
weichweich, line 154: When the old types are in their own file i would just have the current types in the root? I think that's just a bit more straight forward. You don't have to think about which version is the latest etc.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+
+// The "manager" of the pallet's storage migrations. It contains a vector of
+// version migrations each of which corresponds to a new runtime upgrade to run,
+// sequencially. It interacts with the `LastUpgradeVersion` of the pallet's
+// storage.
+//
+// When a new runtime upgrade needs to be added, a new component implementing
+// the `VersionMigrator` trait must be defined and added at the end of the
+// `migrations` vector, so that it is always executed after all
+// previous migrations have taken place.
```
wischli, line 54: Please use triple slashes for the entire comment 😊
```suggestion
/// The "manager" of the pallet's storage migrations. It contains a vector of
/// version migrations each of which corresponds to a new runtime upgrade to run,
/// sequencially. It interacts with the `LastUpgradeVersion` of the pallet's
/// storage.
///
/// When a new runtime upgrade needs to be added, a new component implementing
/// the `VersionMigrator` trait must be defined and added at the end of the
/// `migrations` vector, so that it is always executed after all
/// previous migrations have taken place.
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+pub(crate) struct StorageMigrator<T: Config> {
+	// The vector of version migrators.
+	migrations: Vec<Box<dyn VersionMigrator<T>>>,
```
wischli, line 57: Maybe we could use `BoundedVec` instead of `Vec<Box<...>>` here?! I think we can somehow get rid of the Box, which we should, if possible?!

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	pub(crate) fn pre_migration(&self) -> Result<(), DelegationMigrationError> {
+		ensure!(
+			LastUpgradeVersion::<T>::get() <= migrations::LATEST_UPGRADEABLE_VERSION,
+			DelegationMigrationError::AlreadyLatest
+		);
+
+		Ok(())
+	}
+
+	// It retrieves the latest version deployed on chain, and sequencially applies
```
wischli, line 79: Typo again
```suggestion
	// It retrieves the latest version deployed on chain, and sequentially applies
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	// any runtime migration until the state is update to the latest version.
```
wischli, line 80: Typo
```suggestion
	// any runtime migration until the state is updated to the latest version.
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	// It provides a test feature that panics whenver something goes wrong for one
```
wischli, line 81: Typo. I can recommend the VS Code plugin "Spell Checker" (recommended to me once by Albi).
```suggestion
	// It provides a test feature that panics whenever something goes wrong for one
```

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -52,87 +57,158 @@ impl Default for Permissions {
...
+
+		pub fn new_node(
+			hierarchy_root_id: DelegationNodeIdOf<T>,
+			parent: DelegationNodeIdOf<T>,
+			details: DelegationDetails<T>,
+		) -> Self {
+			let mut new_node = Self::new_root_node(hierarchy_root_id, details);
+			new_node.parent = Some(parent);
+
+			new_node
```
weichweich, line 183: i would not call the `new_root_node` method here but also just create the struct. This makes a simple think look more complicated. As someone who never saw the code before I'm unsure what `new_root_node` does? Is there some hidden magic happening? Also when we change `new_root_node` to do something special for root nodes, this would happen for all the nodes then.

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -52,87 +57,158 @@ impl Default for Permissions {
...
-		permissions: Permissions,
-	) -> Self {
-		DelegationNode {
-			root_id,
-			parent: Some(parent),
-			owner,
-			permissions,
-			revoked: false,
+	#[derive(Clone, Debug, Encode, Decode, PartialEq)]
+	pub struct DelegationDetails<T: Config> {
```
weichweich, line 192: why do we move the details to a separate struct?

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	// migrations will go through.
+	pub(crate) fn migrate(&self) -> Weight {
+		let mut total_weight_used: Weight = 0;
+		let current_version = LastUpgradeVersion::<T>::get();
+		// Applies all the needed migration from the current version up to
+		// LATEST_UPGRADEABLE_VERSION (which is always the latest possible version - 1).
+		for version in current_version..=LATEST_UPGRADEABLE_VERSION {
+			// It is assumed that the are exactly LATEST_UPGRADEABLE_VERSION+1 migrators, so
+			// as long as the pre-condition is met, the following is assumed to never panic.
+			let version_migrator: &dyn VersionMigrator<T> = self.migrations[version as usize].as_ref();
```
wischli, line 92: Instead of assuming, we should rather check with `get(version as usize)`. Moreover, you should ensure this exists in `pre_migrate` to not be able to panic here at all.
```suggestion
			let version_migrator: &dyn VersionMigrator<T> = self.migrations.get(version as usize).unwrap().as_ref();
```

pallets/delegation/src/lib.rs
```diff
@@ -87,47 +88,87 @@ pub mod pallet {
...
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade() -> Result<(), &'static str> {
+			migrations::StorageMigrator::<T>::new()
+				.post_migration()
+				.map_err(|_| "post-migration failed")
+		}
+	}
 
-	/// Delegation root nodes stored on chain.
+	/// Delegation root nodes stored on chain. DEPRECATED.
```
weichweich, line 111: ah yeah we talked about that, they should go away

pallets/delegation/src/lib.rs
```diff
@@ -87,47 +88,87 @@ pub mod pallet {
...
 	/// It maps from a delegation node ID, including the root node, to the list
 	/// of children nodes, sorted by time of creation.
 	#[pallet::storage]
 	#[pallet::getter(fn children)]
 	pub type Children<T> = StorageMap<_, Blake2_128Concat, DelegationNodeIdOf<T>, Vec<DelegationNodeIdOf<T>>>;
 
+	/// Contains the version of the latest runtime upgrade performed.
+	#[pallet::storage]
+	#[pallet::getter(fn last_upgrade_version)]
+	pub(crate) type LastUpgradeVersion<T> = StorageValue<_, u16, ValueQuery>;
```
weichweich, line 138: didn't checkout the rest of the code but i like [StorageVersion](https://github.com/KILTprotocol/mashnet-node/blob/c82cb51b3c8e20f13aa2f9dbd587890f5e31da97/pallets/parachain-staking/src/lib.rs#L530) more.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+	// Delegations, and Children storage entries to a simpler HierarchyInfos,
+	// DelegationNodes, while maintaining the original information an all
+	// parent-child links.
+	pub struct V0Migrator();
+
+	impl<T: Config> VersionMigrator<T> for V0Migrator {
+		#[cfg(feature = "try-runtime")]
+		fn pre_migrate(&self) -> Result<(), DelegationMigrationError> {
+			ensure!(
+				LastUpgradeVersion::<T>::get() == 0,
```
wischli, line 140:
```suggestion
				LastUpgradeVersion::<T>::get().is_zero(),
```

pallets/delegation/src/lib.rs
```diff
@@ -231,16 +278,17 @@ pub mod pallet {
 		pub fn add_delegation(
 			origin: OriginFor<T>,
 			delegation_id: DelegationNodeIdOf<T>,
-			root_id: DelegationNodeIdOf<T>,
-			parent_id: Option<DelegationNodeIdOf<T>>,
+			root_node_id: DelegationNodeIdOf<T>,
+			parent_id: DelegationNodeIdOf<T>,
```
weichweich, line 282: would it be possible to pass a different `root_node_id` than the one that is stored in the parent? Since we always query the parent,  no need to have the root node as a parameter?

pallets/delegation/src/lib.rs
```diff
@@ -422,23 +459,48 @@ pub mod pallet {
 
 impl<T: Config> Pallet<T> {
 	// Calculate the hash of all values of a delegation transaction
-	fn calculate_hash(
+	fn calculate_delegation_hash_root(
```
weichweich, line 462: why the root in the end?

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+					// Old roots did not have any permissions. So now we give them all permissions.
+					permissions: Permissions::all(),
+					revoked: old_root_node.revoked,
+				};
+				// In here, we already check for potential children of root nodes and ONLY
+				// update the children information. The parent information will be updated
+				// later, when we know we have seen all the children already.
+				let mut new_root_node = DelegationNode::new_root_node(old_root_id, new_root_details);
+				if let Some(root_children_ids) = Children::<T>::take(old_root_id) {
+					new_root_node.children = root_children_ids.iter().copied().collect();
```
wischli, line 173: Placeholder for potential comment

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+				}
+				// Add Chilred::take() weight
```
wischli, line 175: Typo
```suggestion
				// Add Children::take() weight
```

weichweich (commented):
didn't look at `migrations.rs` yet

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+					revoked: old_node.revoked,
+				};
+				// In the old version, a parent None indicated the node is a child of the root.
+				let new_node_parent_id = old_node.parent.unwrap_or(old_node.root_id);
+				let mut new_node =
+					DelegationNode::<T>::new_node(old_node.root_id, new_node_parent_id, new_node_details);
+				if let Some(children_ids) = Children::<T>::take(old_node_id) {
+					new_node.children = children_ids.iter().copied().collect();
+				}
+				// Add Chilred::take() weight
```
wischli, line 199: Same same
```suggestion
				// Add Children::take() weight
```

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+			log::info!("v0 -> v1 delegation storage migrator finished!");
+			total_weight
+		}
+
+		#[cfg(feature = "try-runtime")]
+		fn post_migrate(&self) -> Result<(), DelegationMigrationError> {
+			ensure!(
+				LastUpgradeVersion::<T>::get() == 1,
+				DelegationMigrationError::MigrationResultInconsistent
+			);
```
wischli, line 236: I would definitely iterate over all delegation nodes and check whether all `(parent, child)` pairs match! I know you are doing unit tests (which I like!) but still.

wischli (changes requested):
Mainly some minor typos and scepticism about relying on assumptions in `migrate` as well as insufficient checks in `{pre|post}_migrate`.

I am also unsure about the `u16` StorageVersion to be honest.

pallets/delegation/src/delegation_hierarchy.rs
wischli: Agree. Thanks, I also wanted to point this out. Could be `deprecated.rs` or something like this.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,480 @@
...
+
+	// Checks whether there is at least one storage migration to perform.
+	#[cfg(any(feature = "try-runtime", test))]
+	#[allow(clippy::absurd_extreme_comparisons)]
+	pub(crate) fn pre_migration(&self) -> Result<(), DelegationMigrationError> {
+		ensure!(
+			LastUpgradeVersion::<T>::get() <= migrations::LATEST_UPGRADEABLE_VERSION,
+			DelegationMigrationError::AlreadyLatest
+		);
+
```
wischli, line 75: Somehow my comment got lost...

I would definitely add more checks here which ensure you cannot panic in `migrate`, .e.g.,

1. `self.migrations.get(version as usize)` exists indeed
2. `LATEST_UPGRADEABLE_VERSION` can indeed be incremented by one, e.g. `LATEST_UPGRADEABLE_VERSION.saturating_add(1)` > `LATEST_UPGRADEABLE_VERSION`
3. At least that the `pre_migrate` for the current version does not panic

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: Because I think it's easier to self-contain whatever does not belong to a node as such, such as a parent, an ID, and the children.

weichweich: So you would like to have an abstract tree and then just some content. I wouldn't call it `DelegationTree` then but `Tree<Delegation>` maybe there is already a tree implementation that we could use. 🤔 

But i think it makes it not clearer and just adds one unnecessary step. 🤷🏻 But that is subjective.

pallets/delegation/src/migrations.rs
Diiaablo95: I think it depends whether there's a comma or not. Not entirely sure about that. But I'll have a comment refining round and will keep this in mind!

pallets/delegation/src/migrations.rs
Diiaablo95: As it's still a draft, I'll definitely make sure to have doc comments eventually ☺️

wischli: Yeah I thought it was already final. I remember you said it's in draft early this week but due to the "reviewable" state and the amount of commits I thought this final.

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: I don't think we need to have such a general implementation of a tree, but on the contrary I think that not having everything on the same level is a good idea.

pallets/delegation/src/migrations.rs
Diiaablo95: The thing about the Box is that we would be using several different types that all implement the `VersionMigrator<T>` interface, and their size is known at compile time. I am not aware of any methods that would allow us to get rid of the `Box`, unless we replace the trait with an enum.

pallets/delegation/src/migrations.rs
Diiaablo95: I guess there's no difference between [index] and `.get().unwrap()`? We still need to assume it exists via some tests before running the migration. Once we are in the migration, it's already too late, right?

pallets/delegation/src/migrations.rs
wischli: Okay. If we don't replace it with an enum, regard this comment as resolved.

pallets/delegation/src/migrations.rs
wischli: That is correct, yes. That's why I would like to check the existence of this in the `pre_migrate` for all versions (see comment [below](https://github.com/KILTprotocol/mashnet-node/pull/226#discussion_r670265414)).

pallets/delegation/src/lib.rs
Diiaablo95: How did we decide to change the deprecated types? 😅

wischli: I think we agreed they should be moved to a deprecated file. I am unsure whether you might into problems during the migration then though.

Diiaablo95: I think the problem is that they are still marked with `pallet::storage`, so not sure how to better structure such types.

wischli: You should remove it from the pallet to `deprecated` and initialize it there via the `decl_storage` macro:

* PR in which a storage type was changed and moved to deprecated: https://github.com/hicommonwealth/substrate/pull/14/files#diff-00c91017af35e93cc641a7bc0d73202f7bd349f13d5bc7f0736d7a272919195dR5
* PR in which a Storage entry was moved to an associated type in the config: https://github.com/paritytech/substrate/pull/8773/files#diff-70e9723e9db62816e35f6f885b6770a8449c75a6c2733e9fa7a245fe52c4656cL621

It seems like we also need to remove the Storage Prefix: https://github.com/paritytech/substrate/pull/8773/files#diff-acaa239811fe81b5d496c1b3d5059f43de8f08cd530c4bbc44fac857612936d4R30

pallets/delegation/src/lib.rs
wischli: Note: Soon, there will be support for a `pallet-version` key in the Cargo.toml which represents the `PalletVersion`. Guess what type it will be? Yes, `u16` 😅

https://github.com/paritytech/substrate/pull/9165

Diiaablo95: Ahah too late that I already converted everything to enums 😂

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/0040f2d75ca135d86b18c86cea6f54e0cb7e24e5.

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/0040f2d75ca135d86b18c86cea6f54e0cb7e24e5.

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/cbc29eb171e169aa6cc223b017b57d2d7b4303bd.

pallets/delegation/src/lib.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/0040f2d75ca135d86b18c86cea6f54e0cb7e24e5.

pallets/delegation/src/lib.rs
Diiaablo95: Please take a new look at the whole migrations file.

pallets/delegation/src/lib.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/cbc29eb171e169aa6cc223b017b57d2d7b4303bd.

pallets/delegation/src/lib.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/cbc29eb171e169aa6cc223b017b57d2d7b4303bd.

pallets/delegation/src/lib.rs
```diff
@@ -185,25 +226,31 @@ pub mod pallet {
...
+				!<DelegationHierarchies<T>>::contains_key(&root_node_id),
+				Error::<T>::HierarchyAlreadyExists
+			);
 
 			ensure!(
 				<ctype::Ctypes<T>>::contains_key(&ctype_hash),
 				<ctype::Error<T>>::CTypeNotFound
 			);
 
-			log::debug!("insert Delegation Root");
```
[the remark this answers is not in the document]
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/cbc29eb171e169aa6cc223b017b57d2d7b4303bd.

pallets/delegation/src/migrations.rs
Diiaablo95: Changed almost everything in the new migrations module.

pallets/delegation/src/migrations.rs
Diiaablo95: Changed almost everything in the new migrations module. Please review the new documentation.

pallets/delegation/src/migrations.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/commit/25a01bd68312b445ced61688cebebada5861613f.

Diiaablo95:
@wischli @weichweich please you can now take a second review at the PR, which is to be considered the final draft. The whole migration logic has changed, and I managed to move all the deprecated types in their own `deprecated.rs` module. I will open a types PR in the meanwhile and also run the benchmarks on the benchmarking server.

Diiaablo95:
We should also probably bump the runtime version if we want to test the upgrade where https://github.com/KILTprotocol/mashnet-node/pull/222 has already been deployed. And to my knowledge the runtime version must also be bumped in the wasm embedded in the chain spec, right?

pallets/delegation/src/deprecated.rs
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	pub(crate) mod storage {
+		use frame_support::{decl_module, decl_storage};
+		use sp_std::prelude::*;
+
+		use super::*;
+
+		decl_module! {
+			pub struct OldPallet<T: Config> for enum Call where origin: T::Origin {}
```
wischli, line 92: Maybe we want to give this a more descriptive name like `PalletV1`?

wischli (commented):
> We should also probably bump the runtime version if we want to test the upgrade where #222 has already been deployed. And to my knowledge the runtime version must also be bumped in the wasm embedded in the chain spec, right?

Yes, we should bump the `spec_version` for all three runtimes. The chain specs do not need to be updated until we need to restart one of our networks from scratch.

wischli (commented):
Very minor things on the nitpick level. Otherwise, LGTM! Really like the updated approach. Looks much cleaner.

pallets/delegation/src/deprecated.rs
Diiaablo95: It's already part of the the `v1` module. It is the same reasoning as in https://github.com/KILTprotocol/mashnet-node/pull/226#discussion_r670231697.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,736 @@
...
+	/// It migrates the old storage entries to the new ones.
+	///
+	/// Specifically, for each entry in Roots, a new entry in
+	/// DelegationHierarchies + a new node in DelegationNodes is created.
+	/// Furthermore, nodes in Delegations are migrated to the new structure and
+	/// stored under DelegationNodes, with any children from the Children
+	/// storage entry added to the nodes under the children set.
+	pub(crate) fn migrate<T: Config>() -> Weight {
+		log::info!("v1 -> v2 delegation storage migrator started!");
+		let mut total_weight = 0u64;
```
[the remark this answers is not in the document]
Diiaablo95: Fixed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/78ea8b51f15e76eafedd411eed18cebeb4dcce18.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,736 @@
...
+		total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
+		log::debug!("Total weight consumed: {}", total_weight);
+		log::info!("v1 -> v2 delegation storage migrator finished!");
+
+		total_weight
+	}
+
+	fn migrate_roots<T: Config>(new_nodes: &mut BTreeMap<DelegationNodeIdOf<T>, DelegationNode<T>>) -> Weight {
+		let total_weight =
+			deprecated::v1::storage::Roots::<T>::iter().fold(0u64, |mut total_weight, (old_root_id, old_root_node)| {
```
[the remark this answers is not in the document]
Diiaablo95: Fixed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/78ea8b51f15e76eafedd411eed18cebeb4dcce18.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,736 @@
...
+				DelegationHierarchies::insert(old_root_id, new_hierarchy_details);
+				// Adds a read from Roots::drain() and a write from
+				// DelegationHierarchies::insert() weights
+				total_weight = total_weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+				// Add the node to the temporary map of nodes to be added at the end.
+				new_nodes.insert(old_root_id, new_root_node);
+
+				total_weight
+			});
+		// If runtime testing, makes sure that the old number of roots is reflected in
```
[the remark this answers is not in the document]
Diiaablo95: Fixed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/78ea8b51f15e76eafedd411eed18cebeb4dcce18.

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,736 @@
...
+				// Add Children::take() weight
+				total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+				// Adds a read from Roots::drain() weight
+				total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+				new_nodes.insert(old_node_id, new_node);
+
+				total_weight
+			},
+		);
+		// If runtime testing, makes sure that the old number of delegations is
```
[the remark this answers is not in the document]
Diiaablo95: Fixed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/78ea8b51f15e76eafedd411eed18cebeb4dcce18.

wischli (approved):
LGTM! I suppose Albi should still take a look once he's back.

pallets/delegation/src/delegation_hierarchy.rs
weichweich: I think I couldn't convey my point so let me rephrase this.

In my opinion it's always a good idea to structure data when there is a category to structure them by. Eg. don't have `x` and `y` fields, but have a nested `position` field that has `y` and `x`. It makes sense to have position because it is a meaningful group for `x` and `y`.

Looking at the `DelegationDetails` it's not clear what this category is. What goes into details and what doesn't? `details` is not a real category. It's just a way of saying the number of fields in this struct was to high and i move it some where else.

But I think you actually split the fields by things that are relevant for the tree structure and things that are not relevant for the tree structure. So for me a meaningful category would be `tree_content` which makes only sense if you have a generic tree data structure.

For me it looks like the structure does more harm than it helps at the moment (especially `DelegationHierarchyDetails` which only has one field).

That said, this is my opinion and it's a style thing. So if i couldn't convince you go ahead and merge.

tl;dr; I fucking hate the word details! could be anything! What even are details?!

pallets/delegation/src/lib.rs
```diff
@@ -451,16 +496,13 @@ impl<T: Config> Pallet<T> {
...
 	) -> Result<(bool, u32), DispatchError> {
-		let delegation_node = <Delegations<T>>::get(delegation).ok_or(Error::<T>::DelegationNotFound)?;
+		let delegation_node = <DelegationNodes<T>>::get(delegation).ok_or(Error::<T>::DelegationNotFound)?;
 
 		// Check if the given account is the owner of the delegation and that the
 		// delegation has not been removed
-		if &delegation_node.owner == identity {
-			Ok((!delegation_node.revoked, 0u32))
+		if &delegation_node.details.owner == identity {
+			Ok((!delegation_node.details.revoked, 0u32))
```
weichweich, line 504: 😰

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,741 @@
...
+pub trait VersionMigratorTrait<T> {
+	#[cfg(feature = "try-runtime")]
+	fn pre_migrate(&self) -> Result<(), &str>;
+	fn migrate(&self) -> Weight;
+	#[cfg(feature = "try-runtime")]
+	fn post_migrate(&self) -> Result<(), &str>;
+}
+
+/// Storage version of the delegation pallet.
+#[allow(non_camel_case_types)]
```
weichweich, line 38: The capitalized `V1` should work? why not use that?

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,741 @@
...
+#[derive(Copy, Clone, Encode, Eq, Decode, Ord, PartialEq, PartialOrd)]
+pub enum DelegationStorageVersion {
+	v1,
+	v2,
+}
+
+#[allow(dead_code)]
```
weichweich, line 45: you could also add the `#[cfg(feature = "try-runtime")]` feature flag?

pallets/delegation/src/migrations.rs
```diff
@@ -0,0 +1,741 @@
...
+	#[cfg(feature = "try-runtime")]
+	fn post_migrate(&self) -> Result<(), &str> {
+		match *self {
+			Self::v1 => v1::post_migrate::<T>(),
+			Self::v2 => Err("Migration from v2 should have never happened in the first place."),
+		}
+	}
+}
+
+mod v1 {
```
weichweich, line 95: i would move this to it's own file. Especially with the tests in the end it becomes a bit overwhelming.

weichweich (changes requested):
LGTM

pallets/delegation/src/lib.rs
```diff
@@ -421,24 +434,56 @@ pub mod pallet {
 }
 
 impl<T: Config> Pallet<T> {
-	// Calculate the hash of all values of a delegation transaction
-	fn calculate_hash(
+	// Calculates the hash of all values of a delegation creation transaction.
+	fn calculate_delegation_creation_hash(
```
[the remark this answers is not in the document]
Diiaablo95: For the first point, the function is used only to re-create the information that the delegatee signs upon delegation creation. I agree it does refer to the delegation content, but it is used only for creations. And it is not even the hash of the whole delegation, but only what is needed to be protected when creating a new delegation.

For the second point, we have removed the root_id from the extrinsic parameter, so the submitter is not required to know what `hierarchy_id` the node will end up into, right? Otherwise are you suggesting we should still include the `root_id` in the hash but not include it as a parameter in the extrinsic call?

pallets/delegation/src/delegation_hierarchy.rs
Diiaablo95: I see your point, and I probably agree that details is not the best word. Plus, you got the point that my intention here is twofold:

- group everything that is tree-related and everything else that is not tree related. How could that be achieved in a different way that perhaps satisfies both preferences?
- for the hierarchy, it is true that for now it is just the CTYPE, but could be in future something else as well. And we would have to make it a struct anyway eventually, right? In order to avoid having tuples around. What would be your suggestion here?

pallets/delegation/src/lib.rs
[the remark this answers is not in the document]
weichweich: > For the first point, the function is used only to re-create the information that the delegatee signs upon delegation creation. I agree it does refer to the delegation content, but it is used only for creations. And it is not even the hash of the whole delegation, but only what is needed to be protected when creating a new delegation.

The signature is used to ensure the delegation is created with consent of the receiver. i would feel better if the rootId is included.
 
> For the second point, we have removed the root_id from the extrinsic parameter, so the submitter is not required to know what `hierarchy_id` the node will end up into, right? Otherwise are you suggesting we should still include the `root_id` in the hash but not include it as a parameter in the extrinsic call?

The submitter is required to know to which delegation hierarchy the new node will be added to. But the submitter is someone else. The signature is done by the receiver of the delegation. And the receiver needs to be conscious about which delegation hierarchy he will be added to.

pallets/delegation/src/delegation_hierarchy.rs
weichweich: imHo i wouldn't have a nested structure here. 😅 If you prefer it nested, I don't see a better solution.

pallets/delegation/src/lib.rs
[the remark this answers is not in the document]
Diiaablo95: Re-added `root_id` in the computation of the delegation creation hash: https://github.com/KILTprotocol/mashnet-node/pull/226/commits/310cad8bc19ba5887a722660543ab4e0b8d6ce17.

pallets/delegation/src/migrations.rs
Diiaablo95: Addressed in https://github.com/KILTprotocol/mashnet-node/pull/226/commits/ef941d3877348943b383b329bfa62d393c9562af.

pallets/delegation/src/migrations.rs
Diiaablo95: Thanks for this -> https://github.com/KILTprotocol/mashnet-node/pull/226/commits/ef941d3877348943b383b329bfa62d393c9562af.

pallets/delegation/src/migrations.rs
Diiaablo95: Does this look good enough? https://github.com/KILTprotocol/mashnet-node/pull/226/commits/e96bc31fb6d8f4809f7a29616a363ce4b45af05f.

weichweich: Looks awesome! But maybe we keep the `migrations.rs` file instead of `migrations/mod.rs`? There is no clear rule for rust projects and both are equivalent (you can still have `migrations/v1.rs`). I personally favor `migrations.rs` since you won't have a thousand `mod.rs` files open where you don't know which one is which. 😅 
I researched this once and couldn't find a clear rule for that. So i guess do what you like more.

weichweich (approved):
everything has been said!

PR #229, opened 2021-07-21
feat: add bootnodes&telemetry to chain spec

weichweich:
## feat: add bootnodes&telemetry to chain spec 

* ci: always run pipelines. 😢 The changes where not detected correctly 

## Checklist:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,11 +6,6 @@ clippy:
   image: paritytech/ci-linux:105b919b-20210617
   stage: test
   timeout: 1 hours
-  only:
-    refs:
-      - branches
-    changes:
-      - ./**/*.{rs,toml,lock}
   script:
     - rustup component add clippy --toolchain nightly
     - cargo +nightly clippy -- -D warnings
@@ -20,18 +15,14 @@ fmt:
   stage: test
   timeout: 1 hours
   script:
-    - rustup component add rustfmt
-    - cargo fmt -- --check
+    - rustup component add rustfmt --toolchain nightly
+    - ls .
+    - cargo +nightly fmt -- --check
 
 test:
   image: paritytech/ci-linux:105b919b-20210617
   stage: test
   timeout: 1 hours
-  only:
-    refs:
-      - branches
-    changes:
-      - ./**/*.{rs,toml,lock}
   script:
     - cargo test --all --all-features
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,7 +16,6 @@ fmt:
   timeout: 1 hours
   script:
     - rustup component add rustfmt --toolchain nightly
-    - ls .
     - cargo +nightly fmt -- --check
 
 test:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,7 +8,7 @@ clippy:
   timeout: 1 hours
   script:
     - rustup component add clippy --toolchain nightly
-    - cargo +nightly clippy -- -D warnings
+    - cargo +nightly clippy --all-features --all-targets -- -D warnings
 
 fmt:
   image: paritytech/ci-linux:105b919b-20210617
@@ -23,7 +23,7 @@ test:
   stage: test
   timeout: 1 hours
   script:
-    - cargo test --all --all-features
+    - cargo test --all-features --all-targets
 
 build:
   image:
--- a/nodes/parachain/src/chain_spec/spiritnet.rs
+++ b/nodes/parachain/src/chain_spec/spiritnet.rs
@@ -168,10 +168,10 @@ pub fn get_chain_spec_wilt() -> Result<ChainSpec, String> {
 		vec![
 			"/dns4/bootnode.kilt.io/tcp/30360/p2p/12D3KooWRPR7q1Rgwurd4QGyUUbVnN4nXYNVzbLeuhFsd9eXmHJk"
 				.parse()
-				.expect("bootnode address is formatted correctly."),
+				.expect("bootnode address is formatted correctly; qed"),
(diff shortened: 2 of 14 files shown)


.gitlab-ci.yml
```diff
@@ -20,18 +15,13 @@ fmt:
...
   image: paritytech/ci-linux:105b919b-20210617
   stage: test
   timeout: 1 hours
-  only:
-    refs:
-      - branches
-    changes:
-      - ./**/*.{rs,toml,lock}
   script:
     - cargo test --all --all-features
```
Diiaablo95, line 26: Same as above: perhaps `--all-targets`?

.gitlab-ci.yml
```diff
@@ -6,11 +6,6 @@ clippy:
...
   stage: test
   timeout: 1 hours
-  only:
-    refs:
-      - branches
-    changes:
-      - ./**/*.{rs,toml,lock}
   script:
     - rustup component add clippy --toolchain nightly
     - cargo +nightly clippy -- -D warnings
```
Diiaablo95, line 11: Would it make sense to add `--all-features --all-targets` as well? At least for clippy, `--all-targets` detects a whole new class of warnings that are otherwise not caught.

nodes/parachain/src/chain_spec/spiritnet.rs
```diff
@@ -164,8 +165,15 @@ pub fn get_chain_spec_wilt() -> Result<ChainSpec, String> {
 				id,
 			)
 		},
-		vec![],
-		None,
+		vec![
+			"/dns4/bootnode.kilt.io/tcp/30360/p2p/12D3KooWRPR7q1Rgwurd4QGyUUbVnN4nXYNVzbLeuhFsd9eXmHJk"
+				.parse()
+				.expect("bootnode address is formatted correctly."),
```
Diiaablo95, line 171: Should it be bootnode address is NOT formatted correctly? As the message is shown when `expect()` fails, right?

nodes/parachain/src/chain_spec/spiritnet.rs
```diff
@@ -164,8 +165,15 @@ pub fn get_chain_spec_wilt() -> Result<ChainSpec, String> {
...
+			"/dns4/bootnode.kilt.io/tcp/30361/p2p/12D3KooWDAEqpTRsL76itsabbh4SeaqtCM6v9npQ8eCeqPbbuFE9"
+				.parse()
+				.expect("bootnode address is formatted correctly."),
```
Diiaablo95, line 174: Same as above

nodes/parachain/src/chain_spec/spiritnet.rs
```diff
@@ -164,8 +165,15 @@ pub fn get_chain_spec_wilt() -> Result<ChainSpec, String> {
...
+		],
+		Some(TelemetryEndpoints::new(vec![(TELEMETRY_URL.to_string(), 0)]).expect("KILT telemetry url is valid; qed")),
```
Diiaablo95, line 176: Same here?

nodes/parachain/src/chain_spec/spiritnet.rs
```diff
@@ -216,7 +224,7 @@ pub fn get_chain_spec_spiritnet() -> Result<ChainSpec, String> {
 			)
 		},
 		vec![],
-		None,
+		Some(TelemetryEndpoints::new(vec![(TELEMETRY_URL.to_string(), 0)]).expect("KILT telemetry url is valid; qed")),
```
Diiaablo95, line 227: ... and here

Diiaablo95 (changes requested):
A couple of questions about the flags we use when running `clippy` and `fmt`. Plus, some minor comments about error messages.

.gitlab-ci.yml
weichweich: That's completely unrelated to my PR! 😭 But sure i will add that...

nodes/parachain/src/chain_spec/spiritnet.rs
weichweich: `KILT telemetry url is valid; qed` so the intention is to put the proof in the expect message.

> this will never fail because the URL is valid

nodes/parachain/src/chain_spec/spiritnet.rs
weichweich: a87edf7

nodes/parachain/src/chain_spec/spiritnet.rs
weichweich: a87edf7

nodes/parachain/src/chain_spec/spiritnet.rs
weichweich: a87edf7

nodes/parachain/src/chain_spec/spiritnet.rs
weichweich: a87edf7

.gitlab-ci.yml
weichweich: bfb0a36

.gitlab-ci.yml
weichweich: bfb0a36

Diiaablo95 (approved):
Ok, I must say that I am not entirely sold on the "logic" of the `expect` message. But anyway, it is a minor detail and it should definitely not block this PR from being merged.

Hence 🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁
LGTM 🤌🤌🤌🤌🤌

weichweich:
> Ok, I must say that I am not entirely sold on the "logic" of the `expect` message. But anyway, it is a minor detail and it should definitely not block this PR from being merged.
> 
> Hence 🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁
> LGTM 🤌🤌🤌🤌🤌

Yeah I also think that you are right. Let's rework all the expect messages in a separate PR 😉

PR #230, opened 2021-07-30
fix: authorship pallet order

wischli:
## Attempts to fix KILTProtocol/ticket#1483
Attempts to fix our storage root mismatch issue.
* Adds missing registration of unchecked weight to `note_author`
* Adds logs to `note_author` each time an account is rewarded
* Changes order of pallets in `peregrine` and `spiritnet` runtimes based on the recommendation made by Shawn. We should leave more room for future pallets. Since we are not using the scheduler or should not have delayed transactions, we should be fine to change the indices without a migration.
  * Important note: Moved `Authorship` to first consensus pallet based on [order in Statemine runtime](https://github.com/paritytech/cumulus/blob/master/polkadot-parachains/statemine/src/lib.rs#L681) as well as [order in Kusama/Polkadot runtimes](https://github.com/paritytech/polkadot/blob/master/runtime/kusama/src/lib.rs#L1430)
    * Also talked to Kian about it: Since we use the `Authorship Eventhandler` in the staking pallet, Authorship needs to be initiated before staking 

Currently planned to be deployed on Staging network!

## Checklist:


weichweich (approved):
LGTM

Diiaablo95 (approved):
👍

PR #231, opened 2021-08-04
feat: use DidOrigin for DID update and deletion

Diiaablo95:
## fixes KILTProtocol/ticket#1490

## Checklist:

  - [x] If not, I have opened a companion PR with the type changes in the [KILT types-definitions repository](https://github.com/KILTprotocol/type-definitions/pull/12)


pallets/did/src/lib.rs
```diff
@@ -72,12 +72,11 @@
...
-//! - `submit_did_create_operation` - Register a new DID on the KILT blockchain
-//!   under the given DID identifier.
-//! - `submit_did_update_operation` - Update any keys or the endpoint URL of an
-//!   existing DID.
-//! - `submit_did_delete_operation` - Delete the specified DID and all related
-//!   keys from the KILT blockchain.
+//! - `create` - Register a new DID on the KILT blockchain under the given DID
+//!   identifier.
+//! - `update` - Update any keys or the endpoint URL of an existing DID.
+//! - `dekete` - Delete the specified DID and all related keys from the KILT
```
wischli, line 78:
```suggestion
//! - `delete` - Delete the specified DID and all related keys from the KILT
```

weichweich, line 78:
```suggestion
//! - `delete` - Delete the specified DID and all related keys from the KILT
```

pallets/did/src/lib.rs
```diff
@@ -348,53 +349,43 @@ pub mod pallet {
...
+				details.new_key_agreement_keys.len().saturated_into::<u32>(),
+				details.new_endpoint_url.as_ref().map_or(0u32, |url| url.len().saturated_into::<u32>())
+			))
+			.max(<T as pallet::Config>::WeightInfo::create_ecdsa_keys(
+				details.new_key_agreement_keys.len().saturated_into::<u32>(),
+				details.new_endpoint_url.as_ref().map_or(0u32, |url| url.len().saturated_into::<u32>())
+			))
+		)]
+		pub fn create(origin: OriginFor<T>, details: DidCreationDetails<T>, signature: DidSignature) -> DispatchResult {
 			let sender = ensure_signed(origin)?;
```
weichweich, line 366: we can't use the did origin here because the did doesn't exists on chain yet?

runtimes/peregrine/src/lib.rs
```diff
@@ -852,6 +856,9 @@ impl did::DeriveDidCallAuthorizationVerificationKeyRelationship for Call {
 			Call::Attestation(_) => Some(did::DidVerificationKeyRelationship::AssertionMethod),
 			Call::Ctype(_) => Some(did::DidVerificationKeyRelationship::AssertionMethod),
 			Call::Delegation(_) => Some(did::DidVerificationKeyRelationship::CapabilityDelegation),
+			// DID creation is not allowed through the DID proxy.
```
wischli, line 859: Just wondering what can happen if we allow this?

wischli (approved):
LGTM! Just found a minor typo and I have a question.

runtimes/peregrine/src/lib.rs
Diiaablo95: The verification prior to dispatch would always fail with a `DidNotPresent` error as there's no DID in the storage. I think it would be most clear to fail with a `BadOrigin` error, as is the case now, rather than with a `DidNotPresent`. What do you think?

pallets/did/src/did_details.rs
```diff
@@ -542,18 +542,14 @@ impl<T: Config> TryFrom<(DidCreationOperation<T>, DidVerificationKey)> for DidDe
...
+// storage and by applying the changes in the [DidUpdateDetails].
 //
 // The operation fails with a [DidError] if the update operation instructs to
 // delete a verification key that is not associated with the DID.
-//
-// Please note that this method does not perform any checks regarding
-// the validity of the [DidUpdateOperation] signature nor whether the nonce
-// provided is valid.
-impl<T: Config> TryFrom<(DidDetails<T>, DidUpdateOperation<T>)> for DidDetails<T> {
+impl<T: Config> TryFrom<(DidDetails<T>, DidUpdateDetails<T>)> for DidDetails<T> {
```
weichweich, line 549: Can we move that into a function in `impl DidDetails` that is named something like `update`? This looks like an abuse of the `TryFrom` trait

pallets/did/src/lib.rs
```diff
@@ -348,53 +349,43 @@ pub mod pallet {
...
-			let account_did_auth_key = operation
-				.did
-				.verify_and_recover_signature(&operation.encode(), &signature)
+			let account_did_auth_key = did_identifier
+				.verify_and_recover_signature(&details.encode(), &signature)
 				.map_err(<Error<T>>::from)?;
 
-			let did_entry =
-				DidDetails::try_from((operation.clone(), account_did_auth_key)).map_err(<Error<T>>::from)?;
+			let did_entry = DidDetails::try_from((details, account_did_auth_key)).map_err(<Error<T>>::from)?;
```
weichweich, line 378: This would look better wouldn't it?

```suggestion
			let did_entry = details.update(account_did_auth_key)?;
```

Also do you really need to map the error? This should happen automatically with the `?`?

pallets/did/src/did_details.rs
weichweich: I would also accept it if you say that this is out of scope for this PR...

pallets/did/src/lib.rs
```diff
@@ -172,6 +172,7 @@ pub mod pallet {
 		type Origin: From<DidRawOrigin<DidIdentifierOf<Self>>>;
 		#[cfg(feature = "runtime-benchmarks")]
 		type Origin: From<RawOrigin<DidIdentifierOf<Self>>>;
+		type EnsureOrigin: EnsureOrigin<Success = DidIdentifierOf<Self>, <Self as frame_system::Config>::Origin>;
```
weichweich, line 175: Do we need a generic origin? Can we not simply use the did origin that we defined in this crate?

pallets/did/src/tests.rs
```diff
@@ -23,23 +23,23 @@ use sp_std::{collections::btree_set::BTreeSet, convert::TryFrom};
 use crate::{self as did, mock::*};
 use ctype::mock as ctype_mock;
 
-// submit_did_create_operation
+// create
```
weichweich, line 26: i guess this comment doesn't add much value.

pallets/did/src/tests.rs
```diff
@@ -453,12 +467,11 @@ fn check_successful_keys_deletion_update() {
...
-	let mut operation = generate_base_did_update_operation(alice_did.clone());
-	operation.attestation_key_update = did::DidVerificationKeyUpdateAction::Delete;
-	operation.delegation_key_update = did::DidVerificationKeyUpdateAction::Delete;
-	operation.tx_counter = old_did_details.last_tx_counter + 1u64;
+	let mut details = generate_base_did_update_details();
+	details.attestation_key_update = did::DidVerificationKeyUpdateAction::Delete;
+	details.delegation_key_update = did::DidVerificationKeyUpdateAction::Delete;
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 474: Do we still need to sign here? Can we not just remove this line?

pallets/did/src/tests.rs
```diff
@@ -573,10 +574,10 @@ fn check_successful_keys_multiuse_update() {
 	old_did_details.update_attestation_key(did::DidVerificationKey::from(old_att_key.public()), 0u64);
 
 	// Remove attestation key
-	let mut operation = generate_base_did_update_operation(alice_did.clone());
-	operation.attestation_key_update = did::DidVerificationKeyUpdateAction::Delete;
+	let mut details = generate_base_did_update_details();
+	details.attestation_key_update = did::DidVerificationKeyUpdateAction::Delete;
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 580: same here

pallets/did/src/tests.rs
```diff
@@ -828,25 +665,21 @@ fn check_max_limit_public_keys_to_remove_did_update() {
 	let keys_ids_to_remove =
 		get_public_keys_to_remove(<Test as did::Config>::MaxNewKeyAgreementKeys::get().saturating_add(1));
 
-	let mut operation = generate_base_did_update_operation(alice_did.clone());
-	operation.public_keys_to_remove = keys_ids_to_remove;
+	let mut details = generate_base_did_update_details();
+	details.public_keys_to_remove = keys_ids_to_remove;
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 671: here

pallets/did/src/tests.rs
```diff
@@ -862,25 +695,21 @@ fn check_url_too_long_did_update() {
 	// Max URL length allowed + 1
 	let new_endpoint_url = get_url_endpoint(<Test as did::Config>::MaxUrlLength::get().saturating_add(1));
 
-	let mut operation = generate_base_did_update_operation(alice_did.clone());
-	operation.new_endpoint_url = Some(new_endpoint_url);
+	let mut details = generate_base_did_update_details();
+	details.new_endpoint_url = Some(new_endpoint_url);
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 701: here

pallets/did/src/tests.rs
```diff
@@ -894,31 +723,27 @@ fn check_currently_active_authentication_key_update() {
...
-	operation.public_keys_to_remove = vec![generate_key_id(
+	details.public_keys_to_remove = vec![generate_key_id(
 		&did::DidVerificationKey::from(auth_key.public()).into(),
 	)]
 	.iter()
 	.copied()
 	.collect::<BTreeSet<TestKeyId>>();
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 735: here

pallets/did/src/tests.rs
```diff
@@ -934,29 +759,25 @@ fn check_currently_active_delegation_key_update() {
...
+	let mut details = generate_base_did_update_details();
 	// Trying to remove the currently active delegation key
-	operation.public_keys_to_remove = vec![generate_key_id(&did::DidVerificationKey::from(del_key.public()).into())]
+	details.public_keys_to_remove = vec![generate_key_id(&did::DidVerificationKey::from(del_key.public()).into())]
 		.iter()
 		.copied()
 		.collect::<BTreeSet<TestKeyId>>();
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 769: here

pallets/did/src/tests.rs
```diff
@@ -972,29 +793,25 @@ fn check_currently_active_attestation_key_update() {
...
+	let mut details = generate_base_did_update_details();
 	// Trying to remove the currently active attestation key
-	operation.public_keys_to_remove = vec![generate_key_id(&did::DidVerificationKey::from(att_key.public()).into())]
+	details.public_keys_to_remove = vec![generate_key_id(&did::DidVerificationKey::from(att_key.public()).into())]
 		.iter()
 		.copied()
 		.collect::<BTreeSet<TestKeyId>>();
 
-	let signature = auth_key.sign(operation.encode().as_ref());
+	let _signature = auth_key.sign(details.encode().as_ref());
```
weichweich, line 803: here

weichweich (changes requested):
few minor things that i would like to talk about.

Do we need the generic origin? Can we not require a did origin?

pallets/did/src/lib.rs
Diiaablo95: `details` is what is given in the extrinsic call, so we can't just call `update` on that. We need to create a new instance of `DidDetails` from scratch, and we do that by analysing the information inside the creation details + the authentication key we recover from the signature.

For the second point, it is true that `DidError` automatically maps to `DispatchError`, but `try_from` generates an `InputError`, which we must map to a `DidError` first, which then gets converted to a `DispatchError`.

pallets/did/src/tests.rs
Diiaablo95: It separates the different extrinsics that are tested here. Do you think it's not useful?

pallets/did/src/tests.rs
Diiaablo95: Fault of `clippy --fix` 😅  Good catch, thanks!

pallets/did/src/lib.rs
Diiaablo95: Perhaps we could call `update` from an update details?

weichweich: > `details` is what is given in the extrinsic call, so we can't just call `update` on that. We need to create a new instance of `DidDetails` from scratch, and we do that by analysing the information inside the creation details + the authentication key we recover from the signature.

Yes I agree that `update` doesn't fit 100%. But `update` can return a new instance. I think the convention is to call the fn `updated` than.
`TryFrom` takes one thing and transforms it into another, if possible. I don't feel like it's matching this use case here.  IMO So while update only fits 80%, try from fits like 50%.

> For the second point, it is true that `DidError` automatically maps to `DispatchError`, but `try_from` generates an `InputError`, which we must map to a `DidError` first, which then gets converted to a `DispatchError`.

😂 Ok I see.

pallets/did/src/tests.rs
weichweich: It creates sections? In objective-c you can [make sections in code](https://stackoverflow.com/questions/6853821/what-is-the-significance-of-pragma-marks-why-do-we-need-pragma-marks) which is really cool :D Didn't recognized it as a section. Thought it was only for the next test

pallets/did/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/231/commits/25ba666268942490f25562b95e8778f67583b267

pallets/did/src/tests.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/231/commits/677a827c2531efc1e80df819b201df8fca8d0334

pallets/did/src/lib.rs
Diiaablo95: Maybe something like this? https://github.com/KILTprotocol/mashnet-node/pull/231/commits/d8bddbf9ca78a7d0f3084ee0eac1f781755e782e

pallets/did/src/did_details.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/231/commits/d8bddbf9ca78a7d0f3084ee0eac1f781755e782e

pallets/did/src/lib.rs
weichweich: no need to pass a tuple. 😉 Otherwise it's perfect! 😍 

```
pub fn from_creation_details((details, new_auth_key): (DidCreationDetails<T>, DidVerificationKey)) -> Result<Self, InputError>
```

Diiaablo95: Voilà https://github.com/KILTprotocol/mashnet-node/pull/231/commits/391260f57754d6072b91848e420b4487e229e5b2 🎩

PR #232, opened 2021-08-04
fix: allow any uid when resetting the chain spec

weichweich:
## allow any uid when resetting the chain spec

The reset script could only be executed by the user with uid 1000.

## How to test:

Run the following command with a uid different to the default uid inside the docker img (uid 1000). Or run it as root.

```
python3 .maintain/reset-spec/app.py -i kiltprotocol/peregrine:develop --peregrine-stg -v
```

### custom types

No type changes

## Checklist:


wischli (approved):
Thanks a lot! LGTM!

Diiaablo95 (approved):
🤌

PR #235, opened 2021-08-05
feat: add support for hash, content type, and multiple URLs as service endpoints

Diiaablo95:
## fixes KILTProtocol/ticket#1505

This PR modifies the service endpoints of a DID by adding the possibility to specify a hash of the pointed service endpoints document, a content type to allow other entities to correctly parse the document, and up to three URLs to fetch the document from. The logic about what URL to use in case there are more than ones for a given resource are out of the scope of the chain code.

To do so, the PR updates the `DidVerificationKeyUpdateAction` and makes it more generic into a `DidFragmentUpdateAction<FragmentType>` that is then used for both verification key updates with `DidFragmentUpdateAction<DidVerificationKey` and for service endpoint updates with `DidFragmentUpdateAction<ServiceEndpoints>`, with `ServiceEndpoints` being:

```rust
pub struct ServiceEndpoints {
	pub content_hash: Hash,
	pub urls: Vec<Url>,
	pub content_type: ContentType,
}
```

I have not migrated this struct to use `BoundedVec`, but I hope that is fine. We have one more parameter in the trait's Config called `MaxEndpointUrlsCount` which specifies how many URLs can be included in a `DidFragmentUpdateAction` for the service endpoints. We decided to go with three which seems like a reasonable number to provide good redundancy to the data.

## How to test:

`cargo test --all --all-features --all-targets` or simply `cargo test -p did`

### Custom types

<details>
  <summary>JS-Types</summary>
  
  ```json
{
    "AccountInfo": "AccountInfoWithTripleRefCount",
    "Address": "MultiAddress",
    "AmountOf": "i128",
    "Balance": "u128",
    "BlockNumber": "u64",
    "Index": "u64",
    "LookupSource": "MultiAddress",
    "CtypeCreatorOf": "AccountId",
    "CtypeHashOf": "Hash",
    "ClaimHashOf": "Hash",
    "AttesterOf": "AccountId",
    "AttestationDetails": {
        "ctypeHash": "CtypeHashOf",
        "attester": "AttesterOf",
        "delegationId": "Option<DelegationNodeIdOf>",
        "revoked": "bool"
    },
    "Permissions": "u32",
    "DelegationNodeIdOf": "Hash",
    "DelegatorIdOf": "AccountId",
    "DelegateSignatureTypeOf": "Vec<u8>",
    "DelegationRoot": {
        "ctypeHash": "CtypeHashOf",
        "owner": "DelegatorIdOf",
        "revoked": "bool"
    },
    "DelegationNode": {
[...]

PR #236, opened 2021-08-06
ci: add separate cargo test

wischli:
## No ticket

In #231 none of the CI tests failed but `cargo test` was broken for the did pallet. This should be fixed by #235. However, in #235 we stumbled into the same error. Thus, I think we should add a separate `cargo test step to the CI as proposed here.

## Checklist:


wischli:
CI failure proofs my point of adding this to the CI. They should be fixed after merging #235 as mentioned above.

Diiaablo95:
Should we add `--all-targets` also to `cargo test`? I guess we could go for `cargo test --all --all-targets` and `cargo test --all --all-features --all-targets`?

wischli:
> Should we add `--all-targets` also to `cargo test`? I guess we could go for `cargo test --all --all-targets` and `cargo test --all --all-features --all-targets`?

Yeah, that sounds good.

weichweich (approved):
my only objection is that this will increase the load on the server. We have 2 concurrent jobs at the moment. Maybe we should increase that to 3 or 4. 🤔

PR #242, opened 2021-08-22
chore: upgrade to 0.9.9

weichweich:
* update to 0.9.9
  * use `Contains` as call filter
  * explicitly allow all calls in tests
* adjust collator stake as requested
* fix /KILTprotocol/srlabs-audit/issues/1

## Checklist:


weichweich:
> Why dont we upgrade the CIs rust version instead of downgrading the one in the script?

CI and init script should have the same rust version now.

PR #246, opened 2021-09-01
chore: update DID benchmarks

wischli:
## fixes KILTProtocol/ticket#1542
* Fixes DID benchmark unit tests not working after #239
* Adds benchmarks and documentation for DID extrinsics introduced in #239

## TODO

- [x] Fix bench + unit tests for DIDs
- [x] Add missing benchmarks
- [x] Add documentation
- [x] Add weights to extrinsic weight definition

## Checklist:


pallets/did/src/benchmarking.rs
```diff
@@ -296,11 +295,475 @@ benchmarks! {
...
+
+		// fill up public keys to its max size
+		let mut did_details = generate_base_did_details::<T>(DidVerificationKey::from(old_did_public_auth_key));
+		assert_ok!(did_details.add_key_agreement_keys(get_key_agreement_keys::<T>(T::MaxNewKeyAgreementKeys::get()), block_number));
+		assert_ok!(did_details.update_attestation_key(DidVerificationKey::from(get_ed25519_public_attestation_key()), block_number));
+		assert_ok!(did_details.update_delegation_key(DidVerificationKey::from(get_ed25519_public_delegation_key()), block_number));
+
+		Did::<T>::insert(&did_subject, did_details);
+
+		let new_did_public_auth_key = get_ed25519_public_authentication_key();
```
Diiaablo95, line 313: Are we adding the same authentication key that was already set in `did_details`? Shouldn't we try to add a different one for the worst case? I am not sure how much it matters since probably what is more important is what key is used to verify the operation signature.

pallets/did/src/mock.rs
```diff
@@ -92,8 +92,9 @@ parameter_types! {
 	pub const MaxUrlLength: u32 = 200u32;
 	#[derive(Debug, Clone, PartialEq)]
 	pub const MaxTotalKeyAgreementKeys: u32 = 10u32;
+	// IMPORTANT: Needs to be at least MaxTotalKeyAgreementKeys + 3 (auth, delegation, attestation keys) for benchmarks!
 	#[derive(Debug, Clone)]
-	pub const MaxPublicKeysPerDid: u32 = 10u32;
+	pub const MaxPublicKeysPerDid: u32 = 13u32;
```
Diiaablo95, line 97:
```suggestion
	pub const MaxPublicKeysPerDid: u32 = MaxTotalKeyAgreementKeys + 3;
```

pallets/did/src/benchmarking.rs
```diff
@@ -296,11 +295,475 @@ benchmarks! {
...
 		let did_call_signature = ecdsa_sign(AUTHENTICATION_KEY_ID, &did_public_auth_key, did_call_op.encode().as_ref()).expect("Failed to create DID signature from raw ecdsa signature.");
 	}: submit_did_call(RawOrigin::Signed(submitter), Box::new(did_call_op), DidSignature::from(did_call_signature))
+
+	/* set_authentication_key extrinsic */
+	set_ed25519_authentication_key {
+		let block_number = T::BlockNumber::zero();
+		let old_did_public_auth_key = get_ed25519_public_authentication_key();
+		let did_subject: DidIdentifierOf<T> = MultiSigner::from(old_did_public_auth_key).into_account().into();
+
+		// fill up public keys to its max size
```
Diiaablo95, line 305: I would perhaps add that we are filling up because the # of max public keys = # of max key agreement keys + 3.

Diiaablo95 (approved):
LGMT! Just a couple of minor comments to make it clearer for future maintenance, as these tests and benchmarks are getting quite large 😅

pallets/did/src/benchmarking.rs
wischli: Good catch. In all benchmarks where we wanted to set a new key (authentication, attestation, delegation) I just generated the same key twice 🥶 Should be fixed by e35868ad80fdaf1075baa6328fe2c832fa406f95

pallets/did/src/mock.rs
wischli: Clippy complains about this except if I implement `Add` for `MaxTotalKeyAgreementKeys`. Also tried to add by using saturated into but that would also have to be implemented. Thus, I would prefer keeping it at 13 then except if you have a better solution.

Diiaablo95: Oh yeah, true. Nevermind then 👍

pallets/did/src/benchmarking.rs
wischli: --> a527303

wischli:
@weichweich Please review so we can merge this. Thanks!

weichweich (approved):

PR #251, opened 2021-09-13
fix: irremovability of stakers

wischli:
## fixes https://github.com/KILTprotocol/srlabs-audit/issues/3
* Reduces the maximum of manual unstake requests of candidates and delegators by one to have a placeholder for exiting or forced removals
* Blocks candidates and delegators from (re-) joining if they have max unstake requests. Otherwise they would make themselves unblockable by rejoining after exiting/being kicked with max unstake requests.
* Adds two unit tests for the candidate and delegator cases
* Fixes benchmarks

## Checklist:


weichweich:
What if someone was kicked as a delegator and then delegates somewhere else? Wouldn't he be irremovable again? Is that possible?

pallets/parachain-staking/src/tests.rs
```diff
@@ -2516,6 +2516,67 @@ fn unlock_unstaked() {
...
+			assert_noop!(
+				StakePallet::delegator_stake_less(Origin::signed(5), 1, 1),
+				Error::<Test>::NoMoreUnstaking
+			);
+
+			// Fill last unstake request by replacing delegator
+			assert_ok!(StakePallet::join_delegators(Origin::signed(6), 1, 200));
+			assert_eq!(StakePallet::unstaking(5).into_inner().len(), max_unstake_reqs + 1);
+			assert!(!StakePallet::is_delegator(&5));
+		});
```
weichweich, line 2577: The last assert will fail. Maybe we need to add a check to `join_delegators` that at least 1 free unstaking slot is left?

```suggestion
			// Rejoin and kick again
			assert_ok!(StakePallet::join_delegators(Origin::signed(5), 1, 201));
			assert!(StakePallet::is_delegator(&5));
			assert_ok!(StakePallet::join_delegators(Origin::signed(6), 1, 202));
			assert!(!StakePallet::is_delegator(&5));
		});
```

weichweich (changes requested):
We probably need another check.

pallets/parachain-staking/src/tests.rs
wischli: > The last assert will fail. Maybe we need to add a check to `join_delegators` that at least 1 free unstaking slot is left?

Yes I was also thinking of this.

Diiaablo95 (approved):
+1 for what Albi found, but otherwise it looks fine to me. I just don't really like that we always use `T::MaxUnstakeRequests::get()` when calling `prep_unstake`, so perhaps we could make that part of the function itself?

wischli:
> I just don't really like that we always use `T::MaxUnstakeRequests::get()` when calling `prep_unstake`, so perhaps we could make that part of the function itself?

We are calling it with
* `T::MaxUnstakeRequests::get()` if a candidate wants to exit or a candidate/delegator shall be kicked
* `T::MaxUnstakeRequests::get().saturating_sub(1)` if a candidate/delegator adjusts their stake

We could also replace this input parameter of `prep_unstake` with a boolean `is_exit` which then handles the substitution of either `0` or `1`. But I thought this would be worse. What do you guys think?

pallets/parachain-staking/src/tests.rs
wischli: Please check 19b7c61

pallets/parachain-staking/src/lib.rs
```diff
@@ -1310,7 +1327,7 @@ pub mod pallet {
 			);
 
 			// we don't unlock immediately
-			Self::prep_unstake(&collator, less)?;
+			Self::prep_unstake(&collator, less, T::MaxUnstakeRequests::get().saturating_sub(1))?;
```
weichweich, line 1330: a comment for why we do the sub 1 might be good. this is not immediately obvious i think.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1794,7 +1815,7 @@ pub mod pallet {
 				Error::<T>::NomStakeBelowMin
 			);
 
-			Self::prep_unstake(&delegator, less)?;
+			Self::prep_unstake(&delegator, less, T::MaxUnstakeRequests::get().saturating_sub(1))?;
```
weichweich, line 1818: here too

weichweich (commented):
[*W*](https://careerswitchtocoding.com/blog/moscow-the-best-code-review-technique-youre-not-using): a comment to explain the sub(1) might be nice

Diiaablo95:
Yes I see that it is called with different values, I was just wondering whether there would be a better way to avoid repetition. But I guess it is fine.

pallets/parachain-staking/src/lib.rs
wischli: Based on @Diiaablo95 's discomfort I refactored the limit param of `prep_unstake` in 0e67cbd. I prefer this and hope so do you.

weichweich (approved):
LGTM

W: I don't like`T::MaxUnstakeRequests::get().saturating_sub((!is_removal).into())`. 😂

wischli:
> LGTM
> 
> W: I don't like`T::MaxUnstakeRequests::get().saturating_sub((!is_removal).into())`. I think the flow is not really clear here. Why are we substracting `is_removal`. but the comment makes it clear so it's fine...

How would you do it?

weichweich:
> How would you do it?

something like that:

```rust

let allowed_unstakings = if is_removal {
  // the account was forcefully removed and we allow to fill all unstake requests
  T::MaxUnstakeRequests::get()
} else {
  // we need to reserve a free slot for a forceful removal of the account
  T::MaxUnstakeRequests::get().saturating_sub(1)
};
```

wischli:
> > How would you do it?
> 
> something like that:
> 
> ```rust
> let allowed_unstakings = if is_removal {
>   // the account was forcefully removed and we allow to fill all unstake requests
>   T::MaxUnstakeRequests::get()
> } else {
>   // we need to reserve a free slot for a forceful removal of the account
>   T::MaxUnstakeRequests::get().saturating_sub(1)
> };
> ```

Interesting. This was my initial draft but I thought you preferred less lines of code 🥲

PR #253, opened 2021-09-17
fix: Collator Storage

weichweich:
## Checklist:

### DIFF
--- a/pallets/parachain-staking/src/migrations/v5.rs
+++ b/pallets/parachain-staking/src/migrations/v5.rs
@@ -44,8 +44,23 @@ pub(crate) fn migrate<T: Config>() -> Weight {
 
 #[cfg(feature = "try-runtime")]
 pub(crate) fn post_migrate<T: Config>() -> Result<(), &'static str> {
+	use crate::TopCandidates;
+	use sp_runtime::SaturatedConversion;
+
 	assert_eq!(StorageVersion::<T>::get(), StakingStorageVersion::V5);
-	assert!(CandidateCount::<T>::get() > T::MinCollators::get());
+	log::info!(
+		"CandidateCount = {} >= {} = MinCollators",
+		CandidateCount::<T>::get(),
+		T::MinCollators::get()
+	);
+	assert!(CandidateCount::<T>::get() >= T::MinCollators::get());
+	log::info!(
+		"TopCandidates = {} >= {} = MinCollators",
+		TopCandidates::<T>::get().len(),
+		T::MinCollators::get()
+	);
+	assert!(TopCandidates::<T>::get().len().saturated_into::<u32>() >= T::MinCollators::get());
+	assert!(TopCandidates::<T>::get().len().saturated_into::<u32>() == CandidateCount::<T>::get());
 	Ok(())
 }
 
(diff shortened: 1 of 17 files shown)


wischli (approved):
LGTM!

pallets/parachain-staking/src/migrations/v5.rs
```diff
@@ -31,10 +41,30 @@ pub(crate) fn migrate<T: Config>() -> Weight {
 	// Kill selected candidates list
 	old::SelectedCandidates::<T>::kill();
 
-	// count candidates
-	let counter: u32 = CandidatePool::<T>::iter().fold(0, |acc, _| acc.saturating_add(1));
+	// migrate CandidatePool to TopCandidates
```
Diiaablo95, line 44: Q: There's no added weight for all these reads/writes that we do now?

pallets/parachain-staking/src/migrations/v5.rs
```diff
@@ -31,10 +41,30 @@ pub(crate) fn migrate<T: Config>() -> Weight {
...
+
+	// copy & count candidates
+	let counter: u32 = old::CollatorState::<T>::iter().fold(0, |acc, (key, candidate)| {
+		CandidatePool::<T>::insert(&key, candidate);
+		acc.saturating_add(1)
+	});
 	CandidateCount::<T>::put(counter);
 
+	// manually kill the ParachainStaking + CollatorState storage.
+	unhashed::kill_prefix(
```
Diiaablo95, line 63: C: Just saying, would it be easier to read also afterwards to add the old types to the deprecated module and call `kill` on them? If they are still in the storage, the kill should still take effect. Plus, by having a type defined we can also make sure that in `pre_upgrade` those values exist and in `post_upgrade` that they don't exist anymore.

Diiaablo95 (commented):
Overall looks good. I am not entirely sure that deleting the storage by raw hex is the best approach. I would rather retroactively add the types to remove and check in pre and post upgrade they correct deletion.

pallets/parachain-staking/src/migrations/v5.rs
weichweich: good point completely forgot the weights...

pallets/parachain-staking/src/migrations/v5.rs
weichweich: storagemaps don't have kill. The storage is already in the old `mod`

Diiaablo95: There's a `drain` method, wouldn't that be enough to clear the old storage?

Diiaablo95: I am just saying that we should probably avoid using these hard-coded hex. Even just calling some hex-encoding function on the storage item name would be a better solution.

wischli: I'm not sure `drain` would also kill the prefix. It only drains the entries.

weichweich: don't know who else that works

weichweich: ah drain would be enough. didn't see that!

weichweich: yeah i also don'T like the hardcoded hex

pallets/parachain-staking/src/migrations/v5.rs
weichweich: should be fixed

Diiaablo95 (approved):
🏆

PR #256, opened 2021-09-21
feat: upgrade to polkadot-v0.9.10

wischli:
## fixes KILTProtocol/ticket#1592
* Updates to latest Polkadot version 0.9.10

Note: PR is set to draft because 0.9.10 won't be released before October as it seems (see Element)

## Checklist:


weichweich (approved):

weichweich:
merging this now since we need a fix that will be added by the upgrade.

@wischli @Diiaablo95 we need to remember to `cargo update -p sp-io` etc. before we release.

PR #254, opened 2021-09-20
feat: storage deposits attestation

weichweich:
### custom types

This PR introduces new custom JS-types which are required for compatibility with [our SDK](https://github.com/KILTprotocol/sdk-js) and the [Polkadot Apps](https://polkadot.js.org/apps/#/extrinsics). Please use the following types to test the code with the Polkadot Apps:

<details>
  <summary>JS-Types</summary>
  
  ```json
  {}
  ```
</details>

## Checklist:

- [ ] I have verified that the code works
  - [ ] No panics! (checked arithmetic ops, no indexing `array[3]` use `get(3)`, ...)
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] This PR does not introduce new custom types
  - [ ] If not, I have opened a companion PR with the type changes in the [KILT types-definitions repository](https://github.com/KILTprotocol/type-definitions/pulls)
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #258, opened 2021-09-29
feat: add delegation storage deposits

wischli:
## fixes https://github.com/KILTprotocol/srlabs-audit/issues/6
* Adds a deposit as a `reserve` to a delegation which will be paid by the user who delegates (invites another user)
* Adds `deposit` directly to the `DelegationNode` struct (as is currently WIP for attestations in #254)
* The delegator can remove the delegation at any point (it does not have to be revoked) which will `unreserve` the deposit
  * Added a new `remove` extrinsic similar to `revoke`
    * Difference: We do not need to check for `is_delegating` because we do not allow ancestors to be able to directly remove any of their descendants (only indirectly if they want to remove their node)
  * Upon removal, each sub-delegation is refunded to the corresponding payee as well (from children to the specified node)
    * After a child node is removed, the `children` property of the parent delegation node is updated in case the full removal process cannot be completed and stops prematurely
    * This should not create too much overhead because all storage changes are only committed at the end of the block. Until that point, every change is kept in memory

**_Note: ✅ #254 needs to be merged first because this PR is based on it._**

## TODO

- [x] Fixes after review
- [x] Re-check migrations
- [ ] Update DelegationNodes type
- [ ] Re-run benchmarks

## Checklist:


pallets/delegation/src/mock.rs
```diff
@@ -380,17 +468,28 @@ impl ExtBuilder {
 		self
 	}
 
+	// TODO: Remove optional input ext
 	pub fn build(self, ext: Option<sp_io::TestExternalities>) -> sp_io::TestExternalities {
 		let mut ext = if let Some(ext) = ext {
```
wischli, line 473: For now, I kept this such that I did not need to alter tests which I did not need to touch. This will be fixed by #261. But if you want, I can refactor all tests in this PR as well.

pallets/delegation/src/lib.rs
```diff
@@ -130,21 +143,36 @@ pub mod pallet {
 			Payload = Vec<u8>,
 			Signature = Self::Signature,
 		>;
-		type DelegationEntityId: Parameter;
+		// FIXME: Is there a better solution than restrict this to
+		// Into<AccountIdOf<Self>> to enable conversion between both types?
+		type DelegationEntityId: Parameter + Into<AccountIdOf<Self>>;
```
weichweich, line 148: there should be no need to convert between the two?
DelegationEntityId -> The DID that creates, revokes and owns Delegatoins
AccountId -> the sender of the TX that pays for everything

pallets/attestation/src/lib.rs
```diff
@@ -127,7 +127,7 @@ pub mod pallet {
 		type Currency: Currency<AccountIdOf<Self>> + ReservableCurrency<AccountIdOf<Self>>;
 
 		/// The deposit that is required for storing an attestation.
-		#[pallet::constant]
+		// FIXME: Add back #[pallet::constant] once the fix is merged
```
weichweich, line 130: i would rather comment out the delegation pallet constant attribute. Attestations will be used more than delegations

pallets/delegation/src/lib.rs
```diff
@@ -481,6 +538,72 @@ pub mod pallet {
...
+			let invoker = source.subject();
+
+			let delegation = <DelegationNodes<T>>::get(&delegation_id).ok_or(Error::<T>::DelegationNotFound)?;
+			let deposit_owner = delegation.deposit.owner;
+
+			// Node can only be removed by deposit or node owner, not the parent or another
+			// ancestor
+			ensure!(
+				deposit_owner == sender || deposit_owner == invoker.clone().into(),
+				Error::<T>::UnauthorizedRemoval
```
weichweich, line 582: DID will never own the deposit. For the invoker we should check `is_delegating`?

```suggestion
				deposit_owner == sender,
				Error::<T>::UnauthorizedRemoval
```

pallets/delegation/src/lib.rs
```diff
@@ -481,6 +538,72 @@ pub mod pallet {
...
+		pub fn remove_delegation(
+			origin: OriginFor<T>,
+			delegation_id: DelegationNodeIdOf<T>,
+			max_removals: u32,
+		) -> DispatchResultWithPostInfo {
+			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
+			let sender = source.sender();
+			let invoker = source.subject();
+
+			let delegation = <DelegationNodes<T>>::get(&delegation_id).ok_or(Error::<T>::DelegationNotFound)?;
```
weichweich, line 575:
```suggestion
			let delegation = DelegationNodes::<T>::get(&delegation_id).ok_or(Error::<T>::DelegationNotFound)?;
```

Diiaablo95 (changes requested):
Overall it looks good. There are some comments about stuff I am not 100% sure we want to tackle in this PR, and some comments about tests.

pallets/delegation/src/lib.rs
```diff
@@ -505,18 +628,22 @@ impl<T: Config> Pallet<T> {
...
-		<DelegationNodes<T>>::insert(root_id, root_node.clone());
+		deposit_owner: AccountIdOf<T>,
+	) {
+		let root_node = DelegationNode::new_root_node(
+			root_id,
+			DelegationDetails::default_with_owner(hierarchy_owner),
+			deposit_owner,
+		);
+		<DelegationNodes<T>>::insert(root_id, root_node);
 		<DelegationHierarchies<T>>::insert(root_id, hierarchy_details);
```
weichweich, line 646:
```suggestion
		DelegationNodes::<T>::insert(root_id, root_node);
		DelegationHierarchies::<T>::insert(root_id, hierarchy_details);
```

pallets/delegation/src/lib.rs
```diff
@@ -658,4 +785,96 @@ impl<T: Config> Pallet<T> {
...
+
+				Self::remove(child, remaining_removals).map(|(r, w)| {
+					removals = removals.saturating_add(r);
+					consumed_weight = consumed_weight.saturating_add(w);
+				})?;
+
+				// Remove child from set and update parent node in case of pre-emptive stops due
+				// to insufficient removal gas
+				delegation_node.children.remove(child);
+				<DelegationNodes<T>>::insert(delegation, delegation_node.clone());
```
weichweich, line 824:
```suggestion
				DelegationNodes::<T>::insert(delegation, delegation_node.clone());
```

pallets/delegation/src/lib.rs
```diff
@@ -658,4 +785,96 @@ impl<T: Config> Pallet<T> {
...
+	/// which is bounded by `max_children`.
+	/// - Writes: C * Delegations
+	/// - Reads: C * Delegations
+	/// # </weight>
+	fn remove_children(delegation: &DelegationNodeIdOf<T>, max_removals: u32) -> Result<(u32, Weight), DispatchError> {
+		let mut removals: u32 = 0;
+		let mut consumed_weight: Weight = 0;
+
+		// Can't clear storage until we have reached a leaf
+		if let Some(mut delegation_node) = <DelegationNodes<T>>::get(delegation) {
```
weichweich, line 806:
```suggestion
		if let Some(mut delegation_node) = DelegationNodes::<T>::get(delegation) {
```

pallets/delegation/src/lib.rs
```diff
@@ -658,4 +785,96 @@ impl<T: Config> Pallet<T> {
...
+	/// - Reads: 2 * C * Delegations, C * Balance
+	/// - Writes: C * Delegations, C * Balance
+	/// # </weight>
+	fn remove(delegation: &DelegationNodeIdOf<T>, max_removals: u32) -> Result<(u32, Weight), DispatchError> {
+		let mut removals: u32 = 0;
+		let mut consumed_weight: Weight = 0;
+
+		// Retrieve delegation node from storage
+		// Storage removal has to be postponed until children have been removed
+		let delegation_node = <DelegationNodes<T>>::get(*delegation).ok_or(Error::<T>::DelegationNotFound)?;
```
weichweich, line 846:
```suggestion
		let delegation_node = DelegationNodes::<T>::get(*delegation).ok_or(Error::<T>::DelegationNotFound)?;
```

pallets/delegation/src/lib.rs
```diff
@@ -658,4 +785,96 @@ impl<T: Config> Pallet<T> {
...
+		})?;
+
+		// If we run out of removal gas, we only remove children. The tree will be
+		// changed but is still valid.
+		ensure!(removals < max_removals, Error::<T>::ExceededRemovalBounds);
+
+		// *** No Fail beyond this point ***
+
+		// We can clear storage now that all children have been removed
+		<DelegationNodes<T>>::remove(*delegation);
```
weichweich, line 863:
```suggestion
		DelegationNodes::<T>::remove(*delegation);
```

pallets/delegation/src/migrations/v1.rs
```diff
@@ -16,9 +16,18 @@
...
+use crate::{
+	deprecated::{
+		v1::storage::{Children, Delegations, Roots},
+		v2::{storage::DelegationNodes, DelegationNode as DelegationNodeV2},
+	},
+	migrations::DelegationStorageVersion,
+	Config, DelegationDetails, DelegationHierarchies, DelegationHierarchyDetails, DelegationNodeIdOf, Permissions,
+	StorageVersion, Weight,
+};
+use frame_support::{
```
weichweich, line 28: Did you spent time on making the old migrations work?
Parity deletes old migrations after a while. Maybe we should start doing the same. We cannot support old migrations infinity anyways

pallets/delegation/src/lib.rs
```diff
@@ -306,12 +350,17 @@ pub mod pallet {
 				<ctype::Error<T>>::CTypeNotFound
 			);
 
+			CurrencyOf::<T>::reserve(&payer, deposit_amount)?;
```
weichweich, line 353: M(reserve-deposit part 1): I would like to have the reserve call connected to the Deposit amount that we save in the Hierarchy.
At the moment you must always call `reserve` before you call `create_and_store_new_hierarchy`. If you forget to reserve, the deposit and what's actually reserved diverged.

pallets/delegation/src/lib.rs
```diff
@@ -387,6 +439,10 @@ pub mod pallet {
 				Error::<T>::UnauthorizedDelegation
 			);
 
+			CurrencyOf::<T>::reserve(&payer, deposit_amount)?;
```
weichweich, line 442: M(reserve-deposit part 2): same here, i would reserve inside the `store_delegation_under_parent`

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -89,12 +102,17 @@ impl<T: Config> DelegationNode<T> {
...
+		deposit_owner: AccountIdOf<T>,
 	) -> Self {
 		Self {
 			hierarchy_root_id,
 			parent: Some(parent),
 			children: BoundedBTreeSet::<DelegationNodeIdOf<T>, T::MaxChildren>::new(),
 			details,
+			deposit: Deposit::<AccountIdOf<T>, BalanceOf<T>> {
+				owner: deposit_owner,
+				amount: <T as Config>::Deposit::get(),
```
weichweich, line 114: M(reserve-deposit part 3): if we hard code the deposit here, we never made sure that the deposit was reserved somewhere. So we might end up to store a deposit without reserving.

pallets/delegation/src/delegation_hierarchy.rs
```diff
@@ -69,17 +71,28 @@ pub struct DelegationNode<T: Config> {
...
+		deposit_owner: AccountIdOf<T>,
+	) -> Self {
 		Self {
 			hierarchy_root_id: id,
 			parent: None,
 			children: BoundedBTreeSet::<DelegationNodeIdOf<T>, T::MaxChildren>::new(),
 			details,
+			deposit: Deposit::<AccountIdOf<T>, BalanceOf<T>> {
+				owner: deposit_owner,
+				amount: <T as Config>::Deposit::get(),
```
weichweich, line 94: M(reserve-deposit part 4): if we hard code the deposit here, we never made sure that the deposit was reserved somewhere. So we might end up to store a deposit without reserving.

weichweich (changes requested):
Nice PR! Very on point without much sidetracking! 🚀 

I have a few requests:

* DelegationEntityId should not be needed to be converted to an AccountId?
* We should always make sure that if we store a deposit we also reserved something
* I would rather like to see the constant attribute removed from the delegations than the attestations because attestations will be a core feature while delegations might be used by 1%

I think you put time into old migrations? I think we should start deleting old migrations, but we need to discuss this later.

pallets/attestation/src/lib.rs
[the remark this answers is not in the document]
weichweich: https://github.com/KILTprotocol/substrate/commit/f15c85e09b56a9d7a317e856ae324c47ea924ece

Since there is a deposit in Delegations and Attestations directly depends on delegations, it's not clear which deposit should be used inside the macro

pallets/delegation/src/lib.rs
Diiaablo95: Yes, that is actually a better point. Not sure why I added it there in the first place, but it would make sense to remove it.

pallets/attestation/src/lib.rs
Diiaablo95: Are we assuming the same deposit amount for both? That is why we need to get rid of one of them?

pallets/delegation/src/lib.rs
Diiaablo95: I think we are only allowing the direct owner of a node, so probably not `is_delegating` but a simple check for the delegation owner == DID.

pallets/attestation/src/lib.rs
weichweich: There will be different deposits. Attestations should be cheap, delegations more expensive.

pallets/delegation/src/lib.rs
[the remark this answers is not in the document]
wischli: I was just wondering whether there is a more elegant solution than adding `+ Into<AccountIdOf<Self>>`.

pallets/attestation/src/lib.rs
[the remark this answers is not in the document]
weichweich: https://github.com/paritytech/substrate/pull/9942

pallets/delegation/src/migrations/v1.rs
wischli: Yeah it took me quite some time because of just importing `use crate::*` which caused a bug without a helping error message. I would be super happy to just get rid of it.

pallets/delegation/src/deprecated.rs
```diff
@@ -101,3 +101,60 @@ pub(crate) mod v1 {
 		}
 	}
 }
+
+pub(crate) mod v2 {
```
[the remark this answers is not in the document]
wischli: I would propose to just get rid of the v1 migration instead. Else, I agree.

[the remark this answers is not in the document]
Diiaablo95: I would personally not remove any migrations until we have added the KILT features to spiritnet. After that, I would be okay with removing them.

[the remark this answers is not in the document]
weichweich: The migrations should not get executed on spiritnet. Why keep them?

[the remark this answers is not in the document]
wischli: Why though? When we add them to the Spiritnet, they come fresh and are independent of migrations anyway. I'd rather start fresh.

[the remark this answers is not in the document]
wischli: We should always have the most recent storage version on all of our networks. So we only need to support migrations from `n-1` to `n` if `n` is the most recent version. Once we add `n+1` we remove `n-1`.

I would not reset the StorageVersion but keep the current versioning.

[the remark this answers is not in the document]
Diiaablo95: Yes, that is my point. We can of course remove the old migrations as also Parity does, but we need to establish some process about how we are going to upgrade. So I was indirectly referring to this, since we do not have yet a defined way of making upgrades.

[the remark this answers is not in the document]
weichweich: The process is to not skip runtime upgrades. The migrations are in the runtime

pallets/delegation/src/delegation_hierarchy.rs
wischli: Not sure what your solution proposal would look like after moving the reserving into `store_delegation_under_parent` and `create_and_store_new_hierarchy`.

wischli: Please have a look at f384e3.

pallets/delegation/src/lib.rs
wischli: f384e39

pallets/delegation/src/benchmarking.rs
```diff
@@ -60,6 +64,11 @@ where
 	let ctype_hash = <T::Hash as Default>::default();
 	let hierarchy_root_id = generate_delegation_id::<T>(number);
 
+	<T as Config>::Currency::make_free_balance_be(
+		&root_acc,
+		<T as Config>::Deposit::get() + <T as Config>::Deposit::get(),
```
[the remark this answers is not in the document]
wischli: 1d6ae04

pallets/delegation/src/lib.rs
```diff
@@ -481,6 +538,72 @@ pub mod pallet {
 			)
 			.into())
 		}
+
+		/// Remove a delegation node (potentially a root node) and all its
```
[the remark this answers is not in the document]
wischli: 5730c36

pallets/delegation/src/tests.rs
```diff
@@ -671,46 +680,77 @@ fn list_hierarchy_revoke_root_successful() {
...
+				get_origin(revoker.clone()),
+				operation.id,
+				operation.max_children
+			));
+
+			assert!(Delegation::delegation_nodes(&operation.id).is_none());
+			assert!(Delegation::delegation_hierarchies(&hierarchy_root_id).is_none());
+			assert!(Delegation::delegation_nodes(&parent_id).is_none());
+			assert!(Delegation::delegation_nodes(&delegation_id).is_none());
+			assert!(Balances::reserved_balance(revoker.clone()).is_zero());
```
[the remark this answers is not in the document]
wischli: c56169c

pallets/delegation/src/mock.rs
```diff
@@ -370,6 +448,16 @@ impl ExtBuilder {
 		self
 	}
 
+	pub fn with_balances(mut self, balances: Vec<(AccountIdOf<Test>, BalanceOf<Test>)>) -> Self {
+		self.balances = balances;
+		self
+	}
+
+	pub fn with_ctypes(mut self, ctypes: Vec<(TestCtypeHash, TestCtypeOwner)>) -> Self {
```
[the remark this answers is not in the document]
wischli: 0dbb03

pallets/delegation/src/mock.rs
```diff
@@ -357,6 +422,19 @@ pub struct ExtBuilder {
 	storage_version: DelegationStorageVersion,
 }
 
+impl Default for ExtBuilder {
```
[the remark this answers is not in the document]
wischli: 0dbb03b

pallets/delegation/src/mock.rs
```diff
@@ -341,13 +387,32 @@ pub fn initialize_pallet<T: Config>(
...
 }
 
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub struct ExtBuilder {
+	/// endowed accounts with balances
+	balances: Vec<(AccountIdOf<Test>, BalanceOf<Test>)>,
+	/// initial ctypes & owners
+	ctypes: Vec<(TestCtypeHash, AccountIdOf<Test>)>,
+	ctype_builder: Option<ctype_mock::ExtBuilder>,
```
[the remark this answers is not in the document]
wischli: 0dbb03b

pallets/delegation/src/migrations/v1.rs
```diff
@@ -16,9 +16,18 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
-use crate::*;
-
-use frame_support::{storage::bounded_btree_set::BoundedBTreeSet, IterableStorageMap, StorageMap, StoragePrefixedMap};
+use crate::{
+	deprecated::{
+		v1::storage::{Children, Delegations, Roots},
```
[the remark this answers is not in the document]
wischli: As discussed, we will handle this in another PR.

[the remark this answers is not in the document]
wischli: As discussed, we will handle this in another PR.

pallets/delegation/src/tests.rs
```diff
@@ -655,7 +664,7 @@ fn empty_revoke_root_successful() {
 }
 
 #[test]
-fn list_hierarchy_revoke_root_successful() {
+fn list_hierarchy_revoke_and_remove_root_successful() {
```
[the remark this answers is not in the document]
wischli: IMO `remove_single_hierarchy` and `remove_children_gas_runs_out` are sufficient in this regard?

pallets/delegation/src/mock.rs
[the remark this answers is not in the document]
wischli: 0dbb03b

pallets/delegation/src/tests.rs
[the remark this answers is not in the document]
Diiaablo95: M: If we test the conditions here, we are basically testing that the mock works, right? I would add them in the creation test cases for both root and regular delegation nodes right after the node has been created, to test that the right balance has been reserved. The place for those tests could be, for instance, in `create_root_delegation_successful()` and `create_delegation_direct_root_successful()`. In the same way, we should test that the right amount is unreserved when a node is revoked, as there is no distinction between root and regular nodes in revocation/removal.

pallets/delegation/src/tests.rs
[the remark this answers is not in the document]
Diiaablo95: Yes, I missed those.

pallets/delegation/src/delegation_hierarchy.rs
weichweich: The commented out version looks like the thing I meant. 👍🏻

pallets/attestation/src/lib.rs
weichweich: doesn't work, leave as it is.

pallets/delegation/src/tests.rs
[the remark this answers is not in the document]
wischli: Please check a8bca6a. There are more checks which already exist in other tests. IMO this should suffice.

pallets/delegation/src/delegation_hierarchy.rs
wischli: Great! Hope this issue is resolved by fdbf92d then?

weichweich: Love it!

Diiaablo95 (approved):
👌

PR #264, opened 2021-10-12
feat: add crowdloan contributions pallet

Diiaablo95:
## fixes KILTProtocol/ticket#1587

This PR allows to register on the KILT blockchain the KILT crowdloan contributions.

The new `kilt-crowdloan` pallet exposes the following extrinsics:

- `set_admin_account(origin, new_account)`
- `set_new_contribution(origin, contributor_account, amount)`
- `remove_contribution(origin, contributor_account)`

## How to test:

Run `cargo test -p kilt-crowdloan`.

## Checklist

- [ ] Ask for internal review and integrate changes
- [ ] Open types PR

### Custom types

This PR introduces new custom JS-types which are required for compatibility with [our SDK](https://github.com/KILTprotocol/sdk-js) and the [Polkadot Apps](https://polkadot.js.org/apps/#/extrinsics). Please use the following types to test the code with the Polkadot Apps:

<details>
  <summary>JS-Types</summary>

  ```json
  {}
  ```
</details>

## Checklist:


pallets/crowdloan/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+			// Change admin with sudo account
+			assert_ok!(Crowdloan::set_admin_account(Origin::root(), new_admin.clone()));
+
+			// Test new admin is the one set
+			assert_eq!(Crowdloan::admin_account(), new_admin);
+		});
+}
+
+#[test]
+fn test_no_admin_set() {
```
wischli, line 68: The test name confused me so I had to go back to the mock to check the default admin setting.

It appears that the derived `default` implementation for the admin account is indeed `ACCOUNT_00`? Else the first assert would fail. Not sure what to think about this hidden magic. In the end, it's just a test and does not matter too much. Still wanted to raise this.

W: Rename it.
```suggestion
fn test_no_custom_admin_set {
```

runtimes/standalone/src/migrations/crowdloan_contributions.rs
```diff
@@ -0,0 +1,70 @@
...
+use kilt_primitives::AccountId;
+
+// Same as Mashnet transfer account.
+pub const NEW_ADMIN_ACCOUNT: [u8; 32] = hex!("6a3c793cec9dbe330b349dc4eea6801090f5e71f53b1b41ad11afb4a313a282c");
+
+pub struct CrowdloanContributionsSetup;
+
+impl OnRuntimeUpgrade for CrowdloanContributionsSetup {
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
```
wischli, line 33: I suppose we have separate migrations because of different admin accounts. There should still be a more elegant way. After re-checking the admin accounts here and realizing these are the corresponding transfer accounts, we should be able to do this quite simple:

S: Move migration to crowdloan pallet
* `pre_upgrade` is the same for all runtimes
*  `on_runtime_upgrade` and `post_upgrade` differ in the admin account, which can be retrieved by querying the storage for the current `transfer_account`.

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+
+			Ok(())
+		}
+
+		/// Sets a new contribution amount for a given contributor's account.
+		///
+		/// If a previous contribution is present, it is overridden.
+		///
+		/// The dispatch origin must be the current admin account.
+		///
```
wischli, line 184: M: Missing the emitted event.

```suggestion
		///
		/// Emits `NewContributionSet`.
		///
```

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+			Ok(())
+		}
+
+		/// Removes a contribution entry from the storage, if present.
+		///
+		/// It returns an error if there is no contribution for the given
+		/// contributor's account.
+		///
+		/// The dispatch origin must be the current admin account.
+		///
```
wischli, line 216: M: Missing the emitted event.

```suggestion
		///
		/// Emits `ContributionRemoved`.
		///
```

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+		/// The to-delete contribution is not present.
+		ContributorNotPresent,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Sets a new account as the admin of this pallet.
+		///
+		/// The dispatch origin can be either Sudo or the current admin account.
+		///
```
wischli, line 150: M: Missing the emitted event.

```suggestion
		///
		/// Emits `NewAdminAccountSet `.
		///
```

wischli (changes requested):
Overall looks good except for the three permutations of migrations which only differ in the admin account address AFAICT. 

Apart from that, I noticed you forgot to mention the emitted events in the extrinsics documentation.

runtimes/standalone/src/migrations/crowdloan_contributions.rs
Diiaablo95: I was also going to implement the migration within the pallet, but @weichweich suggested to implement it at the runtime level. I had no strong opinion about it honestly, with the difference that if we do it at the pallet level, we do not know what is the admin level that **should** be set in the migration, while we know that in the runtime (even though we are not able to distinguish between different deployments of the same runtime, e.g., WILT and Spiritnet).

weichweich: If we move it into the pallet we could query the transfer_account. But then we would need to have the `kilt-launch` pallet as a dependency.

I would say, since we will remove the migrations rather soon, this should be fine.

weichweich: I would remove the migration from the standalone runtime. THere is no network that needs migrating.

pallets/crowdloan/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+pallet-balances = {branch = "polkadot-v0.9.10", default-features = false, git = "https://github.com/paritytech/substrate"}
+serde = {version = "1.0.101"}
+sp-keystore = {branch = "polkadot-v0.9.10", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-io = {branch = "polkadot-v0.9.10", default-features = false, git = "https://github.com/paritytech/substrate"}
+
+[dependencies]
+serde = {optional = true, version = "1.0.101"}
+
+# Internal dependencies
+kilt-primitives = {default-features = false, optional = true, path = "../../primitives"}
```
weichweich, line 26: This is not needed

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+//! # Crowdloan rewards Pallet
```
weichweich, line 19: M: This is not a reward pallet. It only registers who contributed how much. I think legal would be unhappy about the wording. At least check with Gustav.

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//!   contribution entries into/from this pallet's storage.
+//! - `set_new_contribution` - Add or replace a crowdload contribution, which
+//!   contains the contributor's address and the contributed amount.
+//! - `remove_contribution` - Remove a contribution entry from the pallet
+//!   storage.
+//!
+//! ## Genesis config
+//!
+//! The crowdloan contributions pallet depends on the [`GenesisConfig`].
+//!
```
weichweich, line 40: S: Maybe add a comment on what this pallet requires from the genesis config. Similar to the sudo pallet:

> //! You need to set an initial superuser account as the sudo `key`.

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//! Provides means of registering the contributors to the KILT crowdloan.
+//!
+//! - [`Config`]
+//! - [`Call`]
+//! - [`Pallet`]
+//!
+//! ## Interface
+//!
+//! ### Dispatchable Functions
```
weichweich, line 29: W: I would remove this from here. Isn't this super redundant? We document the Calls further below. Why do that twice.

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+//! - At any time, there is one and only one admin account which can manage the
+//!   pallet storage.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(test)]
+mod tests;
+
+#[cfg(test)]
+pub(crate) mod mock;
```
weichweich, line 52: S: No need for `pub(crate)`

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+		traits::{Currency, StorageVersion},
+	};
+	use frame_system::{pallet_prelude::*, EnsureOneOf, EnsureRoot, EnsureSigned, WeightInfo};
+	use sp_runtime::{
+		traits::{BadOrigin, StaticLookup},
+		Either,
+	};
+
+	pub type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
+	pub type BalanceOf<T> = <<T as Config>::Currency as Currency<AccountIdOf<T>>>::Balance;
```
weichweich, line 73: S: those two don't need to be public. I would make them `pub(crate)´

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		pub admin_account: AccountIdOf<T>,
```
weichweich, line 97: C: I think admin is a strong word. A block chain should not have an admin. Maybe rename to `registrar`?

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+		/// \[contributor account, old amount (OPTIONAL), new amount\]
+		NewContributionSet(AccountIdOf<T>, Option<BalanceOf<T>>, BalanceOf<T>),
+		/// A contribution has been removed.
+		/// \[contributor account\]
+		ContributionRemoved(AccountIdOf<T>),
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// The to-delete contribution is not present.
```
weichweich, line 144: I would not assume that this error will always be called in the delete call

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+		/// Weight: O(1)
+		/// - Reads: [Origin Account], AdminAccount
+		/// - Writes: AdminAccount
+		#[pallet::weight(1)]
+		pub fn set_admin_account(
+			origin: OriginFor<T>,
+			new_account: <T::Lookup as StaticLookup>::Source,
+		) -> DispatchResult {
+			let who =
+				EnsureOneOf::<AccountIdOf<T>, EnsureRoot<AccountIdOf<T>>, EnsureSigned<AccountIdOf<T>>>::ensure_origin(
```
weichweich, line 166: W: Now that I see this we could rather use the council for this and make the origin configurable `EnsureOrigin` via the Config trait? Changing the account is definitly harder that way, but the process is more democratic.

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+					origin,
+				)?;
+
+			let old_account = AdminAccount::<T>::get();
+
+			if let Either::Right(signed_origin) = who {
```
weichweich, line 172: S: There is also [`EnsureSignedBy`](https://docs.substrate.io/rustdocs/latest/frame_system/struct.EnsureSignedBy.html). Maybe we can use this?

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+		/// The dispatch origin must be the current admin account.
+		///
+		/// Emits `NewContributionSet`.
+		///
+		/// # <weight>
+		/// Weight: O(1)
+		/// - Reads: [Origin Account], AdminAccount, Contributions
+		/// - Writes: Contributions
+		#[pallet::weight(1)]
+		pub fn set_new_contribution(
```
weichweich, line 197: W: Doesn't need to be new right, so maybe just call it `set_new_contribution`

pallets/crowdloan/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+			origin: OriginFor<T>,
+			contributor_account: <T::Lookup as StaticLookup>::Source,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(who == AdminAccount::<T>::get(), BadOrigin);
+
+			let looked_up_account = <T as frame_system::Config>::Lookup::lookup(contributor_account)?;
+			let old_amount = Contributions::<T>::mutate(&looked_up_account, |entry| {
+				let existing_entry = *entry;
```
weichweich, line 207: M: interesting way of doing things. What's your opinion on using [Option::replace](https://doc.rust-lang.org/std/option/enum.Option.html#method.replace) instead. 😜

pallets/crowdloan/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+			assert_eq!(Crowdloan::admin_account(), admin);
+
+			// Change admin
+			assert_ok!(Crowdloan::set_admin_account(
+				Origin::signed(admin.clone()),
+				new_admin.clone()
+			));
+
+			// Test new admin is the one set
+			assert_eq!(Crowdloan::admin_account(), new_admin);
```
weichweich, line 44: W: could we also test for the event?

```
let last_event = System::events().pop().expect("Event expected").event
```

pallets/crowdloan/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+		.with_contributions(vec![(contributor.clone(), contribution)])
+		.build()
+		.execute_with(|| {
+			assert_eq!(Crowdloan::contributions(&contributor), Some(contribution));
+			assert_ok!(Crowdloan::set_new_contribution(
+				Origin::signed(admin.clone()),
+				contributor.clone(),
+				new_contribution
+			));
+			assert_eq!(Crowdloan::contributions(&contributor), Some(new_contribution));
```
weichweich, line 142: W: would be cool to test here too that the event contains the old and new contribution

runtimes/peregrine/src/migrations.rs
```diff
@@ -0,0 +1,19 @@
+// KILT Blockchain – https://botlabs.org
```
weichweich, line 1: W: Since we won't have that many migrations i wouldn't move that into another module. A migration module should be enough. This will be removed rather sooner than later...

weichweich (changes requested):
Very clean code! Really like it. I only got two things that needs changing.

runtimes/standalone/src/migrations/crowdloan_contributions.rs
Diiaablo95: Yes it is just for quick prototyping and testing by spinning up a mashnet node, as the local parachain setup has some issues right now.

pallets/crowdloan/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/b34620daaa1a9c6144a077843986d4c3bfc4c4b6

pallets/crowdloan/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/2f67594b7d6e706d7b158b11881d44f968295082

pallets/crowdloan/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/7e18b4424602754004a8c33b0f8508105d1112a9

pallets/crowdloan/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/9f1dcc447c8d72f7429aea13ee5cfafd0e3d0046

pallets/crowdloan/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/8595ce54c3629bc629dc8fd25a292a7483ea6406

pallets/crowdloan/src/lib.rs
Diiaablo95: I tried to quickly play with it and did not get it to compile. Perhaps we can fix it in the PR for the second part of the pallet.

pallets/crowdloan/src/tests.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/264/commits/f744f4c9dcfb7587ae898b795cb5f0f3793701a6

wischli (approved):
LGTM. I would still be happy if the standalone migration would be removed, see [my comment](https://github.com/KILTprotocol/mashnet-node/pull/264/files#r733688530).

PR #265, opened 2021-10-12
feat: improve CI checks

Diiaablo95:
I keep catching Clippy warnings whenever cargo is run with a combination of features that is neither `--all-features` nor `default-features`, for instance with only `--features runtime-benchmarks`. So I thought having multiple checks in place might help to keep the code cleaner, and as all the checks are within the same CI step, it should not take THAT much longer for the checks to finish. 

If you think this is not a good idea, we can close it, but I though it would at least be worth discussing.

For reference, the [cargo check](https://doc.rust-lang.org/cargo/commands/cargo-check.html) and [cargo test](https://doc.rust-lang.org/cargo/commands/cargo-test.html) docs.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,36 +1,36 @@
 stages:
+  - check
   - test
   - build
 
-clippy:
+check:
   image: paritytech/ci-linux:9575dfcd-20210729
-  stage: test
+  stage: check
   timeout: 1 hours
   script:
     - rustup component add clippy --toolchain nightly
-    - cargo +nightly clippy --all-features --all-targets -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --no-default-features -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features try-runtime -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features runtime-benchmarks -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features mock -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features std -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings
 
 fmt:
   image: paritytech/ci-linux:9575dfcd-20210729
-  stage: test
+  stage: check
   timeout: 1 hours
   script:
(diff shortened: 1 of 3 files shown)


pallets/delegation/src/benchmarking.rs
```diff
@@ -26,7 +26,7 @@ use frame_support::{
 use frame_system::RawOrigin;
 use sp_core::{offchain::KeyTypeId, sr25519};
 use sp_io::crypto::sr25519_generate;
-use sp_runtime::MultiSignature;
+use sp_runtime::{traits::Zero, MultiSignature};
```
weichweich, line 29: why was this added?

.gitlab-ci.yml
```diff
@@ -1,36 +1,36 @@
 stages:
+  - check
```
weichweich, line 2: I would clearly vote against that. This will slow down the CI.

.gitlab-ci.yml
```diff
@@ -1,36 +1,36 @@
...
   timeout: 1 hours
   script:
     - rustup component add clippy --toolchain nightly
-    - cargo +nightly clippy --all-features --all-targets -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --no-default-features -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features try-runtime -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features runtime-benchmarks -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features mock -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --features std -- -D warnings
+    - cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings
```
weichweich, line 17: So we mostly miss unused import warnings. And the imports are only unused in very specific feature combinations? The disadvantages of a slow CI clearly outweigh the gains from me here. 😕

pallets/delegation/src/benchmarking.rs
Diiaablo95: This was a warning I caught, but it has been also addressed in https://github.com/KILTprotocol/mashnet-node/pull/265.

Diiaablo95:
Alright, as it is not useful in any way I will close this.

PR #266, opened 2021-10-12
fix: unwanted delegator state removal when tx fails

wischli:
## fixes KILTProtocol/ticket#1632
* Fixes unwanted removal of delegator state if `join_delegators` or `delegate_another_candidate` extrinsics failed due to failure in `increase_lock` (either insufficient balance or full unstaking list) by delaying removal until tx cannot fail anymore
* Adds balance checks to head of corresponding delegation extrinsics
* Adds corresponding unit test

### TODO

- [x] Write migration for delegator which does not have a `DelegatorState` anymore but is still delegating to a collator

## Checklist:

### DIFF
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -1458,7 +1458,7 @@ pub mod pallet {
 				Self::update_top_candidates(collator.clone(), old_total, new_total);
 			}
 
-			// ** no fail beyond this line *
+			// *** No Fail beyond this point ***
 
 			// update states
 			<CandidatePool<T>>::insert(&collator, state);
@@ -1603,6 +1603,8 @@ pub mod pallet {
 			if state.is_active() {
 				Self::update_top_candidates(collator.clone(), old_total, new_total);
 			}
+
+			// *** No Fail beyond this point ***
 
 			// Update states
 			<CandidatePool<T>>::insert(&collator, state);
@@ -2038,7 +2040,7 @@ pub mod pallet {
 				delegator,
 				collator,
 				delegator_stake,
-				new_total,
+				new_totx1l,
 			));
 			Ok(())
 		}
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -172,6 +172,7 @@ mod types;
 use frame_support::pallet;
 
 pub use crate::{default_weights::WeightInfo, pallet::*};
+use types::KickedDelegator;
 
 #[pallet]
 pub mod pallet {
@@ -1391,9 +1392,17 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let acc = ensure_signed(origin)?;
 			let collator = T::Lookup::lookup(collator)?;
+
+			// check balance
+			ensure!(
+				pallet_balances::Pallet::<T>::free_balance(acc.clone()) >= amount.into(),
+				pallet_balances::Error::<T>::InsufficientBalance
+			);
+
 			// first delegation
 			ensure!(<DelegatorState<T>>::get(&acc).is_none(), Error::<T>::AlreadyDelegating);
 			ensure!(amount >= T::MinDelegatorStake::get(), Error::<T>::NomStakeBelowMin);
+
 			// cannot be a collator candidate and delegator with same AccountId
 			ensure!(!Self::is_active_candidate(&acc).is_some(), Error::<T>::CandidateExists);
 			ensure!(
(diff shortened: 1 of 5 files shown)


pallets/parachain-staking/src/lib.rs
```diff
@@ -2012,19 +2045,49 @@ pub mod pallet {
 			Ok(())
 		}
 
-		fn kick_delegator(delegation: &StakeOf<T>, collator: &T::AccountId) -> DispatchResult {
+		/// Check for remaining delegations of the delegator which has been
+		/// removed from the given collator.
+		///
+		/// Returns the removed delegators address and
```
Diiaablo95, line 2051:
```suggestion
		/// Returns the removed delegator's address and
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -2135,10 +2198,17 @@ pub mod pallet {
 		/// bounded by `MaxDelegatorsPerCollator`.
 		/// - Reads/Writes: 0
 		/// # </weight>
+		#[allow(clippy::type_complexity)]
 		fn do_update_delegator(
 			stake: Stake<T::AccountId, BalanceOf<T>>,
 			mut state: Candidate<T::AccountId, BalanceOf<T>, T::MaxDelegatorsPerCollator>,
-		) -> Result<CandidateOf<T, T::MaxDelegatorsPerCollator>, DispatchError> {
+		) -> Result<
```
Diiaablo95, line 2205: M: We should also update the extrinsic doc to reflect the new return type.

pallets/parachain-staking/src/migrations/v6.rs
```diff
@@ -0,0 +1,72 @@
...
+	log::info!("Migrating staking to StakingStorageVersion::V6");
+
+	let mut reads = 0u64;
+	let mut writes = 1u64;
+
+	// iter candidate pool and check whether any delegator has a cleared state
+	for candidate in CandidatePool::<T>::iter_values() {
+		reads = reads.saturating_add(1u64);
+		for delegation in candidate.delegators.into_iter() {
+			// we do not have to mutate existing entries since MaxCollatorsPerDelegator = 1
```
Diiaablo95, line 41: Should we just have a pre-migrate check where we verify this invariant? Since we had a bug, I would not feel 100% secure the state is consistent, and if we call `.expect()` in the migration we should probably check for the condition in the pre-migration checks.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2012,19 +2045,49 @@ pub mod pallet {
...
-		fn kick_delegator(delegation: &StakeOf<T>, collator: &T::AccountId) -> DispatchResult {
+		/// Check for remaining delegations of the delegator which has been
+		/// removed from the given collator.
+		///
+		/// Returns the removed delegators address and
+		/// * Either the updated delegator state if other delegations are still
+		///   remaining
+		/// * Or `None`, signalling the delegator state should be cleared once
+		///   the transaction cannot fail anymore.
+		fn kick_delegator(
```
Diiaablo95, line 2056: S: Should we rename this perhaps to something like `prep_kick_delegator()` or similar? It is not making any storage changes anymore, so it is not in fact kicking a delegator.

pallets/parachain-staking/src/tests.rs
```diff
@@ -346,7 +346,7 @@ fn collator_exit_executes_after_delay() {
 		.with_balances(vec![
 			(1, 1000),
 			(2, 300),
-			(3, 100),
+			(3, 110),
```
Diiaablo95, line 349: Is this change to allow for transaction fees?

Diiaablo95 (approved):
LGTM! Nevertheless, my main point was inside the v6 migration where we assume storage could not be corrupted in the migrate function, and I think we should validate the assumption in a pre-migration check.

pallets/parachain-staking/src/migrations/v6.rs
```diff
@@ -0,0 +1,72 @@
...
+			if !DelegatorState::<T>::contains_key(delegation.owner.clone()) {
+				DelegatorState::<T>::insert(
+					delegation.owner,
+					Delegator::try_new(candidate.id.clone(), delegation.amount)
+						.expect("MaxCollatorsPerDelegator is greater than one"),
```
weichweich, line 46: I would not expect here, but just ignore the error, if it happens.

pallets/parachain-staking/src/lib.rs
wischli: 1db5f45

pallets/parachain-staking/src/migrations/v6.rs
wischli: 8bd21ae

pallets/parachain-staking/src/migrations/v6.rs
wischli: 8bd21ae

pallets/parachain-staking/src/migrations/v6.rs
weichweich: A panic there means that we don't build the block. How would we recover from that? Will we be able to build the next block without the panic? Will we be able to build a block ever again? It will always panic at that point.

Diiaablo95: Ta-daaaaa!

wischli: 22ff8d0

weichweich (approved):

PR #285, opened 2021-10-28
feat: remove sudo

weichweich:
* try runtime doesn't go through because of the vesting pallet

[M](https://kilt-cast.monday.com/boards/1792114080/pulses/1792582331)

## Checklist:


runtimes/spiritnet/src/migrations/remove_sudo.rs
```diff
@@ -0,0 +1,64 @@
...
+pub struct RemoveSudo;
+
+impl OnRuntimeUpgrade for RemoveSudo {
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		use kilt_primitives::AccountId;
+
+		log::info!("Pre check Sudo-Removal.");
+		let res = frame_support::storage::unhashed::get::<AccountId>(&hex![
+			"5c0d1176a568c1f92944340dbfed9e9c530ebca703c85910e7164cb7d1c9e47b"
```
Diiaablo95, line 32: I would probably derive this value using some function? Like https://docs.substrate.io/rustdocs/latest/frame_support/storage/migration/fn.get_storage_value.html.

runtimes/spiritnet/src/migrations/remove_sudo.rs
```diff
@@ -0,0 +1,64 @@
...
+		if let Some(true) = res {
+			Ok(())
+		} else {
+			Err("Sudo key not present")
+		}
+	}
+
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		// Magic bytes are the sudo pallet prefix
+		let _ = frame_support::storage::unhashed::kill_prefix(&hex!["5c0d1176a568c1f92944340dbfed9e9c"], Some(2));
```
Diiaablo95, line 44: Same here? Perhaps https://docs.substrate.io/rustdocs/latest/frame_support/storage/fn.storage_prefix.html.

runtimes/spiritnet/src/migrations/remove_sudo.rs
```diff
@@ -0,0 +1,64 @@
...
+
+		<Runtime as frame_system::Config>::DbWeight::get().writes(2)
```
Diiaablo95, line 46: Why do we have two writes? What other value are we killing other than the sudo key?

Diiaablo95:
Don't have a strong argument in favor of implementing my suggestions, also because the sudo migration will be removed soon, but I think it would still make it easier to read now and in future, if we need a similar approach for other pallets.

Diiaablo95 (commented):
Don't have a strong argument in favour of implementing my suggestions, especially since the sudo migration will be removed soon. Anyway, I think it would make it easier to read it now and in the future, if we want to apply a similar migration to other pallets.

runtimes/spiritnet/src/migrations/remove_sudo.rs
wischli: We are killing 2 subitems with the sudo prefix

```
kill_prefix(&hex!["5c0d1176a568c1f92944340dbfed9e9c"], Some(2));
```

Diiaablo95: Which are? The sudo key itself and...?

Diiaablo95: Oh the pallet version?

runtimes/spiritnet/src/migrations/remove_sudo.rs
weichweich: that would not work since we already get the prefix of a storage entry and not that of a the whole pallet

weichweich: i could do `twox_128(pallet_name)`.

weichweich: next time...

runtimes/spiritnet/src/migrations/remove_sudo.rs
weichweich: Yes the palletVersion

PR #293, opened 2021-11-04
feat: batch did

weichweich:
# Fixes KILTprotocol/ticket/issues/1537

[M](https://kilt-cast.monday.com/boards/1792114080/pulses/1864620642)

## Checklist:


runtimes/spiritnet/src/lib.rs
```diff
@@ -784,15 +784,23 @@ construct_runtime! {
 
 impl did::DeriveDidCallAuthorizationVerificationKeyRelationship for Call {
 	fn derive_verification_key_relationship(&self) -> Option<did::DidVerificationKeyRelationship> {
+		fn single_key_relationship(calls: &[Call]) -> Option<did::DidVerificationKeyRelationship> {
```
Diiaablo95, line 787: This compiles for me. I am quite sure the `as_ref()` is not needed, while I am not super sure why you went with a `flat_map()` instead of a simple `map()`? When we iterate, we are not iterating over options, so we don't need to flatten the map, as we need in the beginning because of the return type of `get()`.

```suggest
		fn single_key_relationship(calls: &[Call]) -> Option<did::DidVerificationKeyRelationship> {
			let init = calls.get(0).map(Call::derive_verification_key_relationship).flatten()?;
			calls
				.iter()
				.map(Call::derive_verification_key_relationship)
				.try_fold(init, |acc, next| if next == acc { Some(acc) } else { None })
		}
```

runtimes/spiritnet/src/tests.rs
```diff
@@ -27,3 +29,87 @@ fn call_size() {
...
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1.clone(), c2.clone(), c4.clone()],
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1, c2, cb, f3, c4],
+	});
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
Diiaablo95, line 100: S: This returns `Some(AssertionMethod)` if `f3` is removed from the batch, so perhaps I would also add a (positive) test for nested calls where all calls require the same key type (also for peregrine runtime)?

runtimes/spiritnet/src/tests.rs
```diff
@@ -27,3 +29,87 @@ fn call_size() {
...
+}
+
+#[test]
+fn test_derive_did_verification_relation_single() {
```
Diiaablo95, line 104: C: What about a test case when the vector of calls is empty (also for peregrine runtime)?

runtimes/peregrine/src/lib.rs
```diff
@@ -778,15 +778,23 @@ construct_runtime! {
 
 impl did::DeriveDidCallAuthorizationVerificationKeyRelationship for Call {
```
Diiaablo95, line 779: C: I would move this logic (and perhaps anything else that is in common between spiritnet and peregrine) into `primitives`, so that we avoid duplicating code. What do you think?

runtimes/spiritnet/src/lib.rs
Diiaablo95: M: Didn't we also agree to change this to a `Result<did::DidVerificationKeyRelationship, DispatchError>`?

Diiaablo95 (commented):
This is the killer feature for DIDs 🔥
Didn't we agree to change the result type of the function to a `Result` instead of an `Option`? Plus, I would like to also have the same logic in the standalone (with no tests), so to reduce the differences between the runtimes until we get rid of it altogether.

runtimes/peregrine/src/lib.rs
weichweich: would love that! But we implement that for Call which is an enum defined in the runtime

runtimes/spiritnet/src/lib.rs
weichweich: yeah wasn't sure if it only was a "this would be cool" thing.
The issue is that we would need to define the error in the DID pallet. The DID pallet doesn't know which errors we want to return here. We could try to keep the error generic?

* `NotADidCall` for calls that don't support the did origin
* `InvalidCallArgument` for calls with arguments that are not allowed in the did call

runtimes/spiritnet/src/tests.rs
weichweich: e88b802453862ef1218b44a5333cadf33c972440

runtimes/spiritnet/src/tests.rs
weichweich: e88b802453862ef1218b44a5333cadf33c972440

runtimes/peregrine/src/lib.rs
Diiaablo95: Mhh yeah ok, and also the runtime itself is an enum so not very easy to extend with runtime-specific logic if we wanted to 😕

runtimes/spiritnet/src/lib.rs
weichweich: DispatchErrors have to be defined in a Pallet. So we have the issue again that the pallet logic depends on the runtime. :/

runtimes/peregrine/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+		remark: vec![0, 1, 2, 3, 3],
+	});
+	let c4 = Call::Ctype(ctype::Call::add {
+		ctype: vec![0, 1, 2, 100],
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1, c2, c3, c4],
+	});
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 75:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

runtimes/peregrine/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1.clone(), c2.clone(), c4.clone()],
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1, c2, cb, f3, c4],
+	});
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 100:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

runtimes/peregrine/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+		cb.derive_verification_key_relationship(),
+		Some(did::DidVerificationKeyRelationship::AssertionMethod)
+	);
+}
+
+#[test]
+fn test_derive_did_verification_relation_empty() {
+	let cb = Call::Utility(pallet_utility::Call::batch { calls: vec![] });
+
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 146:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

runtimes/spiritnet/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+		remark: vec![0, 1, 2, 3, 3],
+	});
+	let c4 = Call::Ctype(ctype::Call::add {
+		ctype: vec![0, 1, 2, 100],
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1, c2, c3, c4],
+	});
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 75:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

runtimes/spiritnet/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1.clone(), c2.clone(), c4.clone()],
+	});
+
+	let cb = Call::Utility(pallet_utility::Call::batch {
+		calls: vec![c1, c2, cb, f3, c4],
+	});
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 100:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

runtimes/spiritnet/src/tests.rs
```diff
@@ -27,3 +29,119 @@ fn call_size() {
...
+		cb.derive_verification_key_relationship(),
+		Some(did::DidVerificationKeyRelationship::AssertionMethod)
+	);
+}
+
+#[test]
+fn test_derive_did_verification_relation_empty() {
+	let cb = Call::Utility(pallet_utility::Call::batch { calls: vec![] });
+
+	assert_eq!(cb.derive_verification_key_relationship(), None);
```
wischli, line 146:
```suggestion
	assert!(cb.derive_verification_key_relationship().is_none());
```

wischli (commented):
One minor thing and I would love some documentation as the TODO was removed which gave at least some background information.

pallets/did/src/did_details.rs
```diff
@@ -576,6 +576,19 @@ impl<T: Config> sp_std::fmt::Debug for DidCreationDetails<T> {
 	}
 }
 
+/// Errors that might occure while deriving the authorization verification key
```
wischli, line 579:
```suggestion
/// Errors that might occur while deriving the authorization verification key
```

pallets/did/src/did_details.rs
```diff
@@ -576,6 +576,19 @@ impl<T: Config> sp_std::fmt::Debug for DidCreationDetails<T> {
...
+/// relationship.
+#[derive(Clone, Debug, Decode, Encode, PartialEq)]
+pub enum RelationshipDeriveError {
+	/// The call is not callable by a did origin.
+	NotCallableByDid,
+
+	/// The parameter of the call where invalid.
```
wischli, line 586: Either singular or plural
```suggestion
	/// The parameter of the call is invalid.
```

pallets/did/src/lib.rs
```diff
@@ -355,6 +355,9 @@ pub mod pallet {
 		InvalidNonce,
 		/// The called extrinsic does not support DID authorisation.
 		UnsupportedDidAuthorizationCall,
+		/// The called extrinsic had parameter that conflicted with each other
```
wischli, line 358:
```suggestion
		/// The called extrinsic had parameters that conflicted with each other
```

wischli (approved):
LGTM

pallets/did/src/lib.rs
```diff
@@ -355,6 +355,9 @@ pub mod pallet {
 		InvalidNonce,
 		/// The called extrinsic does not support DID authorisation.
 		UnsupportedDidAuthorizationCall,
+		/// The call had parameters that conflicted with each other.
```
Diiaablo95, line 358:
```suggestion
		/// The call had parameters that conflicted with each other
```

Diiaablo95 (commented):
Love it!

pallets/did/src/lib.rs
weichweich: Oh it goes on in the next line. didn't see that. 😂

PR #295, opened 2021-11-05
feat: add deposit reclaim extrinsic for delegations

Diiaablo95:
## fixes KILTProtocol/ticket#1628

Adds the `reclaim_deposit` extrinsic to the delegation pallet, callable only by a signed origin which must be the deposit owner. Added tests and benchmarks, which will have to be re-run after review.

An SDK companion PR will follow.

[M](https://kilt-cast.monday.com/boards/1792114080/pulses/1864631047)

## Checklist:

- [ ] Get review
- [ ] Address review comments
- [ ] Run benchmarks
- [ ] Merge


pallets/delegation/Cargo.toml
```diff
@@ -14,10 +14,12 @@ substrate-wasm-builder-runner = {version = "3.0.0"}
 
 [dev-dependencies]
 ctype = {features = ["mock"], path = "../ctype", version = "1.1.1"}
-env_logger = {version = "0.8.4"}
 kilt-primitives = {default-features = false, path = "../../primitives"}
-pallet-balances = {branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"}
+
```
weichweich, line 18: maybe we could also split the `#External dependencies` into two sections `# External dependencies` and `# Substrate dependencies`?

pallets/delegation/src/lib.rs
```diff
@@ -625,6 +628,62 @@ pub mod pallet {
 			// Add worst case reads from `is_delegating`
 			Ok(Some(<T as Config>::WeightInfo::remove_delegation(removal_checks)).into())
 		}
+
+		/// Reclaim back the deposit for a delegation node (potentially a root
```
weichweich, line 632: I think this is superfluous.

You always reclaim something "back". There is no forward reclaiming. This would probably be claiming then.
Further down you write `Returns the delegation deposit back`. Same there. Returning is always back never forward.

```suggestion
		/// Reclaim the deposit for a delegation node (potentially a root
```

weichweich (approved):
LGTM
It's all there, tests, doc, functionality. Love it!
But I'm triggered by `reclaiming back` and `returning back`! 🍠

pallets/delegation/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/295/commits/cb08cd6d8c074cc1ce722b8b4c6ec0d62502e034

pallets/delegation/Cargo.toml
Diiaablo95: I though it was too much and outscoped, but I see 🧐 -> https://github.com/KILTprotocol/mashnet-node/pull/295/commits/cf2059d968d39f3ac1d129de186d6ad6c180c96d

weichweich: yes I was actually about to say that. But I remembered that positive attitude

pallets/delegation/src/benchmarking.rs
```diff
@@ -329,7 +329,25 @@ benchmarks! {
...
-	}: remove_delegation(RawOrigin::Signed(root_owner.clone()), hierarchy_id, r)
+	}: _(RawOrigin::Signed(root_owner.clone()), hierarchy_id, r)
+	verify {
+		assert!(!DelegationNodes::<T>::contains_key(hierarchy_id));
+		assert!(!DelegationNodes::<T>::contains_key(child_id));
+		assert!(!DelegationNodes::<T>::contains_key(leaf_id));
+		assert!(<T as Config>::Currency::reserved_balance(&root_acc.into()).is_zero());
+	}
+
+	// worst case is achieved by removing the root node, since `is_delegating` is not called in remove extrinsic,
```
wischli, line 340: -1P 😛

```suggestion
	// worst case is achieved by removing the root node, since `is_delegating` is not called in remove extrinsic
```

pallets/delegation/src/lib.rs
```diff
@@ -625,6 +628,62 @@ pub mod pallet {
...
+			// max_removals to account for the node itself), releasing the associated
+			// deposit
+			let (removal_checks, _) = Self::remove(&delegation_id, max_removals.saturating_add(1))?;
+
+			// Delete the delegation hierarchy details, if the provided ID was for a root
+			// node. No event generated as we don't have information about the owner DID
+			// here.
+			DelegationHierarchies::<T>::remove(&delegation_id);
+
+			// Add worst case reads from `is_delegating`
```
wischli, line 684: `is_delegating` is not used here. Just noticed I made the same c&p error when I added deposits 🥲. I propose to remove the comment, same for `fn remove_delegation`. 

Otherwise, something like "Post weight correction to number of actually removed delegations instead of max_removals".

```suggestion
```

wischli (approved):
LGTM! Two super minor things you could resolve.

I am wondering whether we should keep `fn remove_delegation` at all since it is pretty redundant with adding the reclaiming.

Diiaablo95:
@wischli I think we should keep both as each of them gives "power" to a different user type, and if we remove the `remove` function, the owner of a delegation (or attestation, as it has the same feature) would not be able to remove their stuff.

pallets/delegation/src/benchmarking.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/295/commits/3c06d1a213da8081b44891f1baf3c8b4198d1bc8

pallets/delegation/src/lib.rs
Diiaablo95: https://github.com/KILTprotocol/mashnet-node/pull/295/commits/4ad692568b09e6b7a73ac745da48c2f6c3b5a95d

PR #300, opened 2021-11-17
feat: add lazy total stake calculation

wischli:
## fixes KILTProtocol/ticket#1701
* Improves updating of `TotalCollatorStake` storage by removing the iteration over all selected candidates (except for `init_leave_candidate` and `force_remove_candidate`
* Updates documentation

## Checklist:

### DIFF
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -420,7 +420,10 @@ pub mod pallet {
 		/// The collator delegate or the delegator is trying to un-stake more
 		/// funds that are currently staked.
 		Underflow,
-		/// The number of selected candidates or of blocks per staking round is
+		/// The number of selected candidatesper staking round is
+		/// above the maximum value allowed.
+		CannotSetAboveMax,
+		/// The number of selected candidates per staking round is
 		/// below the minimum value allowed.
 		CannotSetBelowMin,
 		/// An invalid inflation configuration is trying to be set.
@@ -447,13 +450,8 @@ pub mod pallet {
 		/// \[account\]
 		LeftTopCandidates(T::AccountId),
 		/// A new account has joined the set of collator candidates.
-		/// \[account, amount staked by the new candidate, new total stake of
-		/// collator candidates\]
-		JoinedCollatorCandidates(T::AccountId, BalanceOf<T>, BalanceOf<T>),
-		/// A collator candidate has been selected for the next validation
-		/// round. \[collator's account, collator's total stake, collator's
-		/// delegators' total stake\]
-		CollatorChosen(T::AccountId, BalanceOf<T>, BalanceOf<T>),
+		/// \[account, amount staked by the new candidate\]
+		JoinedCollatorCandidates(T::AccountId, BalanceOf<T>),
 		/// A collator candidate has increased the amount of funds at stake.
 		/// \[collator's account, previous stake, new stake\]
 		CollatorStakedMore(T::AccountId, BalanceOf<T>, BalanceOf<T>),
@@ -640,13 +638,14 @@ pub mod pallet {
 
 	/// The collator candidates with the highest amount of stake.
 	///
-	/// Each time the stake of a collator is increased, it is checked whether is
-	/// pushes another candidate out of the list. When the stake is reduced, it
-	/// is not checked of another candidate has more stake, since this would
-	/// require the iterating over the [CandidatePool].
+	/// Each time the stake of a collator is increased, it is checked whether
+	/// this pushes another candidate out of the list. When the stake is
+	/// reduced however, it is not checked if another candidate has more stake,
+	/// since this would require iterating over the entire [CandidatePool].
 	///
 	/// There must always be more candidates than [MaxSelectedCandidates] so
-	/// that a collator can drop out of the collator set by reducing his stake.
+	/// that a collator can drop out of the collator set by reducing their
+	/// stake.
 	#[pallet::storage]
 	#[pallet::getter(fn top_candidates)]
 	pub(crate) type TopCandidates<T: Config> =
@@ -718,7 +717,7 @@ pub mod pallet {
 			assert!(self.inflation_config.is_valid(), "Invalid inflation configuration");
 
 			<InflationConfig<T>>::put(self.inflation_config.clone());
-			<MaxCollatorCandidateStake<T>>::put(self.max_candidate_stake);
+			MaxCollatorCandidateStake::<T>::put(self.max_candidate_stake);
 
 			// Setup delegate & collators
 			for &(ref actor, ref opt_val, balance) in &self.stakers {
@@ -740,9 +739,8 @@ pub mod pallet {
 				}
 			}
 			// Set total selected candidates to minimum config
-			<MaxSelectedCandidates<T>>::put(T::MinCollators::get());
+			MaxSelectedCandidates::<T>::put(T::MinCollators::get());
 
-			// Choose top MaxSelectedCandidates collator candidates
 			<Pallet<T>>::update_total_stake();
 
 			// Start Round 0 at Block 0
@@ -839,25 +837,59 @@ pub mod pallet {
 		/// - The transaction's complexity is mainly dependent on updating the
 		///   `SelectedCandidates` storage in `select_top_candidates` which in
 		///   return depends on the number of `MaxSelectedCandidates` (N).
-		/// - For each N, we read `CollatorState` from the storage.
+		/// - For each N, we read `CandidatePool` from the storage.
 		/// ---------
 		/// Weight: O(N) where N is `MaxSelectedCandidates` bounded by
 		/// `MaxTopCandidates`
-		/// - Reads: MaxSelectedCandidates, TopCandidates, N * CollatorState
+		/// - Reads: MaxSelectedCandidates, TopCandidates, N * CandidatePool
 		/// - Writes: MaxSelectedCandidates
 		/// # </weight>
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_max_selected_candidates(*new, (*new).saturating_mul(T::MaxDelegatorsPerCollator::get())))]
 		pub fn set_max_selected_candidates(origin: OriginFor<T>, new: u32) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 			ensure!(new >= T::MinCollators::get(), Error::<T>::CannotSetBelowMin);
-			let old = <MaxSelectedCandidates<T>>::get();
+			ensure!(new <= T::MaxTopCandidates::get(), Error::<T>::CannotSetAboveMax);
+			let old = MaxSelectedCandidates::<T>::get();
 
 			// *** No Fail beyond this point ***
 
-			<MaxSelectedCandidates<T>>::put(new);
+			MaxSelectedCandidates::<T>::put(new);
 
-			// update candidates for next round
-			let (num_collators, num_delegators, _, _) = Self::update_total_stake();
+			// Update total amount at stake for new top collators and their delegators
+			let (num_collators, num_delegators) = match new {
+				x if x > old => {
+					let candidates = TopCandidates::<T>::get();
+					let num_new_collators: usize = candidates
+						.len()
+						.min(new.saturated_into())
+						.saturating_sub(old.saturated_into());
+					let (col_stake, del_stake, num_delegators) =
+						// FIXME: Check for more simple solution to take slice [old + 1, ..., num_new]
+						candidates.into_iter().take(new.saturated_into()).rev().take(num_new_collators).fold(
+							(BalanceOf::<T>::zero(), BalanceOf::<T>::zero(), 0u32),
+							|(col_stake, del_stake, num_delegators), candidate| {
+								if let Some(state) = CandidatePool::<T>::get(&candidate.owner) {
+									(
+										col_stake.saturating_add(state.stake),
+										del_stake.saturating_add(state.total - state.stake),
+										num_delegators.saturating_add(state.delegators.len().saturated_into()),
+									)
+								} else {
+									(col_stake, del_stake, num_delegators)
+								}
+							},
+						);
+
+					TotalCollatorStake::<T>::mutate(|total| {
+						total.collators = total.collators.saturating_add(col_stake);
+						total.delegators = total.delegators.saturating_add(del_stake);
+					});
+
+					(num_new_collators.saturated_into::<u32>(), num_delegators)
+				}
+				x if x < old => Self::update_total_stake(),
+				_ => (0u32, 0u32),
+			};
 
 			Self::deposit_event(Event::MaxSelectedCandidatesSet(old, new));
 
@@ -950,16 +982,16 @@ pub mod pallet {
 		/// - The transaction's complexity is mainly dependent on updating the
 		///   `SelectedCandidates` storage in `select_top_candidates` which in
 		///   return depends on the number of `MaxSelectedCandidates` (N).
-		/// - For each N, we read `CollatorState` from the storage.
+		/// - For each N, we read `CandidatePool` from the storage.
 		/// ---------
 		/// Weight: O(N + D) where N is `MaxSelectedCandidates` bounded by
 		/// `MaxTopCandidates` and D is the number of delegators of the
 		/// collator candidate bounded by `MaxDelegatorsPerCollator`
-		/// - Reads: MaxCollatorCandidateStake, 2 * N * CollatorState,
+		/// - Reads: MaxCollatorCandidateStake, 2 * N * CandidatePool,
 		///   TopCandidates, BlockNumber, D * DelegatorState, D * Unstaking
-		/// - Writes: MaxCollatorCandidateStake, N * CollatorState,
-		///   SelectedCandidates, D * DelegatorState, (D + 1) * Unstaking
-		/// - Kills: CollatorState, DelegatorState for all delegators which only
+		/// - Writes: MaxCollatorCandidateStake, N * CandidatePool, D *
+		///   DelegatorState, (D + 1) * Unstaking
+		/// - Kills: CandidatePool, DelegatorState for all delegators which only
 		///   delegated to the candidate
 		/// # </weight>
 		#[pallet::weight(<T as Config>::WeightInfo::force_remove_candidate(T::MaxTopCandidates::get(), T::MaxTopCandidates::get().saturating_mul(T::MaxDelegatorsPerCollator::get())))]
@@ -969,10 +1001,10 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			ensure_root(origin)?;
 			let collator = T::Lookup::lookup(collator)?;
-			let state = <CandidatePool<T>>::get(&collator).ok_or(Error::<T>::CandidateNotFound)?;
+			let state = CandidatePool::<T>::get(&collator).ok_or(Error::<T>::CandidateNotFound)?;
 			let total_amount = state.total;
 
-			let mut candidates = <TopCandidates<T>>::get();
+			let mut candidates = TopCandidates::<T>::get();
 			ensure!(
 				candidates.len().saturated_into::<u32>() > T::MinRequiredCollators::get(),
 				Error::<T>::TooFewCollatorCandidates
@@ -982,18 +1014,20 @@ pub mod pallet {
 
 			Self::remove_candidate(&collator, &state)?;
 
-			if candidates
+			let (num_collators, num_delegators) = if candidates
 				.remove(&Stake {
 					owner: collator.clone(),
 					amount: state.total,
 				})
 				.is_some()
 			{
+				// update top candidates
 				TopCandidates::<T>::put(candidates);
-			}
-
-			// update candidates for next round
-			let (num_collators, num_delegators, _, _) = Self::update_total_stake();
+				// update total amount at stake from scratch
+				Self::update_total_stake()
+			} else {
+				(0u32, 0u32)
+			};
 
 			Self::deposit_event(Event::CollatorRemoved(collator, total_amount));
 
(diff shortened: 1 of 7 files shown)


pallets/parachain-staking/src/lib.rs
```diff
@@ -420,7 +420,10 @@ pub mod pallet {
 		/// The collator delegate or the delegator is trying to un-stake more
 		/// funds that are currently staked.
 		Underflow,
-		/// The number of selected candidates or of blocks per staking round is
+		/// The number of selected candidatesper staking round is
```
Diiaablo95, line 423:
```suggestion
		/// The number of selected candidates per staking round is
```

pallets/parachain-staking/src/lib.rs
```diff
@@ -1091,42 +1124,41 @@ pub mod pallet {
...
 		///
 		/// NOTE: Upon starting a new session_i in `new_session`, the current
 		/// top candidates are selected to be block authors for session_i+1. Any
 		/// changes to the top candidates afterwards do not effect the set of
 		/// authors for session_i+1.
 		/// Thus, we have to make sure none of these collators can
-		/// leave before session_i+1 ends by keeping them in the ExitQueue for
-		/// at least 2 sessions (= 2 rounds), e.g., the current (i) and the next
-		/// one (i+1).
+		/// leave before session_i+1 ends by keeping them in the delaying their
```
Diiaablo95, line 1136:
```suggestion
		/// leave before session_i+1 ends by keeping them in the ExitQueue delaying their
```

wischli: I actually meant to say

> by ~keeping them in the~delaying their exit...

pallets/parachain-staking/src/lib.rs
```diff
@@ -839,25 +837,63 @@ pub mod pallet {
...
+								} else {
+									(col_stake, del_stake, num_delegators)
+								}
+							},
+						);
+
+					TotalCollatorStake::<T>::mutate(|total| {
+						total.collators = total.collators.saturating_add(col_stake);
+						total.delegators = total.delegators.saturating_add(del_stake);
+					});
```
weichweich, line 890: we don't read from the old total. So i would just use `set`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -839,25 +837,63 @@ pub mod pallet {
...
 			// *** No Fail beyond this point ***
 
-			<MaxSelectedCandidates<T>>::put(new);
-
-			// update candidates for next round
-			let (num_collators, num_delegators, _, _) = Self::update_total_stake();
+			MaxSelectedCandidates::<T>::put(new);
+
+			// Update total amount at stake for new top collators and their delegators
+			let (num_collators, num_delegators) = match new {
```
weichweich, line 863: This whole match statement looks quite complicated to me. 🤔 

I think this could maybe be simpler?

```rust
let start = old.min(new);
let end = old.max(new);

let candidates = TopCandidates::<T>::get();

let diff_delegation = 0;
let diff_collation = 0;
for candidate in candidates.skip(start).take(end - start) {
  if let Some(state) = CandidatePool::<T>::get(&candidate.owner) {
    diff_collation.saturating_add(state.stake);
    diff_delegation.saturating_add(state.total - state.stake);
  }
}

TotalCollatorStake::<T>::mutate(...)
```

weichweich (commented):
didn't finish yet.

pallets/parachain-staking/src/lib.rs
wischli: We do read from the old total by incrementing the existing values?!

weichweich: completely overlooked that. 😂

pallets/parachain-staking/src/lib.rs
wischli: Thanks. I am very unhappy with the readability of the entire block. Your proposal looks much cleaner.

weichweich (commented):
I think update_top_candidates is going to be more complicated?

wischli:
Some parts of the weight complexity doc comments in `parachain-staking/src.lib.rs` should be reverted back to their original version but AFAIK we want to get rid of them anyways, right @weichweich ?

weichweich:
> Some parts of the weight complexity doc comments in `parachain-staking/src.lib.rs` should be reverted back to their original version but AFAIK we want to get rid of them anyways, right @weichweich ?

Yes we have the benchmarks and that should be fine

weichweich (changes requested):
I think the `try_insert_replace` method is very confusing and doesn't work as you intended, because it doesn't replace a value if it is equal.

In general, i would be interested in the latest benchmarks ;)
In general2, i would love to see the complexity of this pallet reduced. It's unbelievable complex. I also have no idea how.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1947,84 +2025,138 @@ pub mod pallet {
...
+					TotalCollatorStake::<T>::mutate(|total| {
+						total.collators = total.collators.saturating_sub(old_self).saturating_add(new_self);
+						total.delegators = total
+							.delegators
+							.saturating_sub(old_delegators)
+							.saturating_add(new_delegators);
+					});
+				}
+			// case 2: candidate newly joins TopCandidates and either replaces
+			// another candidate in the list or updates their own stake
```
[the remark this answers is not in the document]
wischli: > isn't this method always called when stake is updated?

This method yes, but `case 2` only applies if the candidate was not part of TopCandidates with the old stake (`case 1`) or did not stake at all previously.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1947,84 +2025,138 @@ pub mod pallet {
...
+				// retrieve position of candidate in TopCandidates
+				if let Ok(i) = top_candidates.linear_search(&Stake {
+					owner: candidate.clone(),
+					amount: new_total,
+				}) {
+					// update TotalStake only if candidate becomes collator
+					let max_selected_candidates = MaxSelectedCandidates::<T>::get().saturated_into::<usize>();
+					if i < max_selected_candidates {
+						// candidate either pushed out the least staked collator or updated their own
+						// stake
```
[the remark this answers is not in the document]
wischli: Yes, but the value is of type `Stake` and not `Balance` or `AccountId`. Thus, by increasing the staked amount, the candidate can replace their own entry in the TopCandidates list. Otherwise, the candidate can replace any other candidate with less stake.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1947,84 +2025,138 @@ pub mod pallet {
...
+								.saturating_sub(drop_delegators)
+								.saturating_add(new_delegators);
+						});
+
+						// only trigger event if candidate and drop_out differ
+						if let Some(Stake { owner: drop_id, .. }) = drop_out {
+							if drop_id != candidate {
+								Self::deposit_event(Event::LeftTopCandidates(drop_id));
+							}
+						}
```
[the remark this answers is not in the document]
wischli: You are mistaking membership with matching `AccountId`s when in fact, the `Stake::<T> { amount: BalanceOf<T>, owner: AccountIdOf<T> }` would have to match, see [above](https://github.com/KILTprotocol/mashnet-node/pull/300#discussion_r764870721).

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
weichweich: I think i still didn't understand. 😂 

case 1: he was in the top candidates and the delegator stake changed, but the self stake stayed the same
case 2: He was not part of top candidates || he was part of the list and changed their own stake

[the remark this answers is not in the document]
wischli: Case 1: Candidate was in TopCandidates with old stake independent of triggered.
Case 2: Not Case 1 && Candidate becomes member of TopCandidates with new stake.

Both cases are independent of who/what triggered the total stake change.

However I just realized that case 1 is incomplete and incorrect: Currently I update the TotalCollatorStake if the candidate was a collator with old stake and don' t check their position with new stake. This only works if we can assume the candidate still is a collator 🙈

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: > You are mistaking membership with matching `AccountId`s when in fact, the `Stake::<T> { amount: BalanceOf<T>, owner: AccountIdOf<T> }` would have to match, see [above](https://github.com/KILTprotocol/mashnet-node/pull/300#discussion_r764870721).

I revert this statement. You are right, two Stakes are considered equal in an OrderedSet if the AccountId matches.

pallets/parachain-staking/src/lib.rs
```diff
@@ -825,30 +823,61 @@ pub mod pallet {
...
+		/// Weight: O(N + D) where N is `MaxSelectedCandidates` bounded by
+		/// `MaxTopCandidates` and D is the number of delegators of a
+		/// candidate bounded by `MaxDelegatorsPerCollator`.
+		/// - Reads: MaxSelectedCandidates, TopCandidates, N * CandidatePool
 		/// - Writes: MaxSelectedCandidates
 		/// # </weight>
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_max_selected_candidates(*new, (*new).saturating_mul(T::MaxDelegatorsPerCollator::get())))]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_max_selected_candidates(
+			*new,
+			T::MaxDelegatorsPerCollator::get()
```
Diiaablo95, line 836: Q: I can't see where the second argument for weight computation is useful. We compute it down here, but we always take the max and we never iterate over those, only compute the `len()` of the list, so I don't think it's relevant neither for the benchmark nor to keep track of it in any way?

pallets/parachain-staking/src/lib.rs
```diff
@@ -968,18 +1000,20 @@ pub mod pallet {
 
 			Self::remove_candidate(&collator, &state)?;
 
-			if candidates
+			let (num_collators, num_delegators) = if candidates
```
Diiaablo95, line 1003: Same thing as above: why do we need to know `num_delegators`?

pallets/parachain-staking/src/lib.rs
```diff
@@ -1802,35 +1858,47 @@ pub mod pallet {
 			ensure!(!more.is_zero(), Error::<T>::ValStakeZero);
 
 			let candidate = T::Lookup::lookup(candidate)?;
-			let mut delegations = <DelegatorState<T>>::get(&delegator).ok_or(Error::<T>::DelegatorNotFound)?;
-			let mut collator = <CandidatePool<T>>::get(&candidate).ok_or(Error::<T>::CandidateNotFound)?;
+			let mut delegations = DelegatorState::<T>::get(&delegator).ok_or(Error::<T>::DelegatorNotFound)?;
+			let mut collator = CandidatePool::<T>::get(&candidate).ok_or(Error::<T>::CandidateNotFound)?;
 			ensure!(!collator.is_leaving(), Error::<T>::CannotDelegateIfLeaving);
-			let delegator_total = delegations
+			let delegatodrop_total = delegations
```
Diiaablo95, line 1864: Is `delegatodrop_total` the actual name here?

Diiaablo95:
To answer @weichweich point2 about complexity, I think it would help if we can have all the storage reads within the extrinsic, and then have all the auxiliary functions work as a function of the input, without performing additional logic. Splitting responsibilities in several smaller pure functions, and accessing all the storage in the extrinsic. Those functions are then, in my opinion, more easily testable. Nevertheless, this is food for thought for the future.

pallets/parachain-staking/src/lib.rs
wischli: I hoped to get rid of the required number of delegator arguments for most benchmarks but unfortunately they are still necessary. However, before we took the sum of the delegators of all the collators we iterate in an extrinsic over as the value which was incorrect. Ideally, we would take something the average number of delegators each collator, which is relevant for the call, has. For worst case handling, we take the maximum number of delegators of a collator here.

I don't see why this should not be relevant? And what else should we do other than taking the `len()` of the list? This directly tells us the number of delegators for a specific collator which is all that we are interested in for this parameter.

pallets/parachain-staking/src/lib.rs
wischli: Same argument as above.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1049,20 +1082,23 @@ pub mod pallet {
...
-			// update candidates for next round
-			let (num_collators, num_delegators, total_collators, _) = Self::update_total_stake();
-
-			Self::deposit_event(Event::JoinedCollatorCandidates(sender, stake, total_collators));
+			Self::deposit_event(Event::JoinedCollatorCandidates(sender, stake));
 			Ok(Some(<T as pallet::Config>::WeightInfo::join_candidates(
-				num_collators,
-				num_delegators,
+				n,
+				T::MaxDelegatorsPerCollator::get(),
```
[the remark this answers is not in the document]
wischli: The benchmark demands the delegators parameter.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1124,20 +1160,22 @@ pub mod pallet {
 
 			// *** No Fail beyond this point ***
 
-			if candidates
+			let (num_collators, num_delegators) = if candidates
```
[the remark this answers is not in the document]
wischli: See above.

wischli:
> I think it would help if we can have all the storage reads within the extrinsic, and then have all the auxiliary functions work as a function of the input, without performing additional logic. Splitting responsibilities in several smaller pure functions, and accessing all the storage in the extrinsic

I am honestly unsure whether this reduce the complexity of this pallet because this will add a LOT of duplicate code. But it's definitely something worth trying out in the future.

pallets/parachain-staking/src/lib.rs
wischli: No :( --> fee213a

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: 1cf58e8

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: 1cf58e8

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: 1cf58e8

pallets/parachain-staking/src/lib.rs
wischli: The other option would be to run the benchmarks always with the maximum number of delegators.

Diiaablo95: What I am saying is that, taking the `set_max_selected_candidates(origin, new)` as example, the only places where we use the information about the number of delegators are the following:

- when iterating over `TopCandidates`, where we take, for each candidate, its `state.delegators.len().saturated_into::<u32>()`;
- in the `reduce`, we then take the max value for that field among all the collators we iterate over

Once that is done, we update `TotalCollatorStake`, where we don't use the information about the number of delegators.
What I am saying is that we are not iterating over the number of delegators we get from the iteration, in this specific extrinsic, so unless the `.len()` function is dependent on the size of the delegators set, I do not see how the max number of delegators we find for a collator affects the runtime execution in any way.
Am I missing something else?

wischli: As discussed, I suppose the number of delegators factor in because when reading the state of a candidate from `CandidatePool`, the delegations still have to be decoded.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+					}
+					(true, false) => {
+						// candidate was displaced out of collators the current last collator with index
+						// max_selected_candidates - 1
+						let i = max_selected_candidates.min(top_candidates.len()).saturating_sub(1);
+
+						// get displacer
+						let (add_collators, add_delegators) = Self::get_top_candidate_stake_at(&top_candidates, i)
+							// shouldn't be possible to fail, but we handle it gracefully
+							.unwrap_or((new_self, new_delegators));
```
Diiaablo95, line 2123: I think we can fail if `top_candidates < max_selected_candidates`, right? Kind of an out-of-bound error. So I think that returning `(new_self, new_delegators)` is the right default here.

Diiaablo95 (approved):
I like the new approach better, as it is a bit easier to follow now (praised be the Rust pattern matching 🙏). I would say it looks good to me, also considering the additional test cases added 👍. I do have few comments about clarity tho to make it a bit easier to go back to this code in 6 months time.

pallets/parachain-staking/src/lib.rs
wischli: Even that should be covered because our replacement index is min(top_candidates.len(), max_selected_candidates - 1). It can only fail if the candidate at the replacement index does not have an entry in `CandidatePool` which should be impossible.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+			let mut num_of_delegators = 0u32;
+			let mut collator_stake = BalanceOf::<T>::zero();
+			let mut delegator_stake = BalanceOf::<T>::zero();
+
+			let collators = Self::selected_candidates();
+
+			// Snapshot exposure for round for weighting reward distribution
+			for account in collators.iter() {
+				let state =
+					<CandidatePool<T>>::get(&account).expect("all members of TopCandidates must be candidates q.e.d");
```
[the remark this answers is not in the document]
wischli: 46f8ca7

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+					.map(|i| i < max_selected_candidates)
+					.unwrap_or(false);
+
+				// only update TopCollatorStake storageif candidate was or will be a collator
+				match (was_collating, is_collating) {
+					(true, true) => {
+						Self::update_total_stake_by(new_self, new_delegators, old_self, old_delegators);
+					}
+					(true, false) => {
+						// candidate was displaced out of collators the current last collator with index
```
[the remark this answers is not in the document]
wischli: 46f8ca7

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+						// max_selected_candidates - 1
```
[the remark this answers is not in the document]
wischli: 46f8ca7

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+
+						// get displacer
+						let (add_collators, add_delegators) = Self::get_top_candidate_stake_at(&top_candidates, i)
+							// shouldn't be possible to fail, but we handle it gracefully
+							.unwrap_or((new_self, new_delegators));
+						Self::update_total_stake_by(add_collators, add_delegators, old_self, old_delegators);
+					}
+					(false, true) => {
+						// candidate was pushed out of collators, e.g. index is max_selected_candidates
+						let i = max_selected_candidates.min(top_candidates.len().saturating_sub(1));
```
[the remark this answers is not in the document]
wischli: 46f8ca7

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
+
+				// only update TopCollatorStake storageif candidate was or will be a collator
+				match (was_collating, is_collating) {
+					(true, true) => {
+						Self::update_total_stake_by(new_self, new_delegators, old_self, old_delegators);
+					}
+					(true, false) => {
+						// candidate was displaced out of collators the current last collator with index
+						// max_selected_candidates - 1
+						let i = max_selected_candidates.min(top_candidates.len()).saturating_sub(1);
```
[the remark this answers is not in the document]
wischli: 46f8ca7

pallets/parachain-staking/src/lib.rs
```diff
@@ -2061,85 +2068,162 @@ pub mod pallet {
...
-			let new_total = new_self.saturating_add(new_delegators);
+			let new_stake = Stake {
+				owner: candidate.clone(),
+				amount: new_self.saturating_add(new_delegators),
+			};
 
-			// case 1: candidate has been a member of TopCandidates
-			if let Ok(i) = top_candidates.linear_search(&old_stake) {
+			// update TopCandidates set
+			let working_title = if let Ok(i) = top_candidates.linear_search(&old_stake) {
```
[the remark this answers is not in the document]
wischli: Good catch! 46f8ca7

weichweich (approved):
LGTM

pallets/parachain-staking/src/lib.rs
```diff
@@ -2406,16 +2539,17 @@ pub mod pallet {
...
-		/// candidate bounded by `MaxDelegatorsPerCollator`
+		/// Weight: O(D + U) where D is the number of delegators of the collator
+		/// candidate bounded by `MaxDelegatorsPerCollator` and U is the
+		/// number of locked unstaking requests bounded by `MaxUnstakeRequests`.
 		/// - Reads: BlockNumber, D * DelegatorState, D * Unstaking
 		/// - Writes: D * DelegatorState, (D + 1) * Unstaking
-		/// - Kills: CollatorState, DelegatorState for all delegators which only
+		/// - Kills: CandidatePool, DelegatorState for all delegators which only
 		///   delegated to the candidate
 		/// # </weight>
```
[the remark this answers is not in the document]
wischli: Yes. The changes here were made before this decision.

PR #306, opened 2021-12-13
fix: no default accounts

weichweich:
## fixes KILTProtocol/ticket#1715

This should not change anything. But just to make sure that these potentially problematic lines are not copied and used somewhere else, I would change it.

## Checklist:


primitives/src/constants.rs
```diff
@@ -219,12 +219,10 @@ pub mod did {
...
 pub mod treasury {
 	use crate::{Balance, BlockNumber};
-	use frame_support::PalletId;
 
 	use super::{BLOCKS_PER_YEAR, KILT};
 
 	pub const INITIAL_PERIOD_LENGTH: BlockNumber = BLOCKS_PER_YEAR.saturating_mul(5);
 	const YEARLY_REWARD: Balance = 2_000_000u128 * KILT;
 	pub const INITIAL_PERIOD_REWARD_PER_BLOCK: Balance = YEARLY_REWARD / (BLOCKS_PER_YEAR as Balance);
-	pub const TREASURY_PALLET_ID: PalletId = PalletId(*b"kilt/tsy");
```
wischli, line 229: Q: Why did you move this out of constants?

pallets/kilt-launch/src/mock.rs
```diff
@@ -111,12 +111,17 @@ parameter_types! {
 	pub const AutoUnlockBound: u32 = 70;
 }
 
+parameter_types! {
+	pub const LaunchPalletId: PalletId = PalletId(*b"kilt/thx");
```
wischli, line 115: M: Should be
```suggestion
	pub const LaunchPalletId: PalletId = PalletId(*b"kilt/lch");
```

wischli (commented):
I would like to move the pallet ids to the constants and import them from there.

primitives/src/constants.rs
weichweich: The IDs should all be at the same place so that you can spot duplicates quickly.
We could have them in the primitives constants. But IMHO this makes it only more complicated. They are mostly used in a single place. It doesn't matter if they differ between peregrine and spiritnet so I don't see the point in moving them to the primitives.

I'm also not sure if it's a good idea to have a fixed prefix like `kilt/`. effectively there are only three letters left to change?

pallets/kilt-launch/src/mock.rs
weichweich: fbf0251

primitives/src/constants.rs
weichweich: I moved it back to the primitives.

primitives/src/lib.rs
```diff
@@ -178,3 +151,14 @@ pub type FeeSplit<R, B1, B2> = SplitFeesByRatio<R, FeeSplitRatio, B1, B2>;
 /// https://w3f-research.readthedocs.io/en/latest/polkadot/Token%20Economics.html#-2.-slow-adjusting-mechanism
 pub type SlowAdjustingFeeUpdate<R> =
 	TargetedFeeAdjustment<R, TargetBlockFullness, AdjustmentVariable, MinimumMultiplier>;
+
+/// Unique identifiers for pallets.
```
Diiaablo95, line 155: W: I would move the pallet_id module into its own file. I know it does not have a lot of information, but it is easier to find. This lib.rs file feels like a big pot of stuff.

Diiaablo95 (approved):
LGMT! I did not check the crowdloan pallet as I guess we need to merge `master` back into `develop` and `develop` back into this branch.

primitives/src/lib.rs
weichweich: 7857d6d

primitives/src/constants.rs
weichweich: 143923e

wischli (approved):
LGTM!

PR #308, opened 2021-12-15
fix: remove left overs from crowdloan pallet

weichweich:


Diiaablo95 (approved):
LGTM! Just a `try-runtime` test and then it could be pushed.

wischli (approved):
LGTM

PR #309, opened 2022-01-11
chore: substrate upgrade 0.9.13

weichweich:
## fixes KILTProtocol/ticket#1751

* update to 0.9.13
* Have the same order in cargo.toml
* use same log dependency specification everywhere
* update copyright to 2022
* add a [rust toolchain file](https://rust-lang.github.io/rustup/overrides.html) so everyone uses the correct nightly
* update to rust 2021
* Noteworthy changes in 0.9.13:
  * Events have now named fields (Pallets: Collective, Balance, Democracy, )

[Substrate diff](https://github.com/paritytech/substrate/compare/polkadot-v0.9.12...polkadot-v0.9.13)
[Cumulus diff](https://github.com/paritytech/cumulus/compare/polkadot-v0.9.12...polkadot-v0.9.13)

* [x] update pallet-utility benchmarks (new extrinsic `dispatch_as`)
* [x] review `dispatch_as` extrinsic

## Checklist:


.dockerignore
```diff
@@ -2,7 +2,7 @@
 .github
 .gitignore
 .gitlab-ci.yml
-**/target
+**/target*
```
weichweich, line 5: This was unintentional. But maybe it doesn't hurt. I have a separate rust-analyser target dir...

nodes/parachain/src/cli.rs
```diff
@@ -20,12 +20,11 @@ use crate::chain_spec;
...
 use structopt::StructOpt;
 
-pub const DEFAULT_RUNTIME: &str = "peregrine";
-pub const DEFAULT_PARA_ID: &str = "2000";
+pub(crate) const DEFAULT_RUNTIME: &str = "peregrine";
 
 /// Sub-commands supported by the collator.
 #[derive(Debug, StructOpt)]
-pub enum Subcommand {
+pub(crate) enum Subcommand {
```
weichweich, line 27: with `pub(crate)` we get unused warnings. These things are never used outside anyways...

runtimes/peregrine/src/lib.rs
```diff
@@ -689,53 +692,53 @@ construct_runtime! {
 		UncheckedExtrinsic = UncheckedExtrinsic,
 	{
 		// Basic stuff; balances is uncallable initially.
-		System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage} = 1,
+		System: frame_system = 0,
```
weichweich, line 695: Not sure about this change. Maybe we rather revert it. There could be an unintentional change hidden here. E.g. we didn't include `call` for a specific pallet but now it's included.

runtimes/standalone/Cargo.toml
```diff
@@ -75,10 +76,11 @@ runtime-benchmarks = [
   "ctype/runtime-benchmarks",
   "delegation/runtime-benchmarks",
   "did/runtime-benchmarks",
-  "frame-benchmarking",
+  "frame-benchmarking/runtime-benchmarks",
   "frame-support/runtime-benchmarks",
   "frame-system/runtime-benchmarks",
   "frame-system-benchmarking",
+  "hex-literal",
```
wischli, line 83: See above.

runtimes/standalone/src/lib.rs
```diff
@@ -497,41 +498,41 @@ construct_runtime!(
...
-
-		Ctype: ctype::{Pallet, Call, Storage, Event<T>} = 9,
-		Attestation: attestation::{Pallet, Call, Storage, Event<T>} = 10,
-		Delegation: delegation::{Pallet, Call, Storage, Event<T>} = 11,
-		Did: did::{Pallet, Call, Storage, Event<T>, Origin<T>} = 12,
-		DidLookup: pallet_did_lookup::{Pallet, Call, Storage, Event<T>} = 13,
-
-		Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>} = 15,
-		Authorship: pallet_authorship::{Pallet, Call, Storage} = 16,
+		System: frame_system = 0,
```
wischli, line 501: Placeholder. See your suggestion above.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly-2021-11-08"
+targets = [ "wasm32-unknown-unknown" ]
```
wischli, line 3: Very nice! 👍

wischli (commented):
Overall looks good. I have two questions and would also suggest to revert the changes within the `construct_runtime!` macro.

pallets/attestation/Cargo.toml
```diff
@@ -17,33 +17,34 @@ ctype = {features = ["mock"], path = "../ctype"}
...
-sp-keystore = {branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-core = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-io = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-keystore = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
 
 [dependencies]
+log = "0.4"
+codec = {default-features = false, features = ["derive"], package = "parity-scale-codec", version = "2.3.1"}
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+serde = {optional = true, version = "1.0.124"}
```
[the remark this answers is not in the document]
weichweich: these are not added. I Just split all the crates.io dependencies from the substrate git dependencies. We do that in most of of tomls

Diiaablo95 (commented):
Nice clean-up, and +1 for reverting the runtime configuration, at least for now.

runtimes/peregrine/Cargo.toml
```diff
@@ -100,10 +101,11 @@ runtime-benchmarks = [
   "cumulus-pallet-session-benchmarking/runtime-benchmarks",
   "delegation/runtime-benchmarks",
   "did/runtime-benchmarks",
-  "frame-benchmarking",
+  "frame-benchmarking/runtime-benchmarks",
   "frame-support/runtime-benchmarks",
   "frame-system-benchmarking",
   "frame-system/runtime-benchmarks",
+  "hex-literal",
```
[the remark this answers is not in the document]
weichweich: I changed it to being optional. We only use the hex dependency in code that is only enabled in benchmarks. So we only need the dependency when running benchmarks. Before my change we always had this dependency no matter which features are enabled.

runtimes/peregrine/src/lib.rs
weichweich: To be honest, I don't see the advantage of having them. If we always use all components, it's only error prone to list them manually. if there are cases where we don't what specific parts of the pallet, it's not clear if we left them out by mistake or which pallets are not included completely.

pallets/attestation/Cargo.toml
```diff
@@ -17,33 +17,34 @@ ctype = {features = ["mock"], path = "../ctype"}
...
 serde = {version = "1.0.124"}
-sp-core = {branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"}
-sp-io = {branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"}
-sp-keystore = {branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-core = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-io = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
+sp-keystore = {branch = "polkadot-v0.9.13", default-features = false, git = "https://github.com/paritytech/substrate"}
 
 [dependencies]
+log = "0.4"
```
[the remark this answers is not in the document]
weichweich: here you go: d634c62

runtimes/peregrine/src/lib.rs
Diiaablo95: I would say that we could just reproduce what we have in the current runtime in the new syntax? Meaning that we could manually remove the elements that we have not manually added to the current configuration, since now components are included by default. We will have to make the switch sooner or later anyway.

Diiaablo95: For instance `cargo expand`ing both runtimes to see if there's anything different by just using the default values.

weichweich: > For instance `cargo expand`ing both runtimes to see if there's anything different by just using the default values.

that'll be interesting! Actually I don't expect to see a difference. I don't know if we excluded something. 🙈

nodes/parachain/src/chain_spec/peregrine.rs
```diff
@@ -104,9 +105,10 @@ pub fn make_dev_spec(id: ParaId) -> Result<ChainSpec, String> {
 	))
 }
 
-pub fn make_new_spec(id: ParaId) -> Result<ChainSpec, String> {
+pub fn make_new_spec() -> Result<ChainSpec, String> {
 	let properties = get_properties("PILT", 15, 38);
 	let wasm = WASM_BINARY.ok_or("No WASM")?;
+	let id: ParaId = 1000.into();
```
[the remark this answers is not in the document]
weichweich: 04ff21e I actually found a constant, which is the lowest available para id, which is 2000.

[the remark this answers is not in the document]
wischli: Very nice.

runtimes/peregrine/src/lib.rs
weichweich: So the files are really messy! But the difference is:

* Call: ParachainInfo & AuraExt are added. They both don't have any calls to call.
* StorageMetadata: Aura, AuraExt, Utility & Inflation are added
* Genesis Config: Indices, TransactionPayment, Scheduler, are added
* Inherent check: Authorship
* ValidateUnsigned: ParachainSystem

Diiaablo95: So it is not worth to exclude them explicitly to maintain the same runtime?

weichweich: You can compare the runtimes here: All is with all parts of a pallet specified. Selected is with only the parts that differ specified. There are still a few changes. I think the StorageMetadata is actually a good addition, but the ParachainSystem call is a bit strange. 🤔 

[Archive.zip](https://github.com/KILTprotocol/mashnet-node/files/7879819/Archive.zip)

runtimes/peregrine/src/lib.rs
```diff
@@ -692,52 +692,52 @@ construct_runtime! {
...
 
 		// Consensus support.
 		// The following order MUST NOT be changed: Authorship -> Staking -> Session -> Aura -> AuraExt
 		Authorship: pallet_authorship::{Pallet, Call, Storage} = 20,
-		ParachainStaking: parachain_staking::{Pallet, Call, Storage, Event<T>, Config<T>} = 21,
-		Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>} = 22,
+		ParachainStaking: parachain_staking = 21,
+		Session: pallet_session = 22,
 		Aura: pallet_aura::{Pallet, Config<T>} = 23,
 		AuraExt: cumulus_pallet_aura_ext::{Pallet, Config} = 24,
```
wischli, line 710: M: Add Storage here or use the _new way_. I validated this by checking in the [Statemine configuration](https://github.com/paritytech/cumulus/blob/master/polkadot-parachains/statemine/src/lib.rs#L731).

runtimes/peregrine/src/lib.rs
wischli: > So the files are really messy! But the difference is:
> 
> * Call: ParachainInfo & AuraExt are added. They both don't have any calls to call.
> * StorageMetadata: Aura, AuraExt, Utility & Inflation are added
> * Genesis Config: Indices, TransactionPayment, Scheduler, are added
> * Inherent check: Authorship
> * ValidateUnsigned: ParachainSystem

Really interesting! I would propose to either the differences explicitly or with the new implicit way, except for the calls.

runtimes/peregrine/src/lib.rs
weichweich: 8521b59

wischli (approved):
Thanks for doing a thorough inspection of changes invoked by implicitly adding pallets to the runtime. LGTM now!

PR #316, opened 2022-01-20
feat: generic access controll

weichweich:
## fixes KILTProtocol/ticket#1697

* Use generic `AuthorizationId` in attestation
* add AuthorizationId enum to runtime-common
  * each method has it's own variant
  * first and only variant is the delegation pallet
* add AccessControl trait

## Checklist:

### DIFF
--- a/pallets/attestation/src/lib.rs
+++ b/pallets/attestation/src/lib.rs
@@ -97,7 +97,6 @@ pub mod pallet {
 		BoundedVec,
 	};
 	use frame_system::pallet_prelude::*;
-	use sp_runtime::traits::Zero;
 	use sp_runtime::DispatchError;
 
 	use ctype::CtypeHashOf;
@@ -111,13 +110,13 @@ pub mod pallet {
 
 	impl<T: Config> AttestationAccessControl<T> for () {
 		fn can_attest(_who: &T::AttesterId, _authorization_id: &T::AuthorizationId) -> Result<Weight, DispatchError> {
-			Ok(Weight::zero())
+			Err(DispatchError::Other("Unimplemented"))
 		}
 		fn can_revoke(_who: &T::AttesterId, _attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError> {
-			Ok(Weight::zero())
+			Err(DispatchError::Other("Unimplemented"))
 		}
 		fn can_remove(_who: &T::AttesterId, _attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError> {
-			Ok(Weight::zero())
+			Err(DispatchError::Other("Unimplemented"))
 		}
 	}
 
@@ -147,7 +146,7 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 
 		/// The currency that is used to reserve funds for each attestation.
-		type Currency: Currency<AccountIdOf<Self>> + ReservableCurrency<AccountIdOf<Self>>;
+		type Currency: ReservableCurrency<AccountIdOf<Self>>;
 
 		/// The deposit that is required for storing an attestation.
 		#[pallet::constant]
--- a/pallets/attestation/src/lib.rs
+++ b/pallets/attestation/src/lib.rs
@@ -103,20 +103,38 @@ pub mod pallet {
 	use kilt_support::{deposit::Deposit, traits::CallSources};
 
 	pub trait AttestationAccessControl<T: Config> {
-		fn can_attest(who: &T::AttesterId, authorization_id: &T::AuthorizationId) -> Result<Weight, DispatchError>;
-		fn can_revoke(who: &T::AttesterId, attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError>;
-		fn can_remove(who: &T::AttesterId, attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError>;
+		fn can_attest(&self, who: &T::AttesterId) -> Result<Weight, DispatchError>;
+		fn can_revoke(&self, who: &T::AttesterId, attestation: &AttestationDetails<T>)
+			-> Result<Weight, DispatchError>;
+		fn can_remove(&self, who: &T::AttesterId, attestation: &AttestationDetails<T>)
+			-> Result<Weight, DispatchError>;
+		fn authorization_id(&self) -> T::AuthorizationId;
 	}
 
-	impl<T: Config> AttestationAccessControl<T> for () {
-		fn can_attest(_who: &T::AttesterId, _authorization_id: &T::AuthorizationId) -> Result<Weight, DispatchError> {
+	impl<T> AttestationAccessControl<T> for ()
+	where
+		T: Config,
+		T::AuthorizationId: Default,
+	{
+		fn can_attest(&self, _who: &T::AttesterId) -> Result<Weight, DispatchError> {
 			Err(DispatchError::Other("Unimplemented"))
 		}
-		fn can_revoke(_who: &T::AttesterId, _attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError> {
+		fn can_revoke(
+			&self,
+			_who: &T::AttesterId,
+			_attestation: &AttestationDetails<T>,
+		) -> Result<Weight, DispatchError> {
 			Err(DispatchError::Other("Unimplemented"))
 		}
-		fn can_remove(_who: &T::AttesterId, _attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError> {
+		fn can_remove(
+			&self,
+			_who: &T::AttesterId,
+			_attestation: &AttestationDetails<T>,
+		) -> Result<Weight, DispatchError> {
 			Err(DispatchError::Other("Unimplemented"))
+		}
+		fn authorization_id(&self) -> T::AuthorizationId {
+			Default::default()
 		}
 	}
 
@@ -161,7 +179,7 @@ pub mod pallet {
 
 		type AuthorizationId: Parameter;
 
-		type AccessControl: AttestationAccessControl<Self>;
+		type AccessControl: Parameter + AttestationAccessControl<Self>;
 	}
 
 	#[pallet::pallet]
@@ -261,7 +279,7 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			claim_hash: ClaimHashOf<T>,
 			ctype_hash: CtypeHashOf<T>,
-			authorization_id: Option<AuthorizationIdOf<T>>,
+			authorization: Option<T::AccessControl>,
 		) -> DispatchResult {
 			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
 			let payer = source.sender();
@@ -278,9 +296,8 @@ pub mod pallet {
 			);
 
 			// Check for validity of the delegation node if specified.
-			if let Some(ref authorization_id) = authorization_id {
-				T::AccessControl::can_attest(&who, authorization_id)?;
-			}
+			authorization.as_ref().map(|ac| ac.can_attest(&who)).transpose()?;
+			let authorization_id = authorization.as_ref().map(|ac| ac.authorization_id());
 
 			let deposit = Pallet::<T>::reserve_deposit(payer, deposit_amount)?;
 
@@ -288,18 +305,23 @@ pub mod pallet {
 
 			log::debug!("insert Attestation");
 
+			Self::deposit_event(Event::AttestationCreated(
+				who.clone(),
+				claim_hash,
+				ctype_hash,
+				authorization_id.clone(),
+			));
 			Attestations::<T>::insert(
 				&claim_hash,
 				AttestationDetails {
 					ctype_hash,
 					attester: who.clone(),
-					authorization_id: authorization_id.clone(),
+					authorization_id,
 					revoked: false,
 					deposit,
 				},
 			);
 
-			Self::deposit_event(Event::AttestationCreated(who, claim_hash, ctype_hash, authorization_id));
 			Ok(())
 		}
 
@@ -321,7 +343,11 @@ pub mod pallet {
 		/// - Writes: Attestations, DelegatedAttestations
 		/// # </weight>
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::revoke())]
-		pub fn revoke(origin: OriginFor<T>, claim_hash: ClaimHashOf<T>) -> DispatchResultWithPostInfo {
+		pub fn revoke(
+			origin: OriginFor<T>,
+			claim_hash: ClaimHashOf<T>,
+			authorization: Option<T::AccessControl>,
+		) -> DispatchResultWithPostInfo {
 			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
 			let who = source.subject();
 
@@ -330,7 +356,9 @@ pub mod pallet {
 			ensure!(!attestation.revoked, Error::<T>::AlreadyRevoked);
 
 			if attestation.attester != who {
-				T::AccessControl::can_revoke(&who, &attestation)?;
+				authorization
+					.ok_or(Error::<T>::Unauthorized)?
+					.can_revoke(&who, &attestation)?;
 			}
 
 			// *** No Fail beyond this point ***
@@ -367,14 +395,20 @@ pub mod pallet {
 		/// - Writes: Attestations, DelegatedAttestations
 		/// # </weight>
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::remove())]
-		pub fn remove(origin: OriginFor<T>, claim_hash: ClaimHashOf<T>) -> DispatchResultWithPostInfo {
+		pub fn remove(
+			origin: OriginFor<T>,
+			claim_hash: ClaimHashOf<T>,
+			authorization: Option<T::AccessControl>,
+		) -> DispatchResultWithPostInfo {
 			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
 			let who = source.subject();
 
 			let attestation = Attestations::<T>::get(&claim_hash).ok_or(Error::<T>::AttestationNotFound)?;
 
 			if attestation.attester != who {
-				T::AccessControl::can_remove(&who, &attestation)?;
+				authorization
+					.ok_or(Error::<T>::Unauthorized)?
+					.can_remove(&who, &attestation)?;
 			}
 
 			// *** No Fail beyond this point ***
--- a/pallets/attestation/src/lib.rs
+++ b/pallets/attestation/src/lib.rs
@@ -102,39 +102,28 @@ pub mod pallet {
 	use ctype::CtypeHashOf;
 	use kilt_support::{deposit::Deposit, traits::CallSources};
 
-	pub trait AttestationAccessControl<T: Config> {
-		fn can_attest(&self, who: &T::AttesterId) -> Result<Weight, DispatchError>;
-		fn can_revoke(&self, who: &T::AttesterId, attestation: &AttestationDetails<T>)
-			-> Result<Weight, DispatchError>;
-		fn can_remove(&self, who: &T::AttesterId, attestation: &AttestationDetails<T>)
-			-> Result<Weight, DispatchError>;
-		fn authorization_id(&self) -> T::AuthorizationId;
+	pub trait AttestationAccessControl<AttesterId, AuthorizationId, T: Config> {
+		fn can_attest(&self, who: &AttesterId) -> Result<Weight, DispatchError>;
+		fn can_revoke(&self, who: &AttesterId, attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError>;
+		fn can_remove(&self, who: &AttesterId, attestation: &AttestationDetails<T>) -> Result<Weight, DispatchError>;
+		fn authorization_id(&self) -> AuthorizationId;
 		fn weight(&self) -> Weight;
 	}
 
-	impl<T> AttestationAccessControl<T> for ()
+	impl<AttesterId, AuthorizationId, T: Config> AttestationAccessControl<AttesterId, AuthorizationId, T> for ()
 	where
-		T: Config,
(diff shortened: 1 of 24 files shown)


pallets/delegation/src/lib.rs
```diff
@@ -939,3 +943,42 @@ impl<T: Config> Pallet<T> {
...
+		match Pallet::<T>::is_delegating(who, &self.0, self.1)? {
+			(true, checks) => Ok(<T as Config>::WeightInfo::can_revoke(checks)),
+			_ => Err(Error::<T>::AccessDenied.into()),
+		}
+	}
+
+	fn can_remove(&self, who: &DelegatorIdOf<T>, auth_id: &DelegationNodeIdOf<T>) -> Result<Weight, DispatchError> {
+		ensure!(auth_id == &self.0, Error::<T>::AccessDenied);
+
+		match Pallet::<T>::is_delegating(who, &self.0, self.1)? {
```
wischli, line 969: This check seems to be incorrect for removal access checking because it returns `(false, _)` when the delegation is revoked. Also, we once said we don't allow parental nodes to remove children except if thei remove their own node. 

Shouldn't this suffice?
```suggestion
		let delegation_node = <DelegationNodes<T>>::get(&self.0).ok_or(Error::<T>::DelegationNotFound)?; 
		if &who == &delegation_node.details.owner { ... } else { Err(...) }
```

runtimes/common/Cargo.toml
```diff
@@ -13,6 +13,8 @@ scale-info = {version = "1.0", default-features = false, features = ["derive"]}
 serde = {version = "1.0.124", optional = true, features = ["derive"]}
 smallvec = "1.7.0"
 
+attestation = {default-features = false, path = "../../pallets/attestation"}
```
wischli, line 16: Is this best practice or should we rather define the trait in our support pallet?

runtimes/standalone/src/lib.rs
```diff
@@ -314,6 +314,9 @@ impl attestation::Config for Runtime {
 	type Currency = Balances;
 	type Deposit = AttestationDeposit;
 	type MaxDelegatedAttestations = MaxDelegatedAttestations;
+	type AttesterId = DidIdentifier;
+	type AuthorizationId = ();
+	type AccessControl = ();
```
wischli, line 319: Q: What is the reason for not mirroring the parachain runtimes here?

wischli (commented):
Really fond of the decoupling, nice work! I have two questions and found one potential error.

Diiaablo95 (commented):
I like a lot the new approach, but some things might have been "forgotten" in the new version.

pallets/delegation/src/lib.rs
weichweich: > Also, we once said we don't allow parental nodes to remove children except if their remove their own node.

This is not about removing delegations but attestations. All parent nodes can remove the attestation?

runtimes/common/Cargo.toml
weichweich: dependencies can go out from any crate in the runtime folder to any crate in the pallet folder. The other direction does not work. We can't import anything from the runtime folder in the pallet folder since this would introduce a cycle.

runtimes/standalone/src/lib.rs
weichweich: yeah overlooked

pallets/attestation/src/lib.rs
```diff
@@ -381,22 +394,26 @@ pub mod pallet {
...
 
 			let attestation = Attestations::<T>::get(&claim_hash).ok_or(Error::<T>::AttestationNotFound)?;
 
-			let delegation_depth = if attestation.attester != who {
-				Self::verify_delegated_access(&who, &attestation, max_parent_checks)?
-			} else {
-				0
-			};
+			if attestation.attester != who {
+				let expected_auth = attestation.authorization_id.as_ref().ok_or(Error::<T>::Unauthorized)?;
```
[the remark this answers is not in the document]
weichweich: above is:
> M: use saturating

you mean:
> S: rename it to, e.g., attestation_auth_id?

pallets/delegation/Cargo.toml
```diff
@@ -28,9 +27,9 @@ sp-keystore = {branch = "polkadot-v0.9.13", default-features = false, git = "htt
 
 [dependencies]
 # Internal dependencies
+attestation = {default-features = false, path = "../attestation"}
```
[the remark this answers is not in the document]
weichweich: the dependency can never have existed since we already had the `attestation -> delegation` dependency!?
The longer i think about it, the more i feel like it should go into runtime-common.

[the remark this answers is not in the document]
weichweich: on the other hand we already have the link to the attestation pallet because of the attest permission. So even if it's not set in code, the dependency is there. So why should we not make this official by having a dependency to the attestation crate.

pallets/delegation/src/lib.rs
```diff
@@ -939,3 +943,42 @@ impl<T: Config> Pallet<T> {
...
+			(true, checks) => Ok(<T as Config>::WeightInfo::can_remove(checks)),
+			_ => Err(Error::<T>::AccessDenied.into()),
+		}
+	}
+
+	fn authorization_id(&self) -> DelegationNodeIdOf<T> {
+		self.0
+	}
+
+	fn weight(&self) -> Weight {
```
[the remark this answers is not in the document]
weichweich: let's have a call

pallets/attestation/src/lib.rs
```diff
@@ -381,22 +394,26 @@ pub mod pallet {
 		/// - Reads per delegation step P: delegation::Delegations
 		/// - Writes: Attestations, DelegatedAttestations
 		/// # </weight>
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::remove(*max_parent_checks))]
+		#[pallet::weight(
+			<T as pallet::Config>::WeightInfo::remove()
+			+ authorization.as_ref().map(|ac| ac.weight()).unwrap_or(0)
```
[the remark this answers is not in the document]
weichweich: f963bfc

pallets/attestation/src/lib.rs
```diff
@@ -329,11 +338,14 @@ pub mod pallet {
 		/// - Reads per delegation step P: delegation::Delegations
 		/// - Writes: Attestations, DelegatedAttestations
 		/// # </weight>
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::revoke(*max_parent_checks))]
+		#[pallet::weight(
+			<T as pallet::Config>::WeightInfo::revoke()
+			+ authorization.as_ref().map(|ac| ac.weight()).unwrap_or(0)
```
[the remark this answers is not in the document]
weichweich: f963bfc

pallets/attestation/src/lib.rs
```diff
@@ -235,12 +267,15 @@ pub mod pallet {
 		///   DelegatedAttestations
 		/// - Writes: Attestations, (DelegatedAttestations)
 		/// # </weight>
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::add())]
+		#[pallet::weight(
+			<T as pallet::Config>::WeightInfo::add()
+			+ authorization.as_ref().map(|ac| ac.weight()).unwrap_or(0)
```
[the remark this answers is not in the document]
weichweich: f963bfc

pallets/attestation/src/lib.rs
```diff
@@ -257,58 +292,32 @@ pub mod pallet {
...
 
 			log::debug!("insert Attestation");
 
-			// write delegation record, if any
-			if let Some((id, delegated_attestation)) = delegation_record {
-				<DelegatedAttestations<T>>::insert(id, delegated_attestation);
-			}
-
-			<Attestations<T>>::insert(
+			Self::deposit_event(Event::AttestationCreated(
```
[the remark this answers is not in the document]
weichweich: Good point, I think we didn't talk about that yet.
The delegated attestations can't be part of the attestation pallet, since it would then depend on the delegation pallet (even tho not in code and only in logic).
Do we need this storage? I guess yes. What if a delegation node gets deleted, but there are still attestations remaining. They cant get removed anymore?

pallets/attestation/src/lib.rs
```diff
@@ -90,21 +90,55 @@ pub use crate::{attestations::AttestationDetails, default_weights::WeightInfo, p
...
 		traits::{Currency, Get, ReservableCurrency},
 		BoundedVec,
 	};
 	use frame_system::pallet_prelude::*;
+	use sp_runtime::DispatchError;
+
+	use ctype::CtypeHashOf;
 	use kilt_support::{deposit::Deposit, traits::CallSources};
 
+	pub trait AttestationAccessControl<AttesterId, AuthorizationId> {
```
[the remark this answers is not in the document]
weichweich: 466df1d

pallets/attestation/src/lib.rs
```diff
@@ -342,16 +354,17 @@ pub mod pallet {
 
 			ensure!(!attestation.revoked, Error::<T>::AlreadyRevoked);
 
-			let delegation_depth = if attestation.attester != who {
-				Self::verify_delegated_access(&who, &attestation, max_parent_checks)?
-			} else {
-				0
-			};
+			if attestation.attester != who {
+				let expected_auth = attestation.authorization_id.as_ref().ok_or(Error::<T>::Unauthorized)?;
```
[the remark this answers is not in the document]
weichweich: 89800d9

pallets/attestation/src/lib.rs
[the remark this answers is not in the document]
weichweich: 89800d9

pallets/delegation/src/lib.rs
```diff
@@ -271,6 +273,8 @@ pub mod pallet {
 		UnauthorizedRemoval,
 		/// The delegation creator is not allowed to create the delegation.
 		UnauthorizedDelegation,
+		/// The access was not granted because of an insufficient delegation.
```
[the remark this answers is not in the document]
weichweich: 7eec4c5 ?

pallets/attestation/src/mock.rs
```diff
@@ -82,6 +74,46 @@ where
 	}
 }
 
+#[derive(Clone, Debug, Encode, Decode, TypeInfo, PartialEq, Eq)]
+#[scale_info(skip_type_params(T))]
+pub struct MockAccessControl<T: Config>(pub T::AttesterId);
```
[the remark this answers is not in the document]
weichweich: a48fa14

pallets/attestation/src/lib.rs
```diff
@@ -154,17 +168,14 @@ pub mod pallet {
...
-	#[pallet::getter(fn delegated_attestations)]
-	pub type DelegatedAttestations<T> = StorageMap<
-		_,
-		Blake2_128Concat,
-		DelegationNodeIdOf<T>,
-		BoundedVec<ClaimHashOf<T>, <T as Config>::MaxDelegatedAttestations>,
-	>;
+	#[pallet::getter(fn external_attestations)]
+	pub type ExternalAttestations<T> =
+		StorageDoubleMap<_, Twox64Concat, AuthorizationIdOf<T>, Blake2_128Concat, ClaimHashOf<T>, bool, ValueQuery>;
```
wischli, line 178: Regarding the question whether we want to have an additional map `ClaimHash --> AuthorizationId`:

Is there any other use-case for `ExternalAttestations ` other than distinguishing between internal and external attestations so far? 

For now I don't see a use-case to store the second map, especially as we did not need it before. However, it will be expensive to add later by iterating over the ExternalAttestations maps.

wischli (commented):
I have a few minor comments. Overall it looks good to me, especially the clean test structure. 

Not super happy about the sneaked in clippy stuff for a newer version. I think this should have been handled in a separate PR which includes switching to a newer rust version for the repository. But it's fine with me.

wischli:
> Delegation hierarchy was deleted but (delegated) attestation remains?
> * Option A: accept that there are attestations that cannot be removed/revoked since the delegation node was gone (only deposit owner can remove attestation)
> * Option B: Just check in the attestation pallet? We don't know how the AuthorizationId looks like in the attestation pallet.

This definitely is a hard question to answer. I propose to discuss this with others, maybe at the end of our deposit meeting or in a separate one.

Previously we revoked all these attestations, correct? I am not a fan of having irrevocable/irremovable attestations. If the main reason to not revoke/remove them upon removing Delegation Hierarchies is computational complexity in a potentially deep tree, then we should add a `request_remove` function callable by anyone. On success, this function could reward the caller with (a portion) of the deposit.

pallets/delegation/src/lib.rs
```diff
@@ -273,6 +275,8 @@ pub mod pallet {
 		UnauthorizedRemoval,
 		/// The delegation creator is not allowed to create the delegation.
 		UnauthorizedDelegation,
+		/// The operation wasn't allowed because of insufficient rights.
+		AccessDenied,
```
Diiaablo95, line 279: Should maybe this be part of the attestation crate? It is only used in the the access control, which requires the implementer to return a generic `DispatchError`, so maybe it would make sense to either declare it in the attestation pallet, or as a separate error?

pallets/delegation/src/mock.rs
```diff
@@ -176,24 +179,30 @@ pub mod runtime {
...
-	type TestDelegateSignature = (SubjectId, Vec<u8>);
-	type TestBalance = runtime_common::Balance;
-	type TestCtypeHash = runtime_common::Hash;
+	pub type Hash = sp_core::H256;
+	pub type Balance = u128;
+	pub type Signature = MultiSignature;
+	pub type AccountPublic = <Signature as Verify>::Signer;
+	pub type AccountId = <AccountPublic as IdentifyAccount>::AccountId;
+
+	pub const UNIT: Balance = 10u128.pow(15);
```
Diiaablo95, line 202: What is this value used for?

Diiaablo95 (commented):
I have a couple of questions regarding the design. Once those are answered/addressed, I would be happy to approve this since otherwise it LGMT!

pallets/delegation/src/lib.rs
weichweich: The point of defining it here is that the Delegation pallet could send more detailed errors. I Would actually prefer if we could give a reason for not allowing access.
If I understood separate error correctly, this would be something that cannot be shown to the user?

pallets/attestation/src/lib.rs
weichweich: > The main point of having a map for delegated attestations was/is to make it easier, when a delegation is revoked, to revoke, if needed, all attestations issued with that specific delegation.

This is a big point. But looking up, which attestations where issued by an delegation node or similar might even be a bigger use case in the future.
Having the storage in the attestation pallet doesn't mean that it's inaccessible for the delegating pallet.
Regarding the cascading deletion. I think we should just fail in that case. These cascading things are tough to weight and should rather be handled by the SDK/UI.

> I would store any information in the VCO and not in the attestation pallet, which otherwise becomes a big pot of everything.

I disagree with the big pot of everything. The `ExternalAttestations` is merely an index for the `Attestation` store. Therefore it should life in the attestation pallet. If we would start to split that across pallets, there would be X different indices that need to be kept up to date.

pallets/attestation/src/mock.rs
```diff
@@ -82,6 +74,81 @@ where
...
+		} else {
+			Err(DispatchError::Other("Unauthorized"))
+		}
+	}
+
+	fn authorization_id(&self) -> T::AuthorizationId {
+		self.0.clone()
+	}
+
+	fn can_attest_weight(&self) -> Weight {
```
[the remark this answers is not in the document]
weichweich: [351e9e4](https://github.com/KILTprotocol/mashnet-node/pull/316/commits/351e9e4bba294bc264d183288d2d83b7a0ab47ac)

pallets/delegation/src/mock.rs
weichweich: [1016572](https://github.com/KILTprotocol/mashnet-node/pull/316/commits/1016572af6dc9a4138bdeec705153437010a2e0b)

pallets/delegation/src/lib.rs
Diiaablo95: I mean the dispatch error would still be bubbled up to the user, I guess. This is not used anywhere inside the pallet, and furthermore it is an attestation error, really, and not a delegation one.

pallets/attestation/src/lib.rs
Diiaablo95: > Regarding the cascading deletion. I think we should just fail in that case. These cascading things are tough to weight and should rather be handled by the SDK/UI.

The operation is not meant to be triggered automatically on deletion, but the original attester (or a delegator) can lookup the attestation hashes, batch them, and revoke/remove them. And I am just reporting what I was told back then, since I also questioned the utility of this additional map. Now it makes definitely sense to have it, but it was never meant to be used for "cleaning up" automatically when a delegation is removed.

> I disagree with the big pot of everything. The ExternalAttestations is merely an index for the Attestation store.

All the attestations are still stored in the attestation pallet, so in my opinion it makes sense that each "authorisable" pallet would keep its own index, especially if something additional is needed or something changes. This would be easier if the access control trait would expose hooks for main operations, which it does not right now and would require further changes. But of course this is just my point of view.

pallets/delegation/src/lib.rs
weichweich: It is used in the delegation pallet. The implementation of `AttestationAccessControl` belongs to the delegation pallet

pallets/attestation/src/mock.rs
```diff
@@ -82,6 +74,81 @@ where
 	}
 }
 
+#[derive(Clone, Debug, Encode, Decode, TypeInfo, PartialEq, Eq)]
```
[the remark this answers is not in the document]
weichweich: [e1d37ce](https://github.com/KILTprotocol/mashnet-node/pull/316/commits/e1d37ce7c17e35d507bfbd226a13fb5476bea42d)

pallets/attestation/src/mock.rs
```diff
@@ -82,6 +74,81 @@ where
...
+	}
+	fn can_revoke_weight(&self) -> Weight {
+		0
+	}
+	fn can_remove_weight(&self) -> Weight {
+		0
+	}
+}
+
+pub fn insert_attestation<T: Config>(claim_hash: ClaimHashOf<T>, details: AttestationDetails<T>) {
```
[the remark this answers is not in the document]
weichweich: I wouldn't since we call `reserve`. We can expect that this call works.

pallets/delegation/src/access_control.rs
```diff
@@ -0,0 +1,511 @@
...
+///
+/// Can revoke attestations if
+///    * delegation node of sender is not revoked
+///    * sender delegation node is equal to OR parent of the delegation node
+///      stored in the attestation
+///
+/// Can remove attestations if <the same as revoke>
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq, TypeInfo)]
+pub struct DelegationAc<T: Config> {
+	pub(crate) sender_node_id: DelegationNodeIdOf<T>,
```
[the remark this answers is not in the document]
weichweich: [c5fec0c](https://github.com/KILTprotocol/mashnet-node/pull/316/commits/c5fec0c9ba40eba90e06c66746fc6e8d6722383d)

runtimes/standalone/src/lib.rs
weichweich: [e675ecc](https://github.com/KILTprotocol/mashnet-node/pull/316/commits/e675ecc7ef7caa82c9b9810be996997b15b86717)

pallets/delegation/src/lib.rs
Diiaablo95: Makes sense.

pallets/attestation/src/lib.rs
weichweich: I still don't see why each pallet should keep their own index. That just makes it more complicated. 🤷 Some pallets might then also choose to not have the index? The code to keep track of the attestations needs to be duplicated for all the other pallets too.

If we want to have the hooks or not is a separate question. And I think we will need them later when we have more advanced use cases.

Diiaablo95: Yes, that for me is one more point: pallets might not need to have an index at all. And I agree that it makes the code more complicated, and that why I am saying that the current solution also works (and can always be migrated in the future, if needed).

weichweich: I would guess that all pallets need an index. But i think only time will tell. :D

tjwelde:
AFAICS, it would be hard to identify the pallet which was used for access control here. Can we add something like a prefix or struct identifier to the Access Control, which would be added to the authorization id, e.g.?

weichweich:
> AFAICS, it would be hard to identify the pallet which was used for access control here. Can we add something like a prefix or struct identifier to the Access Control, which would be added to the authorization id, e.g.?

We already have something like that. [The authorization ID is defined here](https://github.com/KILTprotocol/mashnet-node/blob/2730352bb3b84a512464e86ba8d5b52fb5ea2374/runtimes/common/src/authorization.rs#L31). You should be able to use the enum on the SDK side right?

Diiaablo95:
I think the point is that an `AttestationDetails` struct has `pub authorization_id: Option<AuthorizationIdOf<T>>`, which is simply `Parameter + MaxEncodedLen`. So it might not be possible to understand whether the given attestation was issued under a delegation or under some other access control logic.

weichweich:
> So it might not be possible to understand whether the given attestation was issued under a delegation or under some other access control logic.

That is the point of this PR. The attestation pallet should not care about that. But since the `authorization_id` is actually an Enum, you should be able to distinguish where the attestation was authorized in the SDK.

Diiaablo95 (commented):
Nothing major changed since last time, so I would be in favour of merging this and fix anything else in a different PR, so that we can start working on supporting this in the SDK.

Diiaablo95 (approved):

PR #318, opened 2022-01-26
refactor: switch CandidatePool to be CountedStorageMap

wischli:
## fixes KILTProtocol/ticket#1778
* Removes `CandidateCount` by changing `CandidatePool` to be `CountedStorageMap` instead of `StorageMap`
* Applies new `StorageVersion` paradigm by migrating to the supported `u16` pallet storage version instead of our custom enum

We also use a `DidEndpointsCount` in the DID pallet. However, there is no `DoubleCountedStorageMap` so far which we would require to map `DidIdentifier -> ServiceEndpointId -> DidEndpoint` while we count the first mapping. The only solution would be to nest the second map `ServiceEndpointId -> DidEntpoint` inside a counted version of the first map `DidIdentifier -> ServiceEndpointId`. This does not seem to be feasible to me.

AFAIK, we should not be required to write a migration because we only append a value to our existing `CandidatePool` map. However, I would still bump the Staking `StorageVersion` number and maybe even write a migration check to be bulletproof.

## ToDo

- [x] Increment Staking `StorageVersion`?
- [x] Potentially write a `safety-first` migration
- [x] Test `try-runtime`
- [x] Run benchmarks after receiving review

## How to test:
Run unit tests

## Checklist:


Diiaablo95 (commented):
Overall LGTM! Not sure whether it would be faster to try the runtime upgrade on a local chain and inspect whether there is any issue. But I also saw that Parity has some migrations in place, so it might make sense.

weichweich (commented):
Try-runtime checks to see if the count is correct would be nice. PoV size is OK since there are not a lot of candidates.

pallets/parachain-staking/src/tests.rs
```diff
@@ -3924,3 +3924,22 @@ fn update_total_stake_no_collator_changes() {
...
+
+			// Change to incorrect value via unhashed
+			frame_support::storage::unhashed::put(&CandidatePool::<Test>::counter_storage_final_key(), &100u32);
+			assert_eq!(CandidatePool::<Test>::count(), 100u32);
+
+			// Initialize counter
+			assert_eq!(CandidatePool::<Test>::initialize_counter(), 1);
+			assert_eq!(CandidatePool::<Test>::count(), 1);
+		})
+}
```
weichweich, line 3945: `CandidatePool::<Test>::initialize_counter()` is provided by parity? I think we don't need to test this function. Also the name `migration_to_v7_works` is a bit misleading since the migration itself is not tested.

runtimes/peregrine/src/lib.rs
```diff
@@ -1017,11 +1017,30 @@ impl OnRuntimeUpgrade for SchedulerMigrationV3 {
...
 			Scheduler::current_storage_version()
 		);
-
 		Ok(())
 	}
 }
 
+// Migration for ParachainStaking pallet to mutate CandidatePool to be
+// a CountedStorageMap and switch to new Pallet StorageVersion paradigm.
+pub struct ParachainStakingMigrationV7;
```
weichweich, line 1026: Could we not move this struct to the stacking pallet and just import it from there?

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,80 @@ pub enum StakingStorageVersion {
...
+		traits::{Get, GetStorageVersion, StorageVersion as NewStorageVersion},
+	};
+	use log::info;
+	use sp_runtime::traits::Zero;
+
+	// Get storage items into scope which are removed during this migration
+	generate_storage_alias!(ParachainStaking, CandidateCount => Value<u32>);
+	generate_storage_alias!(ParachainStaking, StorageVersion => Value<StakingStorageVersion>);
+
+	pub fn pre_migrate<T: Config>() -> Result<(), &'static str> {
```
weichweich, line 57: should be possible to have `OnRuntimeUpgrade` implemented here?

weichweich (approved):
LGTM, just a minor thing... I would move the `ParachainStakingMigrationV7` to the pallet. The struct can have a generic type for the used runtime?

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,80 @@ pub enum StakingStorageVersion {
...
+		pallet_prelude::Weight,
+		storage::migration::{have_storage_value, remove_storage_prefix},
+		traits::{Get, GetStorageVersion, StorageVersion as NewStorageVersion},
+	};
+	use log::info;
+	use sp_runtime::traits::Zero;
+
+	// Get storage items into scope which are removed during this migration
+	generate_storage_alias!(ParachainStaking, CandidateCount => Value<u32>);
+	generate_storage_alias!(ParachainStaking, StorageVersion => Value<StakingStorageVersion>);
```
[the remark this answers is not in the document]
wischli: Yup! I also didn't know it before but found it in a Substrate PR.

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,80 @@ pub enum StakingStorageVersion {
...
+			"CandidateCount should not exist anymore"
+		);
+		assert!(
+			!have_storage_value(b"StakePallet", b"CandidateCount", &[]),
+			"CandidateCount should not exist anymore"
+		);
+		assert!(
+			!CandidatePool::<T>::count().is_zero(),
+			"Candidate counter should have been set"
+		);
```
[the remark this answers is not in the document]
wischli: I also thought about this. I will just add it.

pallets/parachain-staking/src/tests.rs
wischli: Removed in 722a317f411931f41d013ff81372d038830ed581

pallets/parachain-staking/src/migrations.rs
wischli: Yup! [73c4cf4](https://github.com/KILTprotocol/mashnet-node/pull/318/commits/73c4cf4c0de0bbe5d765c81acf038e44e60ba2e5)

pallets/parachain-staking/src/tests.rs
```diff
@@ -3924,4 +3924,3 @@ fn update_total_stake_no_collator_changes() {
 			);
 		});
 }
-
```
weichweich: 😵‍💫 files always end with a new line! 🤪

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,87 @@ pub enum StakingStorageVersion {
...
+			// migrate CandidateCount
+			frame_support::migration::remove_storage_prefix(b"ParachainStaking", b"CandidateCount", &[]);
+			let candidate_count = CandidatePool::<T>::initialize_counter();
+
+			// migrate StorageVersion to new paradigm
+			frame_support::migration::remove_storage_prefix(b"ParachainStaking", b"StorageVersion", &[]);
+			NewStorageVersion::new(7).put::<Pallet<T>>();
+
+			info!("💰 completed parachain staking migration to v7 ✅",);
+			T::DbWeight::get().reads_writes(candidate_count.into(), candidate_count.saturating_add(3).into())
```
Diiaablo95, line 73: I am not sure where `candidate_count.saturating_add(3).into()` comes from. Does `remove_storage_prefix` writes `None` into every entry?

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,87 @@ pub enum StakingStorageVersion {
...
+
+			info!("💰 parachain staking migration to v7 passes PRE migrate checks ✅",);
+			Ok(())
+		}
+
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade() -> Result<(), &'static str> {
+			// check count
+			assert!(
+				!frame_support::migration::have_storage_value(b"StakePallet", b"CandidateCount", &[]),
```
Diiaablo95, line 95: Shouldn't we check using the old pallet name?

pallets/parachain-staking/src/migrations.rs
```diff
@@ -32,24 +35,87 @@ pub enum StakingStorageVersion {
...
+				"CandidateCount should not exist anymore"
+			);
+			assert!(
+				!CandidatePool::<T>::count().is_zero(),
+				"Candidate counter should have been set"
+			);
+
+			// check StorageVersion
+			assert!(
+				!frame_support::migration::have_storage_value(b"StakePallet", b"StorageVersion", &[]),
```
Diiaablo95, line 105: Same here, shouldn't we use the old pallet name?

Diiaablo95 (commented):
LGTM! I just have a couple of questions.

pallets/parachain-staking/src/migrations.rs
wischli: What do you mean by old pallet name? The pallet name has never been changed?

Diiaablo95: Isnt't the first parameter the name of the pallet? Is that `ParachainStaking` or `StakePallet`?

wischli: You are right, I had tunnel vision. `StakePallet` is only used in the unit test 🙈 Great catch! I will derive the pallet name directly from storage so that the migration will be correct independent of the corresponding runtime.

pallets/parachain-staking/src/migrations.rs
wischli: The weight of `remove_storage_prefix` is 1 write + some base weight. So these two calls plus the `NewStorageVersion` write amounts to 3.

wischli: Overall the statement seems incorrect though, it should be `(candidate_count + 1, 4)` because we don't write for every candidate.

pallets/parachain-staking/src/migrations.rs
wischli: 1866d37a58f36b18fc57def17555a1e47ac3e60a

pallets/parachain-staking/src/migrations.rs
wischli: 1866d37a58f36b18fc57def17555a1e47ac3e60a

Diiaablo95 (approved):
LGTM

PR #319, opened 2022-01-31
feat: upgrade to Polkadot v0.9.16

wischli:
## fixes KILTProtocol/ticket#1793
* Upgrades to most recent Polkadot version v0.9.16
* Removes crowdloan pallet from the `pallets` directory
* Applies [deprecation of `AllPallets` hook](https://github.com/paritytech/substrate/pull/10043)  which executed pallets in the reverse order of their definition within `construct_runtime!` with the exception that `System` was always executed first
  * If we want to switch to the new _default_ hook `AllPalletsWithSystem`, we have to reorder some of our pallets, see 625df70 and 6363a5c
* Adds asset quota (Substrate [#10382](https://github.com/paritytech/substrate/pull/10382))
* Adds `state_version` to runtime ([Substrate #9732](https://github.com/paritytech/substrate/pull/9732))
  * TODO in future PR: Switch to new trie format
* Adds pallet_preimage + runs migration for Scheduler ([Substrate #10356](https://github.com/paritytech/substrate/pull/10356) + [Polkadot #4166](https://github.com/paritytech/polkadot/pull/4166))
  * TODO in future PR: Set appropriate deposits for the Preimage deposit 
* Substrate switched to generate pallet storage info per default ([Substrate #10662](https://github.com/paritytech/substrate/pull/10662))
  * By using the macro`#[pallet::without_storage_info]`, we skip generating storage info and can worry about this change later. It will be required for PoV benchmarks
  * TODO in future PR: Implement `MaxEncodedLen` wherever necessary

## ToDo

- [x] Fix benchmarks / benchmark unit tests for DID
- [x] Fix benchmark / benchmark unit tests for DID Lookup
- [x] Fix unit tests for staking
- [x] Investigate order execution hook. If we want to be safe, we use `AllPalletsReversedWithSystemFirst` which keeps the exact same order. Wondering why other teams switched to `AllPalletsWithSystem` without any reordering.
- [x] Run pallet scheduler benchmarks
- [x] Run pallet preimage benchmarks
- [x] Execute `try-runtime` against Peregrine-stg
- [x] Execute `try-runtime` against Spiritnet

## Checklist:


weichweich (approved):
lgtm

pallets/did/src/mock_utils.rs
```diff
@@ -93,12 +93,16 @@ pub fn generate_base_did_creation_details<T: Config>(
 	}
 }
 
-pub fn generate_base_did_details<T: Config>(authentication_key: DidVerificationKey) -> DidDetails<T> {
+pub fn generate_base_did_details<T>(authentication_key: DidVerificationKey) -> DidDetails<T>
+where
+	T: crate::Config,
```
Diiaablo95, line 98: We already import `Config`, so we can use the same.

Diiaablo95 (approved):
A super minor nitpick, but otherwise LGMT!

pallets/did/src/mock_utils.rs
wischli: 825a7fb

nodes/parachain/src/command.rs
```diff
@@ -40,6 +38,31 @@ use sp_core::hexdisplay::HexDisplay;
...
+	fn is_peregrine(&self) -> bool;
+	fn is_spiritnet(&self) -> bool;
+}
+
+impl IdentifyChain for dyn sc_service::ChainSpec {
+	fn is_peregrine(&self) -> bool {
+		self.id().starts_with("kilt_peregrine")
+	}
+	fn is_spiritnet(&self) -> bool {
+		self.id().starts_with("kilt_spiritnet")
```
weichweich, line 51: does this also work for existing chain specs?

nodes/parachain/src/command.rs
```diff
@@ -346,14 +373,16 @@ pub fn run() -> Result<()> {
...
 				match cli.runtime.as_str() {
-					"peregrine" => crate::service::start_node::<MashRuntimeExecutor, peregrine_runtime::RuntimeApi>(
-						config,
-						polkadot_config,
-						id,
-					)
-					.await
-					.map(|r| r.0)
-					.map_err(Into::into),
+					"peregrine" => {
```
weichweich, line 376: could we use chain identity here instead?

nodes/parachain/src/command.rs
wischli: It does not. Should be fixed in [a5e9d01](https://github.com/KILTprotocol/mashnet-node/pull/319/commits/a5e9d01128c1c5eb2c0761278f41fe1ac282911d)

nodes/parachain/src/command.rs
wischli: Sure, good catch! [a5e9d01](https://github.com/KILTprotocol/mashnet-node/pull/319/commits/a5e9d01128c1c5eb2c0761278f41fe1ac282911d)

wischli:
Since it has been running smoothly on the local setup, I will merge now and test on Spiritnet.

PR #322, opened 2022-02-07
feat: empty runtime

weichweich:


nodes/parachain/res/void.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "telemetryEndpoints": null,
+  "protocolId": null,
+  "properties": {
+    "ss58Format": 38,
+    "tokenDecimals": 15,
+    "tokenSymbol": "VOID"
+  },
+  "relay_chain": "kusama",
+  "para_id": 2085,
```
wischli: Did we reserve this number or this is C&P result?

nodes/parachain/Cargo.toml
```diff
@@ -19,6 +19,7 @@ substrate-build-script-utils = {git = "https://github.com/paritytech/substrate",
 peregrine-runtime = {path = "../../runtimes/peregrine"}
 runtime-common = {path = "../../runtimes/common"}
 spiritnet-runtime = {path = "../../runtimes/spiritnet"}
+empty-runtime = {path = "../../runtimes/empty"}
```
wischli, line 22: Shouldn't we stick to `void` here to keep it consistent? Else name the chain spec `empty`.

runtimes/empty/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+
+impl cumulus_pallet_aura_ext::Config for Runtime {}
+
+parameter_types! {
+	pub const MaxAuthorities: u32 = constants::staking::MAX_CANDIDATES;
+}
+
+impl pallet_aura::Config for Runtime {
+	type AuthorityId = AuthorityId;
+	//TODO: handle disabled validators
```
wischli, line 207: Can be removed

nodes/parachain/Cargo.toml
weichweich: yeah noticed halfway through that void is cooler

wischli (approved):
LGTM. I would question whether we actually want to merge this though? It will create conflicts when merging master into develop after #321.

weichweich:
The fun is over... until the void is called again from the ashes.

PR #323, opened 2022-02-09
create sporran spec

weichweich:
## fixes KILTProtocol/ticket#1813


nodes/standalone/src/chain_spec.rs
```diff
@@ -162,45 +121,47 @@ impl Alternative {
...
 		})
 	}
 
 	pub(crate) fn from(s: &str) -> Option<Self> {
 		match s {
 			"dev" => Some(Alternative::Development),
 			"kilt-testnet" => Some(Alternative::KiltTestnet),
-			"kilt-devnet" => Some(Alternative::KiltDevnet),
-			"mashnet-staging" => Some(Alternative::MashnetStaging),
+			"sporran-new" => Some(Alternative::SporranTestnet),
```
tjwelde, line 164: C: Should this maybe be rather something like `sporran-testnet`? Since the word `new` might not tell anyone much looking at it in the future.

nodes/standalone/src/chain_spec.rs
```diff
@@ -115,43 +105,12 @@ impl Alternative {
...
-					get_account_id_from_secret::<ed25519::Public>("//Bob"),
-					get_account_id_from_secret::<sr25519::Public>("//Alice"),
-					get_account_id_from_secret::<sr25519::Public>("//Bob"),
-	],
-						)
-					},
-					vec![],
-					None,
-					None,
-					Some(properties),
```
tjwelde, line 130: Q: Are the `properties` given to the development spec at another point now?

nodes/standalone/src/chain_spec.rs
weichweich: It's called `sparran-new` since it creates a new chainspec. For running a node you always need to use the file `--chain dev-specs/mashnet-node/sporran.json`.

nodes/standalone/src/chain_spec.rs
tjwelde: oh sorry, seems to be a confusion because of the diff view here.

weichweich: yeah there was some strange indent going on and i fixed that. But the diff looks strange after that. 😕

nodes/standalone/src/chain_spec.rs
tjwelde: ah ok, understood. Than fine for me

tjwelde (approved):
LGTM

wischli (commented):
LGTM

PR #326, opened 2022-02-15
refactor: migrate pallets' StorageVersion

wischli:
## fixes KILTProtocol/ticket#1627
* Sets `StorageVersion` via pallet macro for all pallets (excluding Staking which is handled in #318)
* Adds Migrations for `DID` and `Delegation` pallets which already used a custom versioning scheme (as done in #318)

## How to test:

```
RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release -p kilt-parachain --features try-runtime -- try-runtime --chain=spiritnet-dev --execution=Native --no-spec-name-check on-runtime-upgrade live --uri=wss://spiritnet.kilt.io:443/para  
```

## Checklist:


Diiaablo95 (approved):
LGTM! Just a question: are we sure that the new storage version is written to storage for a pallet even if we override the default `on_runtime_upgrade()` hook? I am referring to the `delegation` and `did` pallets.

Diiaablo95:
Nevermind, we have checks in `post_upgrade()`, so it should be fine 👍

wischli:
> Nevermind, we have checks in `post_upgrade()`, so it should be fine 👍

In the post check, we also print the value to the console, so feel free to check yourself 😊

weichweich (approved):
LGTM, Only one question. We don't have migrations in pallets that didn't have a storage before. Don't we need to write the current version to storage?

wischli:
> 


> LGTM, Only one question. We don't have migrations in pallets that didn't have a storage before. Don't we need to write the current version to storage?


AFAICT we mainly need the migration to remove the old storage entry. I logged the `current_storage_version` for the pallets which did not have any prior to this PR and they were all set to the value provided in the pallet macro (1) without a migration. 

Just checked for the did and delegation pallets: Seems like we also don't even need to set the StorageVersion in the migration. So I suppose I can just remove these setters.

PR #327, opened 2022-02-16
feat: bound pallets in runtime storage size

wischli:
## fixes KILTProtocol/ticket#1777
* Implements/derives `MaxEncodedLen` for all required pallets' structs such that each pallet's storage size is bound
* Removes `#[pallet::without_storage_info]` from all pallets 

The usage of the `#[codec(mel_bound(Generic: TraitBound))]` feature is often required because the compiler is not smart enough to derive `MaxEncodedLen`
* `mel_bound()` means no bound is added at all. We can use this if encapsulated associated types are already bound by the `MaxEncodedLen` trait
* If we provide attributes, we tell the macro which trait bounds to add. This was only required in the staking pallet where we used unbound generics for the AccountId and balance. In 8a5d744a54343062f1d74d778e43a4c7410a7e3f you can see how many bounds could be removed eventually. Shawn did it similarly in [this Substrate PR](https://github.com/paritytech/substrate/commit/54dc43478f9f4f252fe692bffb83f867c29638cb#diff-2d9d52e6c7d98cac411c9f83d851a0ee4cfaad74140ac2d0a2a07007db1a1f3eL342)

## Checklist:

### DIFF
--- a/pallets/pallet-web3-names/src/web3_name.rs
+++ b/pallets/pallet-web3-names/src/web3_name.rs
@@ -20,7 +20,7 @@
 
 use sp_std::{fmt::Debug, marker::PhantomData, vec::Vec};
 
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::{ensure, sp_runtime::SaturatedConversion, traits::Get, BoundedVec};
 use scale_info::TypeInfo;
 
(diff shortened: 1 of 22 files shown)


pallets/pallet-web3-names/src/web3_name.rs
```diff
@@ -30,9 +30,13 @@ use crate::{Config, Error};
 ///
 /// It is bounded in size (inclusive range [MinLength, MaxLength]) and can only
 /// contain a subset of ASCII characters.
-#[derive(Encode, Decode, TypeInfo)]
-#[scale_info(skip_type_params(T, MaxLength, MinLength))]
-pub struct AsciiWeb3Name<T, MinLength, MaxLength>(pub(crate) BoundedVec<u8, MaxLength>, PhantomData<(T, MinLength)>);
+#[derive(Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T, MinLength, MaxLength))]
+#[codec(mel_bound())]
+pub struct AsciiWeb3Name<T: Config, MinLength: Get<u32>, MaxLength: Get<u32>>(
```
Diiaablo95, line 36: Why did this have to be linked to the pallet's `Config` trait?

PR #328, opened 2022-02-22
ci: fix pipeline

weichweich:
## fixes KILTProtocol/ticket#1695

* You can [check that the yaml is valid in gitlab](https://gitlab.com/kiltprotocol/mashnet-node/-/ci/editor?branch_name=aw-image-pipeline&tab=2)

## Checklist:


.maintain/push-image.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+source_tag=$1
+target_tag=$2
+
+# publish to docker hub
+docker tag kiltprotocol/mashnet-node:$source_tag ${DOCKER_HUB_STANDALONE}:$target_tag
+docker tag kiltprotocol/kilt-node:$source_tag ${DOCKER_HUB_PARACHAIN}:$target_tag
```
Diiaablo95, line 8: Isn't `kiltprotocol/kilt-node` = `${DOCKER_HUB_PARACHAIN}`?

.maintain/build-image.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+target_tag=$1
+
+# build parachain image and standalone image
+docker build --cache-from $AWS_REGISTRY/kilt-parachain/collator:$target_tag --build-arg NODE_TYPE=kilt-parachain -t kiltprotocol/kilt-node:$target_tag .
```
Diiaablo95, line 6: I think we should replace all occurrences of `kiltprotocol/*` with the relative env variable.

.maintain/push-image.sh
```diff
@@ -0,0 +1,18 @@
...
+
+docker push ${DOCKER_HUB_STANDALONE}:$target_tag
+docker push ${DOCKER_HUB_PARACHAIN}:$target_tag
+
+# publish to AWS
+docker tag kiltprotocol/mashnet-node:$source_tag $AWS_REGISTRY/kilt/prototype-chain:$target_tag
```
Diiaablo95, line 14: Same here for the image name, that should be replaced with the respective env variable.

Diiaablo95 (commented):
Looks definitely cleaner! Perhaps I would make more use of the env variables. Plus, a question: I don't see any occurrence of the `latest` tag anywhere. Is that somehow to be triggered manually?

.maintain/build-image.sh
weichweich: I'll rename that. The idea is, that the build script builds the images with a name X which doesn't really matter how it's called.
The publish script tags X to the docker and AWS names, and pushes those two.

weichweich:
> I don't see any occurrence of the `latest` tag anywhere. Is that somehow to be triggered manually?

The magic happens [in the gitlab ci file](https://github.com/KILTprotocol/mashnet-node/pull/328/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R56-R58).

* if branch -> create a tag with short commit SHA
* if branch -> push images with tag `latest-{branch-name}`
* if tag -> push images with tag `{tag-name}`

.maintain/push-image.sh
weichweich: [1b65854](https://github.com/KILTprotocol/mashnet-node/pull/328/commits/1b65854c84e632b2fedd2b871b97fec90baa135a)

Diiaablo95:
Yes, I see that, and as far as I can see, we won't have a `latest` Docker tag unless we push a `latest` git tag, right? I think GitHub releases can only be associated with one git tag, which would be the runtime version, so I am not sure how we can associated the latest release with the `latest` Docker tag.

weichweich:
> Yes, I see that, and as far as I can see, we won't have a `latest` Docker tag unless we push a `latest` git tag, right? I think GitHub releases can only be associated with one git tag, which would be the runtime version, so I am not sure how we can associated the latest release with the `latest` Docker tag.

Ah now i got you! the only latest thing! :joy: yes i forgot that!

Diiaablo95 (approved):
Grrrrrreat! 😎

weichweich:
let's see if it works. 😓

PR #329, opened 2022-02-24
feat: calculate storage deposits based on storage usage

weichweich:
## Checklist:

### DIFF
--- a/runtimes/common/src/constants.rs
+++ b/runtimes/common/src/constants.rs
@@ -88,18 +88,18 @@ pub const fn deposit(items: u32, bytes: u32) -> Balance {
 }
 
 /// The size of an index in the index pallet
-pub const INDICES_SIZE: u32 = 49;
+pub const MAX_INDICES_SIZE: u32 = 49;
 
 parameter_types! {
 	pub const ByteDeposit: Balance = deposit(0, 1);
-	pub const IndicesDeposit: Balance = deposit(1, INDICES_SIZE);
+	pub const IndicesDeposit: Balance = deposit(1, MAX_INDICES_SIZE);
 }
 
 pub mod attestation {
 	use super::*;
 
-	pub const ATTESTATION_SIZE: u32 = 178;
-	pub const ATTESTATION_DEPOSIT: Balance = deposit(2, ATTESTATION_SIZE);
+	pub const MAX_ATTESTATION_SIZE: u32 = 178;
+	pub const ATTESTATION_DEPOSIT: Balance = deposit(2, MAX_ATTESTATION_SIZE);
 }
 
 pub mod delegation {
@@ -207,9 +207,9 @@ pub mod governance {
 pub mod did {
 	use super::*;
 
-	pub const DID_SIZE: u32 = 7418;
+	pub const MAX_DID_SIZE: u32 = 7418;
 
-	pub const DID_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, DID_SIZE);
+	pub const DID_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, MAX_DID_SIZE);
 	pub const DID_FEE: Balance = 50 * MILLI_KILT;
 	pub const MAX_KEY_AGREEMENT_KEYS: u32 = 10;
 	pub const MAX_URL_LENGTH: u32 = 200;
@@ -232,8 +232,8 @@ pub mod did {
 pub mod did_lookup {
 	use super::*;
 
-	pub const CONNECTION_SIZE: u32 = 80;
-	pub const DID_CONNECTION_DEPOSIT: Balance = deposit(1, CONNECTION_SIZE);
+	pub const MAX_CONNECTION_SIZE: u32 = 80;
+	pub const DID_CONNECTION_DEPOSIT: Balance = deposit(1, MAX_CONNECTION_SIZE);
 }
 
 pub mod treasury {
@@ -265,8 +265,8 @@ pub mod web3_names {
 	pub const MIN_LENGTH: u32 = 3;
 	pub const MAX_LENGTH: u32 = 32;
 
-	pub const NAME_SIZE: u32 = 121;
-	pub const DEPOSIT: Balance = deposit(2, NAME_SIZE);
+	pub const MAX_NAME_SIZE: u32 = 121;
+	pub const DEPOSIT: Balance = deposit(2, MAX_NAME_SIZE);
 }
 
 pub mod preimage {
--- a/runtimes/common/src/constants.rs
+++ b/runtimes/common/src/constants.rs
@@ -88,18 +88,18 @@ pub const fn deposit(items: u32, bytes: u32) -> Balance {
 }
 
 /// The size of an index in the index pallet
-pub const MAX_INDICES_SIZE: u32 = 49;
+pub const MAX_INDICES_BYTE_LENGTH: u32 = 49;
 
 parameter_types! {
 	pub const ByteDeposit: Balance = deposit(0, 1);
-	pub const IndicesDeposit: Balance = deposit(1, MAX_INDICES_SIZE);
+	pub const IndicesDeposit: Balance = deposit(1, MAX_INDICES_BYTE_LENGTH);
 }
 
 pub mod attestation {
 	use super::*;
 
-	pub const MAX_ATTESTATION_SIZE: u32 = 178;
-	pub const ATTESTATION_DEPOSIT: Balance = deposit(2, MAX_ATTESTATION_SIZE);
+	pub const MAX_ATTESTATION_BYTE_LENGTH: u32 = 178;
+	pub const ATTESTATION_DEPOSIT: Balance = deposit(2, MAX_ATTESTATION_BYTE_LENGTH);
 }
 
 pub mod delegation {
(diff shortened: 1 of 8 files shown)


runtimes/common/src/constants.rs
```diff
@@ -78,10 +81,24 @@ pub const INFLATION_CONFIG: (Perquintill, Perquintill, Perquintill, Perquintill)
 /// Copied from Kusama & Polkadot runtime
 pub const MAX_VESTING_SCHEDULES: u32 = 28;
 
+/// Calculate the storage deposit based on the number of storage items and the
+/// combined byte size of those items.
+pub const fn deposit(items: u32, bytes: u32) -> Balance {
+	items as Balance * 70 * MILLI_KILT + (bytes as Balance) * 30 * MICRO_KILT
```
Diiaablo95, line 87: S: I would take `70 * MILLI_KILT` and `30 * MICRO_KILT` out as constants. I think it would also be clearer to rename `bytes` to `total_bytes`, to show that is not the # of bytes per item but the total space taken.

runtimes/common/src/constants.rs
```diff
@@ -78,10 +81,24 @@ pub const INFLATION_CONFIG: (Perquintill, Perquintill, Perquintill, Perquintill)
...
+}
+
+pub const INDICES_SIZE: u32 = 49;
```
Diiaablo95, line 90: What does this represent? Probably a comment would help.

runtimes/common/src/constants.rs
```diff
@@ -194,11 +204,11 @@ pub mod governance {
 }
 
 pub mod did {
-	use crate::BlockNumber;
+	use super::*;
 
-	use super::{Balance, HOURS, KILT, MILLI_KILT};
+	pub const DID_SIZE: u32 = 7418;
```
Diiaablo95, line 209: A comment explaining the result would be helpful, so that we can update this value if we make changes to the DID pallet.

runtimes/common/src/constants.rs
```diff
@@ -194,11 +204,11 @@ pub mod governance {
...
 
-	pub const DID_DEPOSIT: Balance = 2 * KILT;
+	pub const DID_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, DID_SIZE);
```
Diiaablo95, line 211: Where does the 2 come from? Is that details + endpoints count entries?

runtimes/common/src/constants.rs
```diff
@@ -218,27 +228,50 @@ pub mod did {
...
 	pub const INITIAL_PERIOD_REWARD_PER_BLOCK: Balance = YEARLY_REWARD / (BLOCKS_PER_YEAR as Balance);
 }
 
 pub mod proxy {
-	pub const MAX_PROXIES: u16 = 10;
-	pub const MAX_PENDING: u16 = 10;
+	use super::*;
+
+	parameter_types! {
+		// One storage item; key size 32, value size 8; .
```
Diiaablo95, line 250: Do we need to specify the key size?

runtimes/common/src/constants.rs
```diff
@@ -218,27 +228,50 @@ pub mod did {
...
-	use crate::Balance;
-
-	use super::KILT;
+	use super::*;
 
 	pub const MIN_LENGTH: u32 = 3;
 	pub const MAX_LENGTH: u32 = 32;
-	pub const DEPOSIT: Balance = 2 * KILT;
+
+	pub const NAME_SIZE: u32 = 121;
```
Diiaablo95, line 267: I would call it `NAME_MAX_SIZE`.

runtimes/peregrine/src/tests.rs
```diff
@@ -30,6 +38,58 @@ fn call_size() {
 	);
 }
 
+#[test]
+fn attestation_storage_sizes() {
+	let attestation_record = attestation::AttestationDetails::<Runtime>::max_encoded_len();
+	let delegation_record = BoundedVec::<
+		<Runtime as frame_system::Config>::Hash,
```
Diiaablo95, line 45: S: These values should be probably split up into different variables first. The same for the rest below and in the other runtimes.

Diiaablo95 (commented):
Not sure how final this PR is, but I would like to see more comments for all the magic numbers and more descriptive names. I would take a second look once it becomes easier to follow 👌

runtimes/common/src/constants.rs
weichweich: I would not. I would need to add MAX everywhere.

runtimes/common/src/constants.rs
weichweich: Is it bothering you? The comment is taken from parity. Together with the values for the deposit function. I added them because they are nice to have and come with no cost.

runtimes/common/src/constants.rs
weichweich: yes

runtimes/common/src/constants.rs
weichweich: [dbd748e](https://github.com/KILTprotocol/mashnet-node/pull/329/commits/dbd748e325eedceee15e2ccf42899dc8b7d95422)

runtimes/common/src/constants.rs
weichweich: I wouldn't split this up any further. The numbers are only used here. If we should use them somewhere else I would add a constant. But since this is the only use i would like to keep it simple.
You can see the whole function with all it's values. You see `items` and `bytes` are the variables and know what the outcome is.

runtimes/peregrine/src/tests.rs
weichweich: Having a type for it might be nice but I would not split this up.

runtimes/common/src/constants.rs
Diiaablo95: I understand, it just seems inconsistent to me since we don't mention key sizes anywhere else, and I am not sure whether it affects anything or not.

runtimes/common/src/constants.rs
Diiaablo95: But that's what it is, isn't it?

weichweich: Should I also add bytes since it's bytes?

weichweich: [c5a8177](https://github.com/KILTprotocol/mashnet-node/pull/329/commits/c5a8177fc4e98b017c9c87028d7db028aaafc1bc)

weichweich: I changed `size` to `byte_length` since that terminology is also used somewhere else. [153e738](https://github.com/KILTprotocol/mashnet-node/pull/329/commits/153e7386734e4a61a3e830df20e3278510d7cc70)

runtimes/peregrine/src/tests.rs
```diff
@@ -30,6 +39,61 @@ fn call_size() {
...
+fn web3_name_storage_sizes() {
+	let owner_size = Web3NameOf::<Runtime>::max_encoded_len();
+	let name_size = Web3OwnershipOf::<Runtime>::max_encoded_len();
+
+	assert_eq!(owner_size + name_size, MAX_NAME_BYTE_LENGTH as usize)
+}
+type Indices = (<Runtime as frame_system::Config>::AccountId, BalanceOf<Runtime>, bool);
+
+#[test]
+fn indices_storage_sizes() {
```
wischli, line 92: S: Move this into the test scope as the type is not needed elsewhere.
```suggestion

#[test]
fn indices_storage_sizes() {
type Indices = (<Runtime as frame_system::Config>::AccountId, BalanceOf<Runtime>, bool);
```

runtimes/spiritnet/src/tests.rs
```diff
@@ -30,6 +39,61 @@ fn call_size() {
...
+}
+
+#[test]
+fn web3_name_storage_sizes() {
+	let owner_size = Web3NameOf::<Runtime>::max_encoded_len();
+	let name_size = Web3OwnershipOf::<Runtime>::max_encoded_len();
+
+	assert_eq!(owner_size + name_size, MAX_NAME_BYTE_LENGTH as usize)
+}
+type Indices = (<Runtime as frame_system::Config>::AccountId, BalanceOf<Runtime>, bool);
```
wischli, line 89: Same, see above.

runtimes/common/src/constants.rs
```diff
@@ -194,11 +205,11 @@ pub mod governance {
 }
 
 pub mod did {
-	use crate::BlockNumber;
+	use super::*;
 
-	use super::{Balance, HOURS, KILT, MILLI_KILT};
+	pub const MAX_DID_BYTE_LENGTH: u32 = 7418;
```
wischli, line 210: C: Add comments for these "magic numbers" explaining that these numbers were actually derived in the runtime unit tests.

wischli (commented):
I have two minor remarks but overall LGTM! Would also love to have comments for all the magic storage numbers "redirecting" to the runtime unit test.

runtimes/common/src/constants.rs
weichweich: [bb39a1b](https://github.com/KILTprotocol/mashnet-node/pull/329/commits/bb39a1b29f6489dba3b51f5b95c6c9efdd7fc5b7)

runtimes/spiritnet/src/tests.rs
weichweich: 3ac2a64741db0b64ef7ce77fd1fdebf56d3a4147

runtimes/peregrine/src/tests.rs
weichweich: 3ac2a64741db0b64ef7ce77fd1fdebf56d3a4147

wischli (approved):
LGTM!

PR #331, opened 2022-02-28
feat: enable fat link time optimization for release builds

trusch:
## fixes KILTProtocol/ticket#1855

make it faster

## How to test:

* check benchmarks

## Checklist:


weichweich:
looks good we need to run every benchmark again. I would also like to check what the SR-Tool does and if that has influence on the benchmarks.

weichweich:
stale

PR #336, opened 2022-03-10
ci: use custom ci image

weichweich:
skipping one step should speed up the build

[The Dockerfile can be found in our deployment repository](https://github.com/KILTprotocol/kubernetes/blob/master/container/ci.Dockerfile)


ggera:
LGTM

PR #337, opened 2022-03-11
Optimizes docker layer

ggera:
Optimizes kilt-node docker layers: Check polkadot builder for comparison [https://github.com/paritytech/polkadot/blob/master/scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile](https://github.com/paritytech/polkadot/blob/master/scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile)

## How to test:
Build a docker image and test if number of layers is lowwered and build time is imrpoved

## Checklist:


tjwelde:
I was seeing that they changed their Dockerfiles, too so I agree with aligning. @weichweich will have to approve though, since he can judge the changes better here.

weichweich (approved):

Dockerfile
```diff
@@ -8,12 +8,7 @@ WORKDIR /build
 
 ARG FEATURES=default
 
-COPY ./nodes /build/nodes
-COPY ./pallets /build/pallets
-COPY ./runtimes /build/runtimes
-COPY ./support /build/support
-COPY ./Cargo.lock /build/Cargo.lock
-COPY ./Cargo.toml /build/Cargo.toml
```
nazar-pc, line 16: Neither new nor old are great. If there is any layer caching at all, `Cargo.toml` and `Cargo.lock` should be copied first, dependencies installed (those don't change often) and then the rest of the code should be copied and the project itself compiled. This way code changes don't result in rebuilding of everything.

.dockerignore
```diff
@@ -2,6 +2,8 @@
 .github
 .gitignore
 .gitlab-ci.yml
+.rustfmt.toml
```
nazar-pc, line 5: Ignoring everything and the  un-ignoring things that are actually needed for build is typically a more productive approach than block-listing

Dockerfile
weichweich: Thanks for the tip! We had dependency caching in our Dockerfile a while ago but experienced some issues. In the end there where some compile errors because of the dependency caching. This and the not so huge improvement by the cache lead us to remove the caching again. Last time we where hoping that [cargo-wharf](https://github.com/denzp/cargo-wharf) would solve the caching, but it looks like no development was done in 2y.

PR #341, opened 2022-03-16
feat: proxy type for disableling deposit claiming

weichweich:
## No Ticket

Add a proxy type that allows paying for DIDs and everything related, but not being able to delete all these entries.
This can be useful if someone wants to pay the storage deposits for others, but doesn't want to have the power to delete the dids, attestations etc. of the other.

[m](https://kilt-cast.monday.com/boards/1792114080/pulses/2435178461)

## Checklist:


runtimes/peregrine/src/lib.rs
```diff
@@ -772,47 +780,132 @@ impl InstanceFilter<Call> for ProxyType {
 			ProxyType::Any => true,
 			ProxyType::NonTransfer => matches!(
 				c,
-				Call::System(..) |
```
ntn-x2, line 775: I though it would be best to make all allowed calls explicit, and everything else is denied by default. The reverse could cause unauthorised calls to be allowed, while in this way potentially authorised calls might be denied, which is a safer approach.

runtimes/peregrine/src/lib.rs
```diff
@@ -772,47 +780,132 @@ impl InstanceFilter<Call> for ProxyType {
...
+			ProxyType::NonDepositClaiming => matches!(
+				c,
+				Call::Attestation(
+						// Excludes `reclaim_deposit`
+						attestation::Call::add { .. }
+							| attestation::Call::remove { .. }
+							| attestation::Call::revoke { .. }
+					)
+					| Call::Authorship(..)
+					// Excludes `Balances`
```
wischli, line 827: Do we really want to exclude Balances here?

runtimes/peregrine/src/lib.rs
wischli: I would assume the same.

wischli (commented):
One question and I agree with Antonio.

runtimes/peregrine/src/lib.rs
weichweich: I would say we should not exclude balance transfers here. All calls should be allowed.

PR #345, opened 2022-03-25
chore: Updates toolchain version

ggera:
## NO TICKET
updated the toolchain init script so that the benchbot and ci can call it


ntn-x2 (commented):
Do we also need to fix the CI Docker images that we use? I made some tests this morning and it was using the old toolchain.

ggera:
> Do we also need to fix the CI Docker images that we use? I made some tests this morning and it was using the old toolchain.

The ci seems to override the docker image toolchain by picking settings in the toolchain file

ntn-x2 (approved):
All good then 👍

ntn-x2:
I just saw that our release pipeline uses srtool v1.57, when we should probably use v1.59 as our toolchain specifies @ggera

ggera:
> 

That is correct, the [srtool rust version seems hard coded ](https://hub.docker.com/layers/srtool/paritytech/srtool/1.57.0/images/sha256-912c690ddf167a4b897a16d120c0af70159b873102184289d135434ecead5bdd?context=explore) and not sure if it overridable.

PR #351, opened 2022-04-06
refactor: DID authorisation logic

ntn-x2:
This PR fixes https://github.com/KILTprotocol/ticket/issues/1897.

It is based on top of https://github.com/KILTprotocol/mashnet-node/pull/350, as they are meant to be considered two parts of the same solution.

## What it does

In short, this PR allows DID creation calls to use a `DidOrigin` instead of a `SignedOrigin`, meaning that it will go through `submit_did_call` and, when batched with other DID management extrinsics, will not require an additional signature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8874,6 +8874,7 @@ name = "runtime-common"
 version = "1.7.0"
 dependencies = [
  "attestation",
+ "did",
  "frame-support",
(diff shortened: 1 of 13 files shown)


weichweich:
IMO this will also be solved differently by https://github.com/KILTprotocol/kilt-node/pull/551

PR #353, opened 2022-04-07
fix: features

weichweich:
## Checklist:


scripts/all.fish
```diff
@@ -0,0 +1,11 @@
+for features in "--all-features" "--features runtime-benchmarks" "--features try-runtime"
```
wischli, line 1: When I try to execute this, I receive an `error: no such subcommand: `workspaces`.

wischli (approved):
LGTM! However, I cannot execute the fish script.

runtimes/common/Cargo.toml
```diff
@@ -30,6 +30,9 @@ sp-std = {git = "https://github.com/paritytech/substrate", default-features = fa
 default = ["std"]
 fast-gov = []
 runtime-benchmarks = [
+  "attestation/runtime-benchmarks",
+  "frame-support/runtime-benchmarks",
```
ntn-x2, line 34: Why does runtime common enable these features upon runtime benchmarks? I'm talking specifically about `attestation` and `parachain-staking`.

scripts/all.fish
ntn-x2, line 1: Would it also make sense to test with no (default) features, and with the mock feature, where specified? Anyway very good solution to the problem!

ntn-x2 (commented):
Just two small questions, but otherwise LGTM!

scripts/all.fish
weichweich: You have to install `cargo-workspaces`. I will add a comment for that.

weichweich: yes that's a good idea

weichweich: default features are a good idea. But mock doesn'T work since not all of the crates have that feature

PR #354, opened 2022-04-08
fix: account linking signature verification

ntn-x2:
Wraps the expected payload with `<Bytes>...</Bytes>` the way [PolkadotJS extension does it](https://github.com/polkadot-js/extension/blob/607f4b3e3b045020659587771fd3eba7b3214862/packages/extension-base/src/background/RequestBytesSign.ts#L24=).


pallets/pallet-web3-names/src/web3_name.rs
```diff
@@ -16,8 +16,6 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
-#![cfg_attr(not(feature = "std"), no_std)]
```
ntn-x2, line 19: Not sure where this was coming from, but Clippy was warning me about it.

pallets/pallet-did-lookup/src/signature.rs
```diff
@@ -0,0 +1,32 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use sp_std::vec::Vec;
+
+// According to https://github.com/polkadot-js/common/blob/master/packages/util/src/u8a/wrap.ts#L13=
```
weichweich, line 21: Could we use a permalink so that it will still point to the same thing next year?

```suggestion
// According to https://github.com/polkadot-js/common/blob/5d5c7e4c0ace06e3301ccadfd3c3351955f1e251/packages/util/src/u8a/wrap.ts#L13
```

weichweich (approved):

pallets/pallet-did-lookup/src/signature.rs
ntn-x2: https://github.com/KILTprotocol/mashnet-node/pull/354/commits/30e04faa724897e99152b39f037a0172fdbe4dd7

ntn-x2:
Waiting for @tjwelde to test the new runtime with the linking website before going on with the merge.

wischli (approved):
LGTM

ntn-x2:
@wischli @weichweich please take another quick look, then we can proceed if everything's fine. I just bumped the crates version.

wischli (approved):
LGMT pt2

PR #348, opened 2022-04-01
feat: did rpc

weichweich:
## fixes [#2016](https://github.com/KILTprotocol/ticket/issues/2016)

* add rpc endpoints to query DID information

todo:

- [x] `RPC methods not decorated`
- [x] service endpoints
- [x] query by did

### query using did

```
❯ curl -s -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "did_query", "params":["5FKxrLQM24ZhLxcaQfJR3uMxMZh5gU6E4CP3yghPJLzCDnHN"]}' http://127.0.0.1:19933/ | jq
```

<details>
<summary>Response</summary>

```
{
  "jsonrpc": "2.0",
  "result": {
    "accounts": [
      "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
    ],
    "details": {
      "attestation_key": null,
      "authentication_key": "0x15ba41bf8842ae185ae06110a63e6cc7ac36d8a8ea217746231bef46c1379ab1",
      "delegation_key": null,
      "deposit": {
        "amount": "2007900000000000",
        "owner": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
      },
      "key_agreement_keys": [],
      "last_tx_counter": 3,
      "public_keys": {
        "0x15ba41bf8842ae185ae06110a63e6cc7ac36d8a8ea217746231bef46c1379ab1": {
          "block_number": 120,
          "key": {
            "PublicVerificationKey": {
              "Sr25519": "5FKxrLQM24ZhLxcaQfJR3uMxMZh5gU6E4CP3yghPJLzCDnHN"
            }
          }
        }
      }
    },
    "identifier": "5FKxrLQM24ZhLxcaQfJR3uMxMZh5gU6E4CP3yghPJLzCDnHN",
    "service_endpoints": [
      {
        "id": "123",
        "service_types": [
          "twitter"
        ],
        "urls": [
          "example.com"
        ]
      }
    ],
    "w3n": "alice-0"
  },
  "id": 1
}
```
</details>

### query using w3n

```
❯ curl -s -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "did_queryByWeb3Name", "params":["alice-0"]}' http://127.0.0.1:19933/ | jq
```

<details>
<summary>Response</summary>

```
{
  "jsonrpc": "2.0",
  "result": {
[...]

PR #355, opened 2022-04-08
feat: support linking ethereum accounts

trusch:
## fixes [KILTProtocol/ticket#1936](https://github.com/KILTprotocol/ticket/issues/1936)
This adds support to link ethereum accounts to a DID. It used moonbeams AccountId20 implementation and related Signer implementations etc. 

Update: Its working now :) Following the instructions below should result in linking your eth account to a DID

## How to test:

Its a bit harder to test, but here are the instructions:
* spin up a local node with from this PR
* install [kiltctl](https://github.com/trusch/kiltctl) from the branch `did-submit`
* create a DID for the alice account
```bash
#!/bin/bash

# for gpg to ask for passphrases within the command line
export GPG_TTY=$(tty)

# endpoint of a freshly started dev node. Please adjust the port to your needs.
export KILTCTL_ENDPOINT=ws://localhost:9948

# path to the key storage for this test
export KILTCTL_STORAGE=./kilt-test.db

# create sr25519 keypair for alice
echo "Create key 'alice'"
kiltctl keys create --seed "//Alice" --name alice

# create account used for payments from the keys
echo "Create account 'alice'"
kiltctl account create --key alice --name alice

# re-use the key and create a DID for alice using that key as auth key
echo "Create DID 'alice'"
kiltctl did create --auth alice --name alice

# register the DID
echo "Register DID 'alice'"
kiltctl did register --payment alice alice

# sanity check -> should resolve the DID document from chain
echo "Resolve DID 'alice'"
kiltctl did resolve alice
```
* Now the `did:kilt:4siJtc4dYq2gPre8Xj6KJcSjVAdi1gmjctUzjf3AwrtNnhvy` should exist
* next steps are for creating the ethereum signate
* `git clone git@github.com:KILTprotocol/minimal-eth-linking-signer.git` ([repo](https://github.com/KILTprotocol/minimal-eth-linking-signer/))
* go to the directory and `yarn install && yarn start`
  * this will spin up the signer application that talks to metamask
* at first you will only see a connect button, click it
* now enter all info that we need (did identifier + expiration) and click "sign"
* in the end it should look like this:
image
* Lets create the actual test call
* go to polkadot apps (pointing to your local node) -> extrinsics -> didLookup -> associateAccount
* Fill out the call and in the end it should look like this: 
image
* copy `encoded call data` (without the leading `0x`!) which is your scale encoded extrinsic 
* go back to your terminal to submit this call signed by the alice DID
* ` KILTCTL_STORAGE=./kilt-test.db kiltctl did sign_and_submit --data "$DATA" --payment alice alice`
* After this is finalized, check the chain state and events if everything worked

## Checklist:


pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,229 @@
+// Copyright 2019-2022 PureStake Inc.
```
trusch, line 1: I had to copy this in because in the moonbeam repo they are depending on their fork of substrate.

pallets/pallet-did-lookup/src/linkable_account.rs
```diff
@@ -0,0 +1,110 @@
...
+	fn from(account_id: AccountId32) -> Self {
+		Self::AccountId32(account_id)
+	}
+}
+
+impl signature::GetWrapType for LinkableAccountId {
+	fn get_wrap_type(&self) -> signature::WrapType {
+		match self {
+			Self::AccountId20(_) => signature::WrapType::Dotsama,
+			Self::AccountId32(_) => signature::WrapType::Ethereum,
```
wischli, line 33: Incorrect matching 😉 I would also rename Dotsama to Substrate
```suggestion
			Self::AccountId20(_) => signature::WrapType::Ethereum,
			Self::AccountId32(_) => signature::WrapType::Dotsama,
```

wischli (dismissed):
Can we use `Substrate` instead of `Dotsama`?

pallets/pallet-did-lookup/src/linkable_account.rs
trusch: Thanks :) fixed it.

pallets/pallet-did-lookup/src/account.rs
trusch: @GustavHemmelmayr I would really appreciate your input here!

trusch:
regarding the needed migration: I see about 1.3k AssociationEstablished events on spiritnet, so I guess that just writing a migration and migrating all the storage at once will not work. What would be the best way to deal with that? @wischli

trusch:
Green CI :) Btw. we have stricter linter settings than moonbeam! They had implemented some Into's for their AccountId20 type and clippy complained it wants From's instead :man_shrugging:

wischli:
> regarding the needed migration: I see about 1.3k AssociationEstablished events on spiritnet, so I guess that just writing a migration and migrating all the storage at once will not work. What would be the best way to deal with that? @wischli

Theoretically, it should easily fit into one block as the migration basically does `1300` reads and writes which amounts to `1300 * (100,000,000 + 25,000,000) = 162,500,000,000` weight which consumes 44% of the available block weight (370,000,000,000 - 125,000,000). In other words, theoretically we could fit ~2900 reads and writes into a single block.

However, I would stretch this into multiple blocks as this will be part of a runtime upgrade. When you can fit such migrations into 2-3 blocks, [the recommendation](https://docs.substrate.io/how-to-guides/v3/parachains/runtime-upgrades/) is to use the Scheduler pallet.

trusch:
@wischli albi suggested that we could also do it lazily by having two storage entries and that we could always first read and write to the new storage item and as backup read old entries from the old storage and convert on the fly. What do you think? I guess in the scheduler approach we would also need the two storage entries to coexist.

wischli:
> @wischli albi suggested that we could also do it lazily by having two storage entries and that we could always first read and write to the new storage item and as backup read old entries from the old storage and convert on the fly. What do you think? I guess in the scheduler approach we would also need the two storage entries to coexist.

We could also go that route but I fear this will make the code less readable and we have to keep that state until we do a force migration anyway. We cannot guarantee users to perform actions which would migrate on the fly.

If @weichweich thinks we should go that way, I am fine with it.

trusch:
@wischli Could you have a look? I left the old storage simply untouched for now and only operate on new storage items ConnectedAccountsV2 and ConnectedDidsV2. I also provide a extrinsic to migrate all accounts associated with the sender did. Do you think this is doable like this?

wischli (commented):
> @wischli Could you have a look? I left the old storage simply untouched for now and only operate on new storage items ConnectedAccountsV2 and ConnectedDidsV2. I also provide a extrinsic to migrate all accounts associated with the sender did. Do you think this is doable like this?

I am not happy with the proposed approach as I see two major issues which have to be improved:
1. If a user does not migrate, it's impossible for them and anyone else to remove their connected DID and/or association.
2. A user can have storage entries in v1 and v2. If they don't migrate, both versions co-exist. a. There is no incentive for a user to migrate as they have to pay the transaction fee and don't get any benefit. We should not require users to actively migrate. They should not have to worry/know about this at all. 
b. Moreover, any frontend that uses our API would have to query twice as many records for potential DID connections. 
c. This also breaks/is not aligned with #348. Of course, we could refactor #348 such that we handle the fallback from v2 to v1 and as long as services use our optimized RPC, they are fine. However, we still have to do redundant checks on our side then.

I can imagine that Albi's lazy migration rather wanted to migrate on the fly when adding a new association or removing an existing association.

All in all, I am not in favor of migrating lazily. We definitely increase technical debt and create problems/migration paths for the SDK and Sporran, at least. There are less than 1400 accounts which we need to migrate. I don't see any issue in going forward with a (scheduled) migration which does not create any secondary problems. Let's ask Shawn to be sure and don't implement any half solution.

pallets/pallet-did-lookup/src/signature.rs
```diff
@@ -16,18 +16,34 @@
...
+	match account.get_wrap_type() {
+		WrapType::Substrate => PAYLOAD_BYTES_WRAPPER_PREFIX
+			.iter()
+			.chain(payload.iter())
+			.chain(PAYLOAD_BYTES_WRAPPER_POSTFIX.iter())
+			.copied()
+			.collect(),
+		WrapType::Ethereum => ETHEREUM_SIGNATURE_PREFIX
+			.iter()
+			.chain(payload.len().to_string().as_bytes().iter())
```
ntn-x2, line 44: Can you add a comment that the length of the payload is also added to the message to sign?

pallets/pallet-did-lookup/src/linkable_account.rs
```diff
@@ -0,0 +1,134 @@
...
+		match self {
+			Self::AccountId20(account_id) => write!(f, "{}", account_id),
+			Self::AccountId32(account_id) => write!(f, "{}", account_id),
+		}
+	}
+}
+
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Eq, PartialEq, Clone, Encode, Decode, MaxEncodedLen, RuntimeDebug, TypeInfo)]
+pub enum LinkableAccountSignature {
```
ntn-x2, line 80: Do we need this? At the end, an Ethereum signature is an Ecdsa signature, which is included in the `MultiSignature` struct. What is the argument in favor of having a new signature type here?

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -51,15 +54,19 @@ pub mod pallet {
...
 
 	pub use crate::connection_record::ConnectionRecord;
 
 	use crate::signature::get_wrapped_payload;
 
-	/// The identifier to which the accounts can be associated.
+	/// The native identifier for accounts in this runtime.
 	pub(crate) type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 
+	/// The identifier of the accounts that can be associated to a did.
```
ntn-x2, line 67:
```suggest
	/// The identifier of the accounts that can be associated to a DID.
```

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -282,17 +323,46 @@ pub mod pallet {
...
+
+			Self::deposit_event(Event::AssociationsMigrated(source.subject()));
+
+			Ok(())
+		}
 	}
 
+
 	impl<T: Config> Pallet<T> {
 		pub(crate) fn add_association(
```
ntn-x2, line 362: What happens if a new account is added to the new storage before the old one is migrated? It would create a situation in which the migration might fail. So we will probably ignore existing keys when migrating over to account for this case, but we should put some comment here.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -306,22 +376,22 @@ pub mod pallet {
...
+			ConnectedAccountsV2::<T>::insert(&did_identifier, &account, ());
 			Self::deposit_event(Event::AssociationEstablished(account, did_identifier));
 
 			Ok(())
 		}
 
-		pub(crate) fn remove_association(account: AccountIdOf<T>) -> DispatchResult {
-			if let Some(connection) = ConnectedDids::<T>::take(&account) {
-				ConnectedAccounts::<T>::remove(&connection.did, &account);
+		pub(crate) fn remove_association(account: LinkableAccountIdOf<T>) -> DispatchResult {
```
ntn-x2, line 392: Similar as above: in this case it might be that a removal will fail if the account has not been migrated yet. It probably deserves a comment until the next runtime.

pallets/pallet-did-lookup/src/linkable_account.rs
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+impl From<EthereumSignature> for LinkableAccountSignature {
+	fn from(signature: EthereumSignature) -> Self {
+		Self::EthereumSignature(signature)
+	}
+}
+
+impl From<sp_core::sr25519::Signature> for LinkableAccountSignature {
```
ntn-x2, line 97: I guess this was only used for the benchmarks, right? In that case, it should be behind the respective feature gate.

pallets/pallet-did-lookup/src/benchmarking.rs
```diff
@@ -57,70 +57,79 @@ benchmarks! {
...
 		make_free_for_did::<T>(&caller);
 
 		// Add existing connected_acc -> previous_did connection that will be replaced
-		Pallet::<T>::add_association(caller.clone(), previous_did.clone(), connected_acc_id.clone()).expect("should create previous association");
-		assert!(ConnectedAccounts::<T>::get(&previous_did, T::AccountId::from(connected_acc)).is_some());
+		Pallet::<T>::add_association(caller.clone(), previous_did.clone(), linkable_id.clone()).expect("should create previous association");
+		assert!(ConnectedAccounts::<T>::get(&previous_did, linkable_id.clone()).is_some());
 		let origin = T::EnsureOrigin::generate_origin(caller, did.clone());
-	}: _<T::Origin>(origin, connected_acc_id, bn, sig)
+	}: _<T::Origin>(origin, linkable_id.clone(), bn, sig)
```
ntn-x2, line 79: No clones in the benchmark itself. Please clone the ID before.

pallets/pallet-did-lookup/Cargo.toml
```diff
@@ -21,6 +21,11 @@ sp-keystore = {branch = "polkadot-v0.9.19", git = "https://github.com/paritytech
 codec = {package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
 log = "0.4.16"
 scale-info = {version = "2.1.1", default-features = false, features = ["derive"]}
+sha3 = { version = "0.9", default-features = false }
```
ntn-x2, line 24: Can you sort dependencies in each section in alphabetical order?

pallets/pallet-did-lookup/Cargo.toml
```diff
@@ -55,6 +60,7 @@ std = [
   "scale-info/std",
   "sp-runtime/std",
   "sp-std/std",
+	"impl-serde/std",
```
ntn-x2, line 63: Don't we need to add other crates with `std` feature as well?

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,216 @@
...
+		EthereumSignature(x)
+	}
+}
+
+impl sp_runtime::traits::Verify for EthereumSignature {
+	type Signer = EthereumSigner;
+	fn verify<L: sp_runtime::traits::Lazy<[u8]>>(&self, mut msg: L, signer: &AccountId20) -> bool {
+		let mut m = [0u8; 32];
+		m.copy_from_slice(Keccak256::digest(msg.get()).as_slice());
+		match sp_io::crypto::secp256k1_ecdsa_recover(self.0.as_ref(), &m) {
```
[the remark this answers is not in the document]
ntn-x2: I think in that way, the account ID would just be the first 20 bytes of the public key hash.

ntn-x2 (commented):
Very nice solution. I agree with William that we should probably use the schedule migration approach. Moreover, I would REALLY like to see some more comments and some clearer variable names, perhaps even splitting up operations into multiple steps to enhance readability, if possible.
Last but not least, I would love to see some test cases for the pallet that test with linking a new Ethereum account, as I don't think there's any right now.
I would then like to give another review once at least the comments and the test cases have been added.

pallets/pallet-did-lookup/src/linkable_account.rs
trusch: The pubkey format for the substrate-ecdsa and the ethereum-ecdsa signatures are different, so its unfortunately not the same

ntn-x2: Then I would love to see the difference in some tests 😁 It is possible to go from one key to the other, if needed, but I am eager to know how that maps with actual values.

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,216 @@
...
+				false
+			}
+		}
+	}
+}
+
+/// Public key for an Ethereum / Moonbeam compatible account
+#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct EthereumSigner([u8; 20]);
```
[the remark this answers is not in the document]
trusch: done

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,216 @@
...
+	use super::*;
+	use sp_core::{ecdsa, Pair};
+	use sp_runtime::traits::IdentifyAccount;
+
+	#[test]
+	fn test_account_derivation_1() {
+		// Test from https://asecuritysite.com/encryption/ethadd
+		let secret_key = hex::decode("502f97299c472b88754accd412b7c9a6062ef3186fba0c0388365e1edec24875").unwrap();
+		let mut expected_hex_account = [0u8; 20];
+		hex::decode_to_slice("976f8456e4e2034179b284a23c0e0c8f6d3da50c", &mut expected_hex_account)
```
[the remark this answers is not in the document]
trusch: done

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,216 @@
...
+		let mut m = [0u8; 32];
+		m.copy_from_slice(Keccak256::digest(msg.get()).as_slice());
+		match sp_io::crypto::secp256k1_ecdsa_recover(self.0.as_ref(), &m) {
+			Ok(pubkey) => {
+				// TODO This conversion could use a comment. Why H256 first, then H160?
+				// TODO actually, there is probably just a better way to go from Keccak digest.
+				AccountId20(H160::from(H256::from_slice(Keccak256::digest(&pubkey).as_slice())).0) == *signer
+			}
+			Err(sp_io::EcdsaVerifyError::BadRS) => {
+				log::error!(target: "evm", "Error recovering: Incorrect value of R or S");
```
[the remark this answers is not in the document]
ntn-x2: Also applies to the other two cases of the switch.

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,219 @@
...
+	}
+}
+
+impl From<[u8; 20]> for EthereumSigner {
+	fn from(x: [u8; 20]) -> Self {
+		EthereumSigner(x)
+	}
+}
+
+impl From<ecdsa::Public> for EthereumSigner {
```
ntn-x2, line 119: This code might not be needed anymore since 0.9.22 contains https://github.com/paritytech/substrate/pull/11087 which exposes a `to_eth_address` for `frame_support::ecdsa::Public`. Might be worth checking it out once William's PR is in 😊

pallets/pallet-did-lookup/Cargo.toml
trusch: done

runtimes/standalone/src/lib.rs
```diff
@@ -916,10 +912,12 @@ impl_runtime_apis! {
 					did::Did::<Runtime>::get(&owner_info.owner).map(|details| (owner_info, details))
 				})
 				.map(|(owner_info, details)| {
-					let accounts = pallet_did_lookup::ConnectedAccounts::<Runtime>::iter_key_prefix(&owner_info.owner).collect();
+					let accounts: Vec<LinkableAccountId> = pallet_did_lookup::ConnectedAccounts::<Runtime>::iter_key_prefix(
+						&owner_info.owner,
+					).collect();
```
weichweich, line 917: Why is the type hint necessary here?

pallets/pallet-did-lookup/Cargo.toml
weichweich: I think we need atleast `sp-io` and `sp-core` here

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,213 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::{crypto::ecdsa::ECDSAExt, RuntimeDebug};
+use scale_info::TypeInfo;
+use sha3::{Digest, Keccak256};
+use sp_core::{ecdsa, H160, H256};
+
+#[cfg(feature = "std")]
+pub use sp_runtime::serde::{de::DeserializeOwned, Deserialize, Serialize};
```
weichweich, line 34: can you use serde directly here.

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,213 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+// This code originally came from the purestake/moonbeam repo.
```
weichweich, line 19: can we add a link to the source?

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,213 @@
...
+use scale_info::TypeInfo;
+use sha3::{Digest, Keccak256};
+use sp_core::{ecdsa, H160, H256};
+
+#[cfg(feature = "std")]
+pub use sp_runtime::serde::{de::DeserializeOwned, Deserialize, Serialize};
+
+//TODO Maybe this should be upstreamed into Frontier (And renamed accordingly)
+// so that it can be used in palletEVM as well. It may also need more traits
+// such as AsRef, AsMut, etc like AccountId32 has.
```
weichweich, line 38: Seems like purestake wasn't the original author?

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// The account type to be used in Moonbeam. It is a wrapper for 20 fixed bytes.
+/// We prefer to use a dedicated type to prevent using arbitrary 20 byte arrays
+/// were AccountIds are expected. With the introduction of the `scale-info`
+/// crate this benefit extends even to non-Rust tools like Polkadot JS.
+
```
weichweich, line 44: Can we update this comment. It's not used in moonbeam anymore. 😉

pallets/pallet-did-lookup/src/account.rs
```diff
@@ -0,0 +1,213 @@
...
+impl From<[u8; 20]> for EthereumSigner {
+	fn from(x: [u8; 20]) -> Self {
+		EthereumSigner(x)
+	}
+}
+
+impl From<ecdsa::Public> for EthereumSigner {
+	fn from(x: ecdsa::Public) -> Self {
+		Self(x.to_eth_address().unwrap())
```
weichweich, line 121: Why is it safe to unwrap here?

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -24,10 +24,12 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+pub mod account;
+mod connection_record;
+
 pub mod default_weights;
 pub mod migrations;
```
weichweich, line 31: I wouldn't mix `pub mod` and `mod`. The style we use is to group those statements by `pub mod`, `mod`, mods with features, etc.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -39,33 +41,45 @@ mod mock;
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
 
+pub mod linkable_account;
```
weichweich, line 44: definitly move this `pub mod` up to the other pub mods

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -39,33 +41,45 @@ mod mock;
...
+
+use codec::{Decode, Encode, MaxEncodedLen};
+use scale_info::TypeInfo;
+
 pub use crate::{default_weights::WeightInfo, pallet::*};
 
+use crate::account::{AccountId20, EthereumSignature};
+use sp_runtime::{AccountId32, MultiSignature};
```
weichweich, line 52: This use statements should be grouped with the other non pub uses

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -197,7 +209,22 @@ pub mod pallet {
...
+					Self::add_association(sender, did_identifier, acc.into())?;
+				}
+				AssociateAccountRequest::Ethereum(acc, proof) => {
+					ensure!(
+						proof.verify(&get_wrapped_payload(&encoded_payload[..], &acc)[..], &acc),
+						Error::<T>::NotAuthorized
+					);
+					Self::add_association(sender, did_identifier, acc.into())?;
+				}
+			};
```
weichweich, line 227: can we add this as a verify method to AssociateAccountRequest?

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
+			// the end. Else we iterate over every key more or less twice.
+			ConnectedDids::<T>::drain().for_each(|(acc_id32, value)| {
+				log::debug!(
+					"🔎 #{:?} Migrating ConnectedDid for account id {:?}",
+					connected_dids.encode(),
+					acc_id32.encode()
+				);
+				let acc_id: LinkableAccountId = acc_id32.into();
+				TmpConnectedDids::<T>::insert(acc_id, value);
+				connected_dids += 1;
```
weichweich, line 91: Would be best to use `saturating_*` everywhere

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
+				);
+				let acc_id: LinkableAccountId = acc_id32.into();
+				TmpConnectedAccounts::<T>::insert(did_id, acc_id, val);
+				connected_accounts += 1;
 			});
+			log::info!("🔎 DidLookup: Migrated all {:?} ConnectedAccounts", connected_accounts);
+
+			// Move TmpStorage
+			move_storage::<Pallet<T>>(b"TmpConnectedDids", b"ConnectedDids");
+			move_storage::<Pallet<T>>(b"TmpConnectedAccounts", b"ConnectedAccounts");
```
weichweich, line 111: So we read everything from ConnectedDids to TmpConnectedDids and write TmpConnectedDids back to TmpConnectedDids?

I would rather propose to translate the storage inplace. THis would cost half as many storage accesses.

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
 
-		Pallet::<T>::current_storage_version().put::<Pallet<T>>();
+			StorageVersion::new(3).put::<Pallet<T>>();
 
+			<T as frame_system::Config>::DbWeight::get().reads_writes(
+				(connected_dids.saturating_add(connected_accounts)).saturating_mul(2),
+				(connected_dids.saturating_add(connected_accounts))
+					.saturating_mul(2)
```
weichweich, line 118: AFAIK delete is also a storage write, so this would be a 4?

weichweich (changes requested):
Very nice PR! 🙂
I have a few minor things and a question regarding the migration. Didn't we want to do the migration lazily? As far as I can see, we do the migration now in the runtime upgrade?

Also I think you are looking for the [translate](https://docs.rs/frame-support/latest/frame_support/storage/trait.IterableStorageMap.html#tymethod.translate) call in the migration?

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
+					acc_id32.encode()
+				);
+				let acc_id: LinkableAccountId = acc_id32.into();
+				TmpConnectedAccounts::<T>::insert(did_id, acc_id, val);
+				connected_accounts += 1;
 			});
+			log::info!("🔎 DidLookup: Migrated all {:?} ConnectedAccounts", connected_accounts);
+
+			// Move TmpStorage
+			move_storage::<Pallet<T>>(b"TmpConnectedDids", b"ConnectedDids");
```
ntn-x2, line 110: Use rather the `move_storage::Pallet::<T>` syntax.

ntn-x2 (changes requested):
Definitely a nice improvement over the previous version, but I still have few questions about some design choices, and I would also love the code to be adjusted and restructured a little bit, so that it follows the style and conventions we have used everywhere else in the codebase.

pallets/pallet-did-lookup/src/benchmarking.rs
```diff
@@ -57,70 +61,115 @@ benchmarks! {
 		let previous_did: T::DidIdentifier = account("prev", 0, SEED + 1);
 		let connected_acc = sr25519_generate(KeyTypeId(*b"aura"), None);
 		let connected_acc_id: T::AccountId = connected_acc.into();
+		let linkable_id: LinkableAccountId = connected_acc_id.clone().into();
 		let bn: <T as frame_system::Config>::BlockNumber = 500_u32.into();
```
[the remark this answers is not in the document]
ntn-x2: If that is not the case, we should at least benchmark all signatures and always consider the one the user is actually providing, for the weight to be as accurate as possible.

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
-		// to be sure.
-		let total_weight: Weight =
-			ConnectedDids::<T>::iter().fold(initial_weight, |total_weight, (account, record)| {
-				ConnectedAccounts::<T>::insert(record.did, account, ());
-				// One read for the `ConnectedDids` entry, one write for the new
-				// `ConnectedAccounts` entry.
-				total_weight.saturating_add(T::DbWeight::get().reads_writes(1, 1))
+		if Pallet::<T>::current_storage_version() == StorageVersion::new(3) {
+			// already on version 3
+			<T as frame_system::Config>::DbWeight::get().reads_writes(1, 0)
```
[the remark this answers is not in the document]
weichweich: never understood why rust removed the `return`. It's that much less readable IMHO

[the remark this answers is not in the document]
ntn-x2: Is it removed? You can still return.

pallets/pallet-did-lookup/src/account.rs
weichweich: If I'm not mistaken `.map_err(drop)` just drops what the error contains i.e. it maps a Result<X, String> to a Result<X, ()> which makes it more light weight.

pallets/pallet-did-lookup/src/migrations.rs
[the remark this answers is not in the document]
weichweich: yeah but this is the last statement in the function. So this will be return. You can't take it out of the if branch. But you don't see that this is the last statement in the function. You have to guess that it is.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -197,7 +209,22 @@ pub mod pallet {
 				Error::<T>::InsufficientFunds
 			);
 
-			Self::add_association(sender, did_identifier, account)?;
+			match req {
+				AssociateAccountRequest::Substrate(acc, proof) => {
```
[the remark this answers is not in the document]
trusch: I think this is not exactly what you asked for, but its very close. The problem is that the request is not self-contained, we need to pass in the did and the expiration from one layer above.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -151,7 +164,10 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T> {
+	impl<T: Config> Pallet<T>
+	where
+		T::AccountId: Into<LinkableAccountId>,
```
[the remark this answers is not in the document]
trusch: This is a result of hard coding the AccountId20 and AccountId32 here in the pallet. To support the old methods without much effort (like associate_sender) I added this requirement here which is automatically fullfilled. Now I can just call into() and for example the sender is translated into a LinkableAccountId. Adding the constraint to the config would basically reintroduce the LinkableAccountId type parameter that I just removed. I also unfortunately didn't find a way of adding requirements to type parameters from frame_system::Config without creating a wrapper type like LinkableAccountId. Something like `AccountId: <Self as frame_system::Config>::AccountId + Into<LinkableAccountId>;` doesnt work unfortunately.

pallets/pallet-did-lookup/src/account.rs
trusch: I added an impl for TryFrom, now its properly handled :)

pallets/pallet-did-lookup/src/lib.rs
trusch: Done :)

pallets/pallet-did-lookup/src/benchmarking.rs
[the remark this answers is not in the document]
trusch: I already added a second benchmark which checks the ecdsa signatures. Can you point me to some docs which show how to update the weight files?

pallets/pallet-did-lookup/src/migrations.rs
weichweich: even if the keys change, you can still simple drain the old storage and write it back in the new one without the additional intermediate storage?

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -331,4 +344,31 @@ pub mod pallet {
...
+			}
+		}
+
+		pub fn get_linkable_account(&self) -> LinkableAccountId {
+			match self {
+				AssociateAccountRequest::Dotsama(acc, _) => LinkableAccountId::AccountId32(acc.clone()),
+				AssociateAccountRequest::Ethereum(acc, _) => LinkableAccountId::AccountId20(*acc),
+			}
+		}
+	}
```
weichweich, line 373: i like to keep the impl block and the struct close together. Maybe even move them to their own file.

pallets/pallet-did-lookup/src/signature.rs
```diff
@@ -16,18 +16,32 @@
...
+}
 
-pub(crate) fn get_wrapped_payload(payload: &[u8]) -> Vec<u8> {
-	PAYLOAD_BYTES_WRAPPER_PREFIX
-		.iter()
-		.chain(payload.iter())
-		.chain(PAYLOAD_BYTES_WRAPPER_POSTFIX.iter())
-		.copied()
-		.collect()
+pub(crate) fn get_wrapped_payload(payload: &[u8], wrap_type: WrapType) -> Vec<u8> {
```
weichweich, line 31: wrapper type should also be `pub(crate)` I guess?

pallets/pallet-did-lookup/src/benchmarking.rs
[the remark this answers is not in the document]
ntn-x2: The `associate_account` uses an Sr25519 signature, but for everyone else than you there is no way to know whether that is actually the most expensive among the `MultiSignature` variants. So I would at least add the Ecdsa case for the `MultiSignature` as well.
Once that is done (and the benchmarks run), you can update the extrinsic weight to take the maximum of the three options, by doing something similar to what is done here: https://github.com/KILTprotocol/kilt-node/blob/40c48d5f89aacac2971ddede247561fd9073a218/pallets/did/src/lib.rs#L501

pallets/pallet-did-lookup/src/benchmarking.rs
```diff
@@ -57,70 +61,115 @@ benchmarks! {
...
 		let previous_did: T::DidIdentifier = account("prev", 0, SEED + 1);
 
 		make_free_for_did::<T>(&caller);
 
 		// Add existing sender -> previous_did connection that will be replaced
-		Pallet::<T>::add_association(caller.clone(), previous_did.clone(), caller.clone()).expect("should create previous association");
-		assert!(ConnectedAccounts::<T>::get(&previous_did, &caller).is_some());
-		let origin = T::EnsureOrigin::generate_origin(caller.clone(), did.clone());
+		Pallet::<T>::add_association(caller.clone(), previous_did.clone(), caller.clone().into()).expect("should create previous association");
+		assert!(ConnectedAccounts::<T>::get(&previous_did, &linkable_id).is_some());
```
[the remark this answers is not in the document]
ntn-x2: Alright let's keep it in then and refactor it in a different PR 👍

pallets/pallet-did-lookup/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: Ok, let's keep it in then, unless @weichweich has a different feeling about it. It does not hurt us in any way right now.

pallets/pallet-did-lookup/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: Yes, good enough 👍

pallets/pallet-did-lookup/src/benchmarking.rs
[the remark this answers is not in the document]
trusch: I added the required benchmarks. Could you tell me how to actually run them and update the weights?

weichweich:
@trusch once the conflicts are resolved you should be able to run the benchmarks by posting this comment:

> /bench runtime spiritnet-runtime pallet-did-lookup

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
+				(connected_dids.saturating_add(connected_accounts))
+					.saturating_mul(2)
+					.saturating_add(1),
+			)
+		}
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		assert_eq!(Pallet::<T>::on_chain_storage_version(), 2);
```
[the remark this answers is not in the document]
trusch: But it is STORAGE_VERSION-1 in that case, I dont think this will work. It is a constant so its resolved at compile time. I can't really imagine how it would be 2 even in the try-runtime

pallets/pallet-did-lookup/src/migrations.rs
trusch: doesn't work, can you suggest how its better?

pallets/pallet-did-lookup/src/migrations.rs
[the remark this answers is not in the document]
ntn-x2: You can see an example here: https://github.com/KILTprotocol/kilt-node/pull/343/files#diff-1731a28fbf5a49737824c0a1abb4a7c11a3f221a33789a288d764a45a0878d83=.

The concept is to update the constant to the new value, and then in the pre-runtime check whether the value stored on chain is different than the constant. If they are the same, the migration will not be applied. Otherwise, the migration is executed and the new value written in the storage.

pallets/pallet-did-lookup/src/migrations.rs
ntn-x2: Sorry you are right. I mistook this for something else. We decided a while ago as a styling convention to prefer `Type::<T>::operation()` over `<Type<T>>::operation()` since they are both valid syntaxes. But this is a different case, so I will resolve the comment, sorry for the spam 😄

pallets/pallet-did-lookup/src/migrations.rs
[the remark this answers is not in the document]
trusch: :+1:

rpc/did/runtime-api/src/lib.rs
```diff
@@ -79,14 +82,25 @@ pub struct DidLinkedInfo<
 ///
 /// This will be returned by the runtime and processed by the client side RPC
 /// implementation.
-pub type RawDidLinkedInfo<DidIdentifier, AccountId, Balance, Key, BlockNumber> =
-	DidLinkedInfo<DidIdentifier, AccountId, Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>, Balance, Key, BlockNumber>;
+pub type RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId, Balance, Key, BlockNumber> = DidLinkedInfo<
```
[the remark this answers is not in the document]
trusch: The account id is the address of the deposit owner of a link.

pallets/pallet-did-lookup/src/benchmarking.rs
```diff
@@ -57,70 +65,214 @@ benchmarks! {
 		let previous_did: T::DidIdentifier = account("prev", 0, SEED + 1);
 		let connected_acc = sr25519_generate(KeyTypeId(*b"aura"), None);
 		let connected_acc_id: T::AccountId = connected_acc.into();
+		let linkable_id: LinkableAccountId = connected_acc_id.clone().into();
```
ntn-x2, line 68: I guess we don't need the `associate_account` benchmark case anymore since we have all the other cases covered below.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -172,9 +174,8 @@ pub mod pallet {
 		#[pallet::weight(<T as Config>::WeightInfo::associate_account())]
 		pub fn associate_account(
 			origin: OriginFor<T>,
-			account: AccountIdOf<T>,
+			req: AssociateAccountRequest,
```
ntn-x2, line 177: [PLACEHOLDER] The extrinsic weight must be updated with the new value once the benchmarks are through

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,144 @@
...
+				(connected_dids.saturating_add(connected_accounts))
+					.saturating_mul(2)
+					.saturating_add(1),
+			)
+		}
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		assert_eq!(Pallet::<T>::on_chain_storage_version() < Pallet::<T>::current_storage_version());
```
ntn-x2, line 119: I think what you were looking for is

```suggestion
		assert!(Pallet::<T>::on_chain_storage_version() < Pallet::<T>::current_storage_version());
```

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,144 @@
...
+		log::info!(
+			"🔎 DidLookup pre migration: Number of connected accounts {:?}",
+			connected_account_count
+		);
+		Ok(())
 	}
 
 	#[cfg(feature = "try-runtime")]
 	fn post_upgrade() -> Result<(), &'static str> {
+		assert_eq!(Pallet::<T>::on_chain_storage_version() == Pallet::<T>::current_storage_version());
```
ntn-x2, line 141: Same here.

Either

```suggestion
		assert_eq!(Pallet::<T>::on_chain_storage_version(), Pallet::<T>::current_storage_version());
```

or

```suggestion
		assert!(Pallet::<T>::on_chain_storage_version() == Pallet::<T>::current_storage_version());
```

ntn-x2 (commented):
Didn't re-review everything again, but the changes look very good to me. I just have few more nitpicks, and then we can get this beast merged!

pallets/pallet-did-lookup/src/migrations.rs
trusch: There is no new storage since we didn't want to change the name of the storage entry.
Its like `a_vec.iter().for_each( |i| a_vec.push_back(2*i))`, it doesn't work.

pallets/pallet-did-lookup/Cargo.toml
```diff
@@ -16,11 +16,17 @@ pallet-balances = {branch = "polkadot-v0.9.24", git = "https://github.com/parity
...
 
 [dependencies]
 codec = {package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
+hex = { version = "0.4", default-features = false }
+impl-serde = { version = "0.3.1", default-features = false }
+libsecp256k1 = { version = "0.7", default-features = false, features = [ "hmac" ] }
 log = "0.4.17"
 scale-info = {version = "2.1.1", default-features = false, features = ["derive"]}
+serde = { version = "1.0.101", default-features = false, features = [ "derive" ] }
+sha3 = { version = "0.10", default-features = false }
```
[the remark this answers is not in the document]
trusch: I tried applying this suggestion but couldn't get it to build regardless how I change it. Even adding the std feature to sha3 destroys it completely.

pallets/pallet-did-lookup/src/migrations.rs
trusch: updated in [719e7d9](https://github.com/KILTprotocol/kilt-node/pull/355/commits/719e7d9bb8c32b7443f4e128b684f9dc1774282a)

pallets/pallet-did-lookup/src/migrations.rs
trusch: clarified in https://github.com/KILTprotocol/kilt-node/pull/355/commits/f049707cb4bd3d7486a6e8424eca79009d2908b1

pallets/pallet-did-lookup/src/migrations.rs
```diff
@@ -16,74 +16,151 @@
...
+
+			// Move TmpStorage
+			move_storage::<Pallet<T>>(b"TmpConnectedDids", b"ConnectedDids");
+			move_storage::<Pallet<T>>(b"TmpConnectedAccounts", b"ConnectedAccounts");
 
-		Pallet::<T>::current_storage_version().put::<Pallet<T>>();
+			StorageVersion::new(3).put::<Pallet<T>>();
 
+			<T as frame_system::Config>::DbWeight::get().reads_writes(
+				(connected_dids.saturating_add(connected_accounts)).saturating_mul(2),
```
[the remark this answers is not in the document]
trusch: clarified in https://github.com/KILTprotocol/kilt-node/pull/355/commits/f049707cb4bd3d7486a6e8424eca79009d2908b1

ntn-x2 (approved):
Nothing more to say, from my side 😁 Nice work!

pallets/pallet-did-lookup/src/migrations.rs
weichweich: That's way better with comments. This way I noticed that we missed the `+1` for reading the storage version? fcca140a40a6c40e99f45931fe2a99636b97a4f8

weichweich (approved):
LGTM!

PR #356, opened 2022-04-20
feat: refcount for DIDs

ntn-x2:
This PR fixes https://github.com/KILTprotocol/ticket/issues/1893.

~~The idea is to declare a `LifecycleHandler` for each of the pallets, which basically allows third parties to monitor the key events for each pallet.~~
~~The current status of the PR has added the functions in the `did` pallet to increase/decrease the refcount for a given DID, and also added an example application in the `attestation` pallet.~~

~~The way Parity does it (as far as I have seen), is to simply make all pallets that interact with an account's refcount dependent on the system pallet, which is something a pallet has to do anyway. In our case, that is probably not the best idea, and I have not been able to come up with a better choice than having a delegate, i.e., the runtime, handle the refcount changes.~~

~~One main problem is that right now the methods in the handler are fallible, which breaks our current design of making all the checks before writing something on storage.~~

~~One solution includes having prechecks for each operation (e.g., `can_attest`, similarly to what the attestation AC does), but it feels that we are going in a direction that is way too verbose for what we want to do.
Another solution would make the trait a bit less generic in scope, and simply having `attestation_created` and `attestation_deleted` functions, and somehow make those infallible.~~

~~Ideas?~~

From https://github.com/KILTprotocol/mashnet-node/pull/356/commits/c569e422a0b5fc112e0c3bbd4f82f4f7d72258a9 on, the design has changed to make use of a builder that checks whether a value can be incremented or decremented an only then returns an object that can be used to perform the operation. In this way, it is possible to create such object at check time, and call the operation on the object at a later time, and this avoids us to break the chec-first-commit-later approach we have been taking so far.

## Checklist

- [ ] Implement a test object that can be used to test that a given extrinsic has properly increased/decreased the refcount of an identity
- [ ] Update and run the DID benchmark code that now generates weights to increase and decrease a DID's refcount
- [ ] Address the FIXMEs and TODOs


pallets/did/src/identity_consumer.rs
```diff
@@ -0,0 +1,204 @@
...
+	// to AsRef, if possible
+	Identity: Into<DidIdentifierOf<T>> + Clone,
+{
+	fn decrement(&mut self) -> frame_support::dispatch::Weight {
+		Pallet::<T>::decrement_consumers_unsafe(&self.0 .1.clone().into());
+		T::WeightInfo::decrement_consumers()
+	}
+}
+
+impl<T: Config, Identity> IdentityConsumer<Identity, u32> for Pallet<T>
```
weichweich, line 80: I would call it Provider since the DidPallet provides identities and doesn't consume them

ntn-x2: The way Parity calls them is consumer for anything that depends on the identity, and provider for anything that make the identity usable. In this case I think we are counting the number of consumers so that the identity cannot be removed if > 0, not interested about whether the identity is usable or not (providers > 0).

ntn-x2: But I think you are right in a sense since this would be more like an `IdentityManager`, which could handle both providers and consumers. Will probably have to think of a better naming 😄

ntn-x2:
This PR has fallen too much behind. Will close this and will work on a new one, if this becomes a pressing issue.

PR #359, opened 2022-04-27
feat: add launch pallet removal migration

wischli:
## fixes KILTProtocol/ticket#1951
* Removes the launch pallet
* Adds migration to kill all storage related to the KiltLaunch storage prefix `0x37be294ab4b5aa76f1df3f80e7c180ef`

## Checklist:


runtimes/common/src/migrations.rs
```diff
@@ -0,0 +1,77 @@
...
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use core::marker::PhantomData;
+use frame_support::traits::Get;
+use hex_literal::hex;
+
+pub struct RemoveKiltLaunch<R>(PhantomData<R>);
+impl<R: frame_system::Config> frame_support::traits::OnRuntimeUpgrade for RemoveKiltLaunch<R> {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		let items = match frame_support::storage::unhashed::kill_prefix(&hex!("37be294ab4b5aa76f1df3f80e7c180ef"), None)
```
ntn-x2, line 26: Can this be generated from the pallet name string?

runtimes/common/src/migrations.rs
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		// FIXME: Why does this fail?
+		log::info!(
+			"🚀 Pre check: Launch pallet storage exists {}?",
+			frame_support::storage::migration::have_storage_value(
+				&hex!("37be294ab4b5aa76f1df3f80e7c180ef"),
+				// b"KiltLaunch"
```
ntn-x2, line 50: Why not deriving the HEX from the name directly?

runtimes/common/src/migrations.rs
wischli: I wanted to do it this way but since we are removing the pallet from the runtime, we can't derive it's pallet name from the storage. You can verify the hex yourself via the Polkadot Apps
Screenshot 2022-04-28 at 10 22 03

runtimes/common/src/migrations.rs
wischli: The hex macro requires an input of type string. I did not manage to quickly get it running with any permutation of the pallet name, e.g. `hex!(b"KiltLaunch".into())`. Just using `b"KiltLaunch"` without hex! wrapping leads to a different storage prefix.

runtimes/common/src/migrations.rs
ntn-x2: Was just playing around, and this is what apparently works: `let hash = sp_io::hashing::twox_128(b"KiltLaunch");`.

No need to change it, but I think it's better for readability and for potentially fix any issues we identify in the future.

runtimes/common/src/migrations.rs
```diff
@@ -0,0 +1,63 @@
...
+		};
+		<R as frame_system::Config>::DbWeight::get().writes(items.into())
+	}
+
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		let prefix: [u8; 16] = sp_io::hashing::twox_128(b"KiltLaunch");
+
+		assert!(
+			sp_io::storage::next_key(&prefix).map_or(true, |next_key| next_key.starts_with(&prefix)),
```
ntn-x2, line 48: Probably not relevant, but shouldn't the assertion fail in case there is no key under the given prefix, as we do indeed expect a key?

```suggestion
			sp_io::storage::next_key(&prefix).map_or(false, |next_key| next_key.starts_with(&prefix)),
```

ntn-x2 (approved):
LGTM! I only have one small question about the logic of the `pre_upgrade()` migration check.

runtimes/common/src/migrations.rs
wischli: Good catch! You are correct. Fixed in a4c3a5e

PR #360, opened 2022-04-28
chore: upgrade and clean up

weichweich:
## No Ticket

* Use RuntimeDebug to avoid manual implementation
* update build script for standalone
* upgrade dependencies
* don't disable default features for dev-dependencies
* fix accounts for sporran chain

## Checklist:

### DIFF
--- a/pallets/did/src/did_details.rs
+++ b/pallets/did/src/did_details.rs
@@ -21,12 +21,14 @@ use frame_support::{
 	ensure,
 	storage::{bounded_btree_map::BoundedBTreeMap, bounded_btree_set::BoundedBTreeSet},
 	traits::Get,
+	RuntimeDebug,
 };
-use kilt_support::deposit::Deposit;
 use scale_info::TypeInfo;
 use sp_core::{ecdsa, ed25519, sr25519};
 use sp_runtime::{traits::Verify, MultiSignature, SaturatedConversion};
 use sp_std::{convert::TryInto, vec::Vec};
+
+use kilt_support::deposit::Deposit;
 
 use crate::{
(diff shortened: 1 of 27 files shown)


pallets/did/src/did_details.rs
```diff
@@ -21,21 +21,23 @@ use frame_support::{
...
+
 use crate::{
 	errors::{DidError, InputError, SignatureError, StorageError},
 	service_endpoints::DidEndpoint,
 	utils, AccountIdOf, BalanceOf, BlockNumberOf, Config, DidCallableOf, DidIdentifierOf, KeyIdOf, Payload,
 };
 
 /// Types of verification keys a DID can control.
-#[derive(Clone, Decode, Debug, Encode, Eq, Ord, PartialEq, PartialOrd, TypeInfo, MaxEncodedLen)]
+#[derive(Clone, Decode, RuntimeDebug, Encode, Eq, Ord, PartialEq, PartialOrd, TypeInfo, MaxEncodedLen)]
```
wischli, line 40: Ah nice. I suppose we can get rid of other custom debug impls then as well, e.g. in parachain staking.

wischli (approved):
LGTM!

I suppose we can assume scale {codec, info} are backwards compatible? And even if not, we would run into errors during the release pipeline testing.

pallets/did/src/did_details.rs
weichweich: yes. Thought it might not work, but it does. :) [b6542ca](https://github.com/KILTprotocol/mashnet-node/pull/360/commits/b6542ca4b12b7ddb7ac8284bcaef76e73e00c841)

PR #364, opened 2022-05-31
refactor: update rilt para id from 2015 to 2108

wischli:
* Switches RILT para id from 2015 to 2108 (our supposedly new Kusama Para ID soon™️)

Context: Several parachain teams requested to be consistent with para ids across Rococo and Kusama to simplify XCM testing. Our current Kusama para ID is 2086, however we propose to swap soon. Then, we would be running with 2108.


ntn-x2 (approved):
LGTM!

PR #365, opened 2022-06-01
fix: rilt para id

wischli:
* Fixes incorrect ParaId introduced in #364 as we indeed keep the same parachain ID [2086 on Kusama](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkusama-rpc.polkadot.io#/parachains), when we swap the lease with the one from 2108.

Screenshot 2022-06-01 at 18 13 58


ggera (approved):

ggera:
LGTM

PR #368, opened 2022-06-09
use ci-linx:production base image

ggera:
## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 # this container builds the kilt-parachain binary from source files and the runtime library
 # pinned the version to avoid build cache invalidation
 
-# c4f0daf3-20220110 -> rust version rustc 1.58.0-nightly (46b8e7488 2021-11-07)
-FROM paritytech/ci-linux:f203f298-20220421 as builder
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,7 @@
 # this container builds the kilt-parachain binary from source files and the runtime library
 # pinned the version to avoid build cache invalidation
 
-# c4f0daf3-20220110 -> rust version rustc 1.58.0-nightly (46b8e7488 2021-11-07)
-FROM paritytech/ci-linux:f203f298-20220421 as builder
+FROM docker.io/paritytech/ci-linux:production as builder
```
wischli, line 4: I fear that the `production` image will be bumped up by Parity while we want to have full control over bumping the image to reduce error likelihood when building.

I suggest to stay with a [distinct hash](https://hub.docker.com/r/paritytech/ci-linux/tags).

```suggestion
FROM paritytech/ci-linux:eb1f6a26-20220518
```

ggera: Yup, I was kinda puzzled. Since ther toolchain file has precedence over the docker image, we kinda have no control.

wischli (approved):
LGTM

PR #377, opened 2022-07-01
chore: 1s mashnet-node dev runtime

ntn-x2:
# Fixes https://github.com/KILTprotocol/ticket/issues/1942.

It cleans up some stuff that is not needed for a dev instance of the mashnet-node, and makes the block time 1s. Currently, integration tests on the SDK are passing if run individually, but failing if run all in parallel. I think this is something that rather needs fixing on the SDK side.


weichweich (approved):

ntn-x2:
The SDK uses a different container for each test suite. This means that if all tests are run against the same node instance, they will fail as the resigning feature has also been removed. Hence, the failure was expected behaviour and we can move on and merge this PR.

runtimes/standalone/src/lib.rs
```diff
@@ -666,8 +647,7 @@ construct_runtime!(
 		// // System scheduler.
 		// Scheduler: pallet_scheduler = 32,
 
-		// Vesting. Usable initially, but removed once all vesting is finished.
-		Vesting: pallet_vesting = 33,
```
wischli, line 670: Did you check with BTE whether removing Vesting does not introduce a critical change? AFAIK, we still have vesting in Spiritnet and thus they will probably have some unit tests involving vesting?

ntn-x2: But BTE would never use the mashnet-node dev, right? They would use Peregrine, which still has the vesting stuff.

ntn-x2: The vesting stuff was not populated in the genesis. Plus, we are talking about a dev environment. I will ask the BTE people whether they need it. If they don't for the near future, I would suggest to leave it as it is an re-enable it later on if and when needed.

PR #379, opened 2022-07-11
feat: call in did creation

weichweich:
## Convenience for the DID creation
fixes https://github.com/KILTprotocol/ticket/issues/1897

Creating a DID and setting all keys requires two signatures. Submitting a call with the create op reduces this to one.

## Checklist:


weichweich:
No need, will be solved by: https://github.com/KILTprotocol/kilt-node/pull/551

PR #387, opened 2022-07-26
chore: remove unused dependencies

ntn-x2:
Remove unused dependencies using `cargo udeps`. It also updates the `all.fish` script to check for `--all-targets` and remove the actual tests running, since we do not need those.


weichweich (approved):
Love it!

PR #388, opened 2022-07-27
feat: deposits pallet

trusch:
This is not meant to be merged. It only serves as a playground for ideas and will be completely rewritten once we consolidated our thoughts on it.

## How to test:
If necessary provide information about dependencies (specific configuration, branches, database dumps, etc.)

- Step 1
- Step 2
- etc.

## Checklist:


pallets/pallet-deposits/src/deposit_change_handler.rs
```diff
@@ -0,0 +1,39 @@
+use frame_system::pallet_prelude::OriginFor;
+use sp_runtime::DispatchResult;
+
+use crate::{Config, DepositIdOf, AccountIdOf, BalanceOf};
+
+/// DepositChangeHandler is meant to be implemented on the runtime level to allow it to react to the event when a deposit changed
+pub trait DepositChangeHandler<T: Config> {
```
ntn-x2, line 7: This trait should probably be generic over `Deposit`, `AccountId` and `Balance` rather than constraining the type. It makes it easier to mock and test.

pallets/pallet-deposits/src/deposit_change_handler.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn on_deposit_paid(_deposit: &DepositIdOf<T>, _account: &AccountIdOf<T>, _amount: BalanceOf<T>) {
+        // noop
+    }
+    fn on_deposit_taken(_deposit: &DepositIdOf<T>, _account: &AccountIdOf<T>, _amount: BalanceOf<T>) {
+        // noop
+    }
+}
+
+// DepositChecker is implemented by the pallet and is meant to be used in other pallets which need deposits to be checked.
+pub trait DepositChecker<T: Config> {
```
ntn-x2, line 33: Same as for the `DepositChangeHandler` trait above.

pallets/pallet-deposits/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+			// emit event
+			Self::deposit_event(Event::DepositTaken(deposit_id, sender, new_total));
+
+			Ok(())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		/// retrieve the total amount of a deposit pot
+		pub fn check_deposit(deposit_id: &DepositIdOf<T>) -> BalanceOf<T> {
```
ntn-x2, line 256: Where are this and `ensure_deposit` functions used? They are not called anywhere inside the pallet, and all they do is accessing the storage of this pallet, so I think we can get rid of the trait altogether.

pallets/pallet-deposits/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+			if block.saturating_sub(current_block) < T::MinimalNoticePeriod::get() {
+				return Err(Error::<T>::NoticePeriodTooShort.into());
+			}
+
+			// create the announcement
+			let announcment = DepositTakeAnnouncement {
+				block,
+				amount,
+			};
+			Announcments::<T>::insert(&deposit_id, &sender,  announcment);
```
ntn-x2, line 200: This is also consuming storage space, so there should probably be another deposit which is released when the original deposit is actually taken.

pallets/pallet-deposits/src/deposit_change_handler.rs
```diff
@@ -0,0 +1,39 @@
...
+        account: &AccountIdOf<T>,
+        amount: BalanceOf<T>,
+    );
+}
+
+impl<T: Config> DepositChangeHandler<T> for () {
+    fn on_deposit_paid(_deposit: &DepositIdOf<T>, _account: &AccountIdOf<T>, _amount: BalanceOf<T>) {
+        // noop
+    }
+    fn on_deposit_taken(_deposit: &DepositIdOf<T>, _account: &AccountIdOf<T>, _amount: BalanceOf<T>) {
```
ntn-x2, line 27: Maybe would be good to either include total balance before and after, or the balance that was taken and the total balance remaining.

ntn-x2 (commented):
In general I think the idea is good. I see this pallet being instantiated multiple times once for each class of currency that could be used for reserves. Very curious to know where and how this pallet would be used!

weichweich:
stale

PR #389, opened 2022-07-28
chore: check for unused dependencies in CI

ntn-x2:
This PR adds one additional check in the `clippy` job for the `test` stage.

Although it is not desirable to run it for every commit, `cargo udeps` runs in few seconds, and because of an open issue it cannot be run into its own job without first running one of `cargo check`, `cargo clippy`, or `cargo build`. Hence, I would suggest to run it on every commit to capture any unused dependencies, and reduce both compilation times and the size of the produced binaries.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,12 +9,14 @@ clippy:
   stage: test
   timeout: 1 hours
   script:
-    - rustup component add clippy --toolchain nightly
-    - cargo +nightly install cargo-udeps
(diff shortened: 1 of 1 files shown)


weichweich (approved):
LGTM

ntn-x2:
Don't seem to be able to make it run on the CI pipeline because of its known bug. Not worth spending any more time at the moment. Might give it another try in the future.

ntn-x2:
Closing again. Doesn't work either by replacing `clippy` with `check`.

PR #390, opened 2022-08-01
chore: fix serde bounds

ntn-x2:
This PR improves the bounds on structs that use `u128` types so that only the bound of the required fields are actually overridden, while the rest is left to the default implementation of `derive(Serialize, Deserialize)`.

It also updates dependencies of `serde` and `serde_json` to the latest available version.


wischli (approved):
LGTM!

weichweich (approved):

PR #392, opened 2022-08-02
fix: changes to public credentials based on feedback

ntn-x2:
This PR is based on top of https://github.com/KILTprotocol/kilt-node/pull/378 and introduces the changes agreed in internal meetings, partially fixes https://github.com/KILTprotocol/ticket/issues/2029.

Specifically, the following changes have been introduced:

- The identifier of a public credential is now generated on chain and is dependent on the credential content. Specifically, it is computed as `H(scale_encoded_credential_input || scale_encoded_attester_identifier)`. This means that there can only be one credential per (subject, attester, authorisation ID, ctype, claims). If any of those elements change, the new credential identifier will be different and hence the credentials will be considered as separate, at least for what the runtime is concerned
- Because we have a reverse index from credential ID to credential subject, to get a single credential it is sufficient to provide such ID, instead of having to provide both ID and subject. This makes it easier to retrieve details of a credential by its ID only
- The `claimer_signature` and `nonce` have disappeared, while the `root_hash` is generated on chain according to the point above
- More information is stored on chain for a given credential now, such as the ctype hash, the attester information, and the authorization ID.
- Given the more rich information under each credential, the RPC function `get_credentials` has been updated to support an optional filter parameter, which instructs the server to perform any additional filtering once all the credentials for a given subject have been fetched. Right now, the two possible filters are by ctype hash and by attester (they cannot be combined together, so it is currently **not** possible to filter by ctype AND attester in the same request).

### Example RPC call to filter public credentials by ctype hash

```
curl -s -H "Content-Type: application/json" -"id":1, "jsonrpc":"2.0", "method": "get_credentials", "params":["did:asset:eip155:0.erc721:0x06012c8cf97bead5deae237070f9587f8e7a266d", {"ctypeHash": "0x0000000000000000000000000000000000000000000000000000000000000000"}]}' http://127.0.0.1:9933 | jq
```

### Example RPC call to filter public credentials by attester

```
curl -s -H "Content-Type: application/json" -"id":1, "jsonrpc":"2.0", "method": "get_credentials", "params":["did:asset:eip155:0.erc721:0x06012c8cf97bead5deae237070f9587f8e7a266d", {"attester": "4nwPAmtsK5toZfBM9WvmAe4Fa3LyZ3X3JHt7EUFfrcPPAZAm"}]}' http://127.0.0.1:9933 | jq
```


pallets/public-credentials/src/tests.rs
```diff
@@ -217,76 +168,344 @@ fn add_not_enough_balance() {
...
+				None,
+			));
+
+			let stored_public_credential_details = Credentials::<Test>::get(&subject_id, &credential_id)
+				.expect("Public credential details should be present on chain.");
+
+			// Test this pallet logic
+			assert!(stored_public_credential_details.revoked);
+
+			// Revoking the same credential does nothing
```
wischli, line 258: Q: Maybe we actually want to throw when double revoking?

pallets/public-credentials/src/tests.rs
```diff
@@ -297,81 +516,128 @@ fn add_ctype_not_found() {
...
-				None,
-			)
-			.expect("Adding the same attestation again should not fail");
+				credential_id,
+				Some(MockAccessControl(wrong_submitter))
+			));
+
+			// Test this pallet logic
+			assert_eq!(Credentials::<Test>::get(&subject_id, &credential_id), None);
+			assert_eq!(CredentialSubjects::<Test>::get(&credential_id), None);
```
wischli, line 572: Better to use `is.none()`

```suggestion
			assert!(Credentials::<Test>::get(&subject_id, &credential_id),.is_none());
			assert!(CredentialSubjects::<Test>::get(&credential_id).is_none());
```

pallets/public-credentials/src/tests.rs
```diff
@@ -382,79 +648,72 @@ fn remove_unauthorized() {
...
+				credential_id
 			));
-			// Test interactions with attestation pallet
-			assert_eq!(Attestations::<Test>::get(&claim_hash), None);
 
 			// Test this pallet logic
-			assert_eq!(Credentials::<Test>::get(&subject_id, &claim_hash), None);
-			assert_eq!(CredentialsUnicityIndex::<Test>::get(&claim_hash), None);
+			assert_eq!(Credentials::<Test>::get(&subject_id, &credential_id), None);
+			assert_eq!(CredentialSubjects::<Test>::get(&credential_id), None);
```
wischli, line 667:
```suggestion
			assert!(Credentials::<Test>::get(&subject_id, &credential_id).is_none());
			assert!(CredentialSubjects::<Test>::get(&credential_id).is_none());
```

wischli (approved):
LGTM apart from the very minor nitpicks I found.

pallets/public-credentials/src/tests.rs
ntn-x2: Yes I thought about that, but I really do not see any issues in revoking something that is already revoked. It won't basically change anything, so I thought it would be fine to simply proceed. What do you think?

pallets/public-credentials/src/credentials.rs
```diff
@@ -74,10 +42,19 @@ pub struct Credential<
 /// block. The block number is used to query the full content of the credential
 /// from archive nodes.
 #[derive(Encode, Decode, Clone, MaxEncodedLen, RuntimeDebug, PartialEq, Eq, PartialOrd, Ord, TypeInfo)]
-pub struct CredentialEntry<BlockNumber, AccountId, Balance> {
+pub struct CredentialEntry<CTypeHash, Attester, BlockNumber, AccountId, Balance, AuthorizationId> {
+	/// The hash of the CType used for this attestation.
+	pub ctype_hash: CTypeHash,
+	/// The attester of the credential.
+	pub attester: Attester,
+	/// A flag indicating the revocation status of the credential
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/lib.rs
```diff
@@ -168,20 +177,16 @@ pub mod pallet {
...
 
-	// Reverse map to make sure that the same claim hash cannot be issued to two
-	// different subjects by issuing it to subject #1, then removing it only from
-	// the attestation pallet and then issuing it to subject #2.
-	// This map ensures that at any time a claim hash is only linked (i.e., issued)
-	// to a single subject.
-	// Not exposed to the outside world.
+	// This map ensures that at any time a credential is only linked (i.e., issued)
+	// to a single subject, as it maps from a credential ID to the subject it was
+	// issued to.
```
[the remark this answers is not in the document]
ntn-x2: Good catch! Yeah it was `pub(crate)` before, then I opened it up for the new runtime API definition, but forgot to update the comment! https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/lib.rs
```diff
@@ -191,15 +196,25 @@ pub mod pallet {
...
 		/// A public credentials has been removed.
 		CredentialRemoved {
 			/// The subject of the removed credential.
 			subject_id: T::SubjectId,
-			/// The root hash of the removed credential.
-			claim_hash: ClaimHashOf<T>,
+			/// The id of the removed credential.
+			credential_id: CredentialIdOf<T>,
+		},
+		/// A public credentials has been revoked.
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/lib.rs
```diff
@@ -191,15 +196,25 @@ pub mod pallet {
...
+		CredentialRevoked {
+			/// The id of the revoked credential.
+			credential_id: CredentialIdOf<T>,
+		},
+		/// A public credentials has been unrevoked.
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/lib.rs
```diff
@@ -233,124 +241,169 @@ pub mod pallet {
...
+			)
+			.into())
+		}
+
+		/// Revokes a public credential.
+		///
+		/// If a credential was already revoked, this function does not fail but
+		/// simply results in a noop.
+		///
+		/// Only authorized callers can revoke the credential.
```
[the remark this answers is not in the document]
ntn-x2: Thanks! https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/tests.rs
ntn-x2: Nice one! https://github.com/KILTprotocol/kilt-node/pull/392/commits/e95d50e7eee6f3542cd090e82930a842d5c0f0fd

pallets/public-credentials/src/tests.rs
wischli: Yeah its fine. I just wanted to raise awareness such that we are all on the same page. It really does not matter much and makes it easier to implement in the SDK.

rpc/public-credentials/src/lib.rs
```diff
@@ -31,24 +31,30 @@ use sp_api::ProvideRuntimeApi;
 use sp_blockchain::HeaderBackend;
 use sp_runtime::{generic::BlockId, traits::Block as BlockT};
 
+/// Filter that can be used after the credentials for a given subject have been
+/// retrieved from the blockchain state.
+pub trait PublicCredentialsFilter<Credential> {
```
weichweich, line 36: There is a [trait in substrate that does this](https://paritytech.github.io/substrate/master/frame_support/traits/trait.Contains.html). This is used everywhere where they want to filter somethings. I think we should follow this even tho the naming of [`Everything`](https://paritytech.github.io/substrate/master/frame_support/traits/enum.Everything.html) and [`Nothing`](https://paritytech.github.io/substrate/master/frame_support/traits/enum.Nothing.html) is a bit confusing.

runtimes/common/src/constants.rs
```diff
@@ -434,7 +434,7 @@ pub mod public_credentials {
 	use super::*;
 
 	/// The size is checked in the runtime by a test.
-	pub const MAX_PUBLIC_CREDENTIAL_STORAGE_LENGTH: u32 = 257;
+	pub const MAX_PUBLIC_CREDENTIAL_STORAGE_LENGTH: u32 = 355;
```
weichweich, line 437: is there a reason for `355` and not `357`?

weichweich (changes requested):
Would like to use the substrate Contains instead of our own thing.

runtimes/common/src/constants.rs
ntn-x2: It's the result of running the unit tests for the runtimes as in https://github.com/KILTprotocol/kilt-node/blob/e95d50e7eee6f3542cd090e82930a842d5c0f0fd/runtimes/spiritnet/src/tests.rs#L106. Agree this is not very intuitive, but that's the structure we have in place right now.

rpc/public-credentials/src/lib.rs
ntn-x2: This does not work in our case. The trait has an associated function and does not take `&self` as a parameter, while we need to know both the values being filtered (from the retrieved collection) and the one that we want to filter by (the passed filter).

ntn-x2: Giving the value to filter in the parameter also would not work, since we could have two different properties that are of the same type, e.g., `Hash`, and the filter would not be able to tell the two apart.

runtimes/common/src/constants.rs
ntn-x2: The increase in value is because we now store more information for a given credential, including CType (for filtering), and authorization info.

weichweich: ah right i remember now. 🙈 It was me who introduced this right?

rpc/public-credentials/src/lib.rs
weichweich: ah ok didn't notice that

runtimes/common/src/constants.rs
ntn-x2: ... and me who pointed out it wasn't super clear 😇

weichweich (approved):
LGTM

PR #393, opened 2022-08-03
test: test runtime

weichweich:


nodes/parachain/src/chain_spec/clone.rs
```diff
@@ -0,0 +1,223 @@
...
+		},
+	))
+}
+
+// pub fn load_clone_spec() -> Result<ChainSpec, String> {
+// 	ChainSpec::from_json_bytes(&include_bytes!("../../res/clone.json")[..])
+// }
+
+#[allow(clippy::too_many_arguments)]
+fn testnet_genesis(
```
ntn-x2, line 182: Because it's used also for production deployments, maybe it could be renamed to something else.

ntn-x2 (approved):
Looks good to me, and I only had a minor nitpick 👍

runtimes/clone/src/lib.rs
```diff
@@ -0,0 +1,734 @@
...
+	type PotId = PotId;
+	type MaxCandidates = MaxCandidates;
+	type MinCandidates = MinCandidates;
+	type MaxInvulnerables = MaxInvulnerables;
+	// should be a multiple of session or things will get inconsistent
+	type KickThreshold = Period;
+	type ValidatorId = <Self as frame_system::Config>::AccountId;
+	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
+	type ValidatorRegistration = Session;
+	type WeightInfo = ();
```
wischli, line 304: Reminder: We need to generate weights or use:

```suggestion
	type WeightInfo = pallet_collator_selection::weights::SubstrateWeight<Runtime>;
```

runtimes/clone/src/lib.rs
```diff
@@ -0,0 +1,734 @@
...
+
+parameter_types! {
+	pub const UncleGenerations: u32 = 0;
+}
+
+impl pallet_authorship::Config for Runtime {
+	type FindAuthor = pallet_session::FindAccountFromAuthorIndex<Self, Aura>;
+	type UncleGenerations = UncleGenerations;
+	type FilterUncle = ();
+	type EventHandler = ();
```
wischli, line 255: On a note: If we wanted to mint block authoring rewards, we would need to configure it here via `(CollatorSelection)`. But since we don't want that, this is perfect.

wischli (approved):
LGTM but we are missing weights for the `collator_selection_pallet`.

wischli:
Note: I have been successfully running and upgrading this chain in a local para env.

runtimes/clone/src/lib.rs
weichweich: Good catch. It's not callable anyways so we won't benchmark yet.

runtimes/clone/src/lib.rs
weichweich: good catch didn't think about that

PR #398, opened 2022-08-24
feat: add migration extrinsics

wischli:
## fixes KILTProtocol/ticket#2161
* Adds call builder for relay chain (bootstraps a few calls to the relay chain, e.g. XCM or batch)
* Adds pallet to send extrinsics required to for the migration 
  * XCM Call from the para- to the relay chain, e.g. to trigger para id swap lease on current relay chain
  * Set associated relay block number check to `Any` or `Strictly`

## Checklist:


ntn-x2 (approved):
👍

weichweich (approved):

PR #409, opened 2022-09-12
fix: clone and allocation while benchmarking

weichweich:
Moves all allocations our of the benchmarking clause. The allocations would otherwise influence the benchmarking results.

## Checklist:


ntn-x2 (commented):
Thank you! I see that the parachain staking pallet also has many origin creations and cloning inside the benchmark itself. Can you please also refactor those?

weichweich:
> Thank you! I see that the parachain staking pallet also has many origin creations and cloning inside the benchmark itself. Could you please also refactor those?

yeah good idea. But William is working on a lot of changes there. Not sure if it's a good idea. But I don't expect a huge merge conflict so I did it anyways...

ntn-x2 (approved):
👌👌👌

PR #411, opened 2022-09-14
feat: remove dyn filter & relay migration

weichweich:
## Checklist:

- [x] ‼️ Migrations to clean up storage


support/src/traits.rs
```diff
@@ -75,40 +68,3 @@ pub trait VersionMigratorTrait<T>: Sized {
 pub trait GenerateBenchmarkOrigin<OuterOrigin, AccountId, SubjectId> {
 	fn generate_origin(sender: AccountId, subject: SubjectId) -> OuterOrigin;
 }
-
-/// Trait to reflect calls to the relaychain which we support on the pallet
-/// level.
-pub trait RelayCallBuilder {
```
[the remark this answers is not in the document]
weichweich: Also thought about either keeping it for later on removing it completely. IMHO the code is not gone. It'll stay in our history and we can bring it back if we need to.

runtimes/common/src/migrations.rs
```diff
@@ -15,3 +15,55 @@
...
+		let has_storage = frame_support::storage::migration::have_storage_value(
+			b"RelayMigration",
+			b"RelayNumberStrictlyIncreases",
+			b"",
+		) && frame_support::storage::migration::have_storage_value(b"DynFilter", b"Filter", b"");
+		if has_storage {
+			Ok(())
+		} else {
+			Err("Pallets not present")
+		}
```
wischli, line 42: W: Use `ensure` instead separated for both pallets.

runtimes/common/src/migrations.rs
```diff
@@ -15,3 +15,55 @@
...
+	}
+
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		let entries: u32 = 1;
+		// there is no sensible action in case of a failure. We will check manually
+		// after the migration was done.
+		let _ = frame_support::storage::unhashed::clear_prefix(&Twox128::hash(b"RelayMigration"), Some(entries), None);
+		let _ = frame_support::storage::unhashed::clear_prefix(&Twox128::hash(b"DynFilter"), Some(entries), None);
```
wischli, line 50: S: Add least post some success and debug message based on the result of the clearing.

wischli (commented):
My two cents

PR #413, opened 2022-09-16
refactor: separate runtime api from custom rpc

wischli:
* Normalizes all runtime APIs (DID, Staking, Public Credentials) by moving them into separate crates within the `runtime-api` directory of root


runtime-api/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod did;
```
ntn-x2, line 20: I am not sure making each runtime API a module rather than a crate on its own is better. Having separate crates does not force us to compile everything every time if only some of the APIs are used (for instance mashnet-node does not use staking stuff). I would rather make each API a different crate. What do you guys think?

pallets/parachain-staking/src/api.rs
```diff
@@ -0,0 +1,87 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
ntn-x2, line 19: Do we need these? Aren't they only used in `lib.rs` files?

runtimes/clone/Cargo.toml
```diff
@@ -17,9 +17,9 @@ static_assertions = "1.1.0"
 hex-literal = {version = "0.3.4", optional = true}
 serde = {version = "1.0.137", optional = true, features = ["derive"]}
 
-# RPC
-did-rpc-runtime-api = {path = "../../rpc/did/runtime-api", default-features = false}
+# RPC /
```
ntn-x2, line 20: I guess this is a typo?

runtime-api/src/lib.rs
ntn-x2: Something like the `rpc` folder, where each subfolder is a different crate. I would see the `runtime-api` folder be structured in the same way.

ntn-x2 (commented):
I have a question regarding the structure of the runtime stuff. Perhaps it would help to split them up in crates rather than in modules.

ntn-x2 (commented):
I have a question regarding the structure of the runtime stuff. Perhaps it would help to split them up in crates rather than in modules.

runtime-api/src/lib.rs
wischli: That seems to make sense.

wischli:
@ntn-x2 Albi wants me to merge all staking refactor PR into a single one. Since you already reviewed #384, could you quickly go over this PR before I merge it and present to Albi?

runtime-api/did/src/did_details.rs
```diff
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
+#![cfg_attr(not(feature = "std"), no_std)]
```
ntn-x2, line 18: We don't need this, I guess.

runtime-api/did/src/service_endpoint.rs
```diff
@@ -15,6 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
+#![cfg_attr(not(feature = "std"), no_std)]
```
ntn-x2, line 18: Same here.

ntn-x2, line 18: Same here.

ntn-x2 (approved):
A minor nitpick, but otherwise LGMT! 🎉

runtimes/common/src/lib.rs
```diff
@@ -16,8 +16,6 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
-//! Primitives used by the Parachains Tick, Trick and Track.
```
weichweich, line 19: nice :D

weichweich (approved):

PR #403, opened 2022-09-06
[DO NOT MERGE] feat: add weights v2

wischli:
## fixes KILTProtocol/ticket#2173

Working branch against Substrate/Polkadot/Cumulus master upstream with goal:

* Converts 1d weight (u64) to two 2d struct
```
pub struct Weight {
	/// The computational time used to execute some logic based on reference hardware.
	ref_time: u64,
	/// The size of the proof needed to execute some logic.
	proof_size: u64,
}
```

## Checklist:

- [ ] I have verified that the code works
  - [ ] No panics! (checked arithmetic ops, no indexing `array[3]` use `get(3)`, ...)
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #416, opened 2022-09-21
feat: upgrade to Polkadot v0.9.29

wischli:
## fixes KILTProtocol/ticket#2200
* Upgrades to latest Polkadot v0.9.29
* Updates weights to v1.5 (a one dimensional struct instead of u64), supersedes #403
* Changes execution order of pallets in `construct_runtime!` macro because our former trait `AllPalletsReversedWithSystemFirst` is now deprecated. To the best of my knowledge, only the staking related pallets have a dependent order which I reversed to replicate the current behaviour. Thanks to the pallet indices, there should be no side effects.

### Note worthy Substrate PRs

* https://github.com/paritytech/substrate/pull/10174
> This removes the `--state-cache` parameter and adds the new `--trie-cache-size` cli parameter. Parachain operators should now be able to just drop `--state-cache 0` and are not required to add the `--trie-cache-size` as the cache is enabled by default.
* https://github.com/paritytech/substrate/pull/12138
* https://github.com/paritytech/substrate/pull/11781
* https://github.com/paritytech/substrate/pull/11407
* https://github.com/paritytech/substrate/pull/11813
* https://github.com/paritytech/substrate/pull/11670

## Checklist:


runtimes/spiritnet/src/lib.rs
```diff
@@ -913,12 +916,12 @@ construct_runtime! {
...
-		Authorship: pallet_authorship::{Pallet, Call, Storage} = 20,
-		ParachainStaking: parachain_staking = 21,
-		Session: pallet_session = 22,
-		Aura: pallet_aura = 23,
+		// The following order MUST NOT be changed: AuraExt -> Aura -> Session -> Staking -> Authorship
 		AuraExt: cumulus_pallet_aura_ext = 24,
+		Aura: pallet_aura = 23,
+		Session: pallet_session = 22,
+		ParachainStaking: parachain_staking = 21,
+		Authorship: pallet_authorship::{Pallet, Call, Storage} = 20,
```
wischli, line 924: Weirdly, when running benchmarks, an error is thrown with this config under `AllPalletsWithSystem`:
```Thread 'main' panicked at 'AuRa authorities empty, maybe wrong order in `construct_runtime!`?', /home/willi/.cargo/git/checkouts/cumulus-59522f43471fa161/35dd14f/pallets/aura-ext/src/lib.rs:96```

When reverting the order (:= `REV` = Authorship -> Staking -> Session -> Aura -> AuraExt), the benchmarks work. That does not make sense though, as we would then run everything in reverse order compared to our current execution. Moreover, there was a user on Discord having problems with `REV` and `AllPalletsWithSystem`. The issue there was that session keys were not propagated in `--dev` environment.

Need to investigate locally...

Comment (wischli) [reply runtimes/spiritnet/src/lib.rs]: After intensive testing, I can say that Aura -> Session -> Staking -> Authorship is fine. However, AuraExt needs to be after Aura because the latter sets the authorities which are queried by AuraExt. Should hopefully be fixed now, see [134d428](https://github.com/KILTprotocol/kilt-node/pull/416/commits/134d428491556358de6b6d271bcb649590ac75e8)

**Comment by ntn-x2** [comment]
@wischli maybe we can wait to see if https://github.com/paritytech/substrate/pull/12351 gets backported into 0.9.29, so that we can push the fix with the next runtime upgrade.

**Comment by ntn-x2** [comment]
@wischli the fix has been backported now. So we should update dependencies and test that the new origin fix works.

**Comment by wischli** [comment]
> @wischli the fix has been backported now. So we should update dependencies and test that the new origin fix works.

Should be fixed in [68056c6](https://github.com/KILTprotocol/kilt-node/pull/416/commits/68056c6ae5289a8e175f758818384fcd8d41980f)

**Comment by wischli** [comment]
@ntn-x2 @weichweich I changed the base from develop to master such that we can quickly start the release process after your reviews. Since this release includes the batch utility fix, we should move quickly. If we merged into develop, we would, again, have to deal with merge conflicts from the Lovely Linker 1.8.0 release.

File: runtimes/clone/src/lib.rs [hunk 2]
```diff
@@ -197,8 +196,8 @@ impl pallet_transaction_payment::Config for Runtime {
 }
 
 parameter_types! {
-	pub const ReservedXcmpWeight: Weight = constants::MAXIMUM_BLOCK_WEIGHT / 4;
-	pub const ReservedDmpWeight: Weight = constants::MAXIMUM_BLOCK_WEIGHT / 4;
+	pub const ReservedXcmpWeight: Weight = constants::MAXIMUM_BLOCK_WEIGHT.saturating_div(4);
```
ntn-x2, line 199: Can we not saturate here but fail since these are constants?

ntn-x2 (commented):
I had few comments that I just noted were pending.

pallets/parachain-staking/src/lib.rs
```diff
@@ -2782,8 +2782,8 @@ pub mod pallet {
 		/// - Writes: (D + 1) * Balance
 		/// # </weight>
 		fn note_author(author: T::AccountId) {
-			let mut reads = Weight::one();
-			let mut writes = Weight::zero();
+			let mut reads = 1u64;
+			let mut writes = 0u64;
```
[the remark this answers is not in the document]
wischli: Weights are converted to a struct. For now, its a one dimensional u64. Thus, it's clear what `Weight::zero()` refers to, but not `Weight::one()`. Below we, we construct Weight from reads and writes of type `u64`. So it does not make sense to initialize them as `Weight`.

runtimes/clone/src/lib.rs
wischli: We could but I wanted to stay in sync with Parity, see [Cumulus](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.29/parachain-template/runtime/src/lib.rs#L217)

runtimes/peregrine/src/lib.rs
```diff
@@ -1064,9 +1068,8 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	// Executes pallet hooks in reverse order of definition in construct_runtime
-	// If we want to switch to AllPalletsWithSystem, we need to reorder the staking pallets
-	AllPalletsReversedWithSystemFirst,
+	// Executes pallet hooks in the order of definition in construct_runtime
+	AllPalletsWithSystem,
```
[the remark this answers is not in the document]
wischli: Please have a look at [my two comments](https://github.com/KILTprotocol/kilt-node/pull/416#discussion_r977553312). When this change was applied to Polkadot/Kusama/Westmint/Statemin{e|t} ([Substrate PR](https://github.com/paritytech/substrate/pull/10043), [Polkadot PR](https://github.com/paritytech/polkadot/pull/4181/files#diff-e5e76e02c0d16e79c70b024cbe3c6ea56f3249382a0f987ba203c34fcb40ed66R1453-R1463)), they made zero changes to their order, despite their comment of order dependency for babe, authorship and session.

For other parachains, I also could not find any project which actually switched any pallet order.

I had a local parachain running with the current configuration for a couple of days and did not run into errors.

runtimes/clone/src/lib.rs
weichweich: Overflow doesn't panic tho do they? And these are not "consts" in a sense of rust "consts". This gets expanded to a function that returns `constants::MAXIMUM_BLOCK_WEIGHT / 4`. So the compiler might optimize that, but he doesn't need to.

The only overflow that can happen while using div is when dividing with `-1` anyways and that's not possible since this is a u64. So I actually might also remove the saturating_div.

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
weichweich: So in other words here it's just counting reads and writes, the actual weight is constructed further below.

runtimes/standalone/src/lib.rs
```diff
@@ -118,7 +119,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	authoring_version: 4,
 	spec_version: 10740,
 	impl_version: 0,
+	#[cfg(not(feature = "disable-runtime-api"))]
 	apis: RUNTIME_API_VERSIONS,
+	#[cfg(feature = "disable-runtime-api")]
+	apis: sp_version::create_apis_vec![[]],
```
weichweich, line 125: interesting. Do you know why this was added?

weichweich (approved):
LGTM

The saturating thing should be resolved so that everyone is happy and I would like to know why the `disable-runtime-api` feature was added to the standalone.

runtimes/clone/src/lib.rs
wischli: I suppose the reason for the saturated div here is that the compiler does not accept just dividing, e.g. it throws `cannot call non-const operator in constants` for `constants::MAXIMUM_BLOCK_WEIGHT / 4`.

I am unhappy with panicking here. But if you prefer, we can do that.

runtimes/standalone/src/lib.rs
wischli: I honestly did not think beyond "yeah maybe we want to be able to disable the runtime API for some testing at some point" since this is part of all relaychain runtimes. However, that has been the case for a long time which I overlooked. So I am perfectly fine with reverting these changes.

runtimes/clone/src/lib.rs
wischli: BTW, when unwrapping or expecting the `checked_div`, the compiler also complains: 

```error: `Option::<T>::unwrap` is not yet stable as a const fn```

Do we value removing saturated division so much that we want to add `#![feature(const_option)]`?

File: rpc/did/runtime-api/Cargo.toml [hunk 6]
```diff
@@ -10,9 +10,9 @@ codec = { package = "parity-scale-codec", version = "3.1.5", default-features =
 scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
 serde = { version = "1.0.136", optional = true, features = ["derive"] }
 
-sp-api = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.28"}
-sp-runtime = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.28"}
-sp-std = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.28"}
+sp-api = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29"}
```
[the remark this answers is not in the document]
wischli: [909317a](https://github.com/KILTprotocol/kilt-node/pull/416/commits/909317ae4182809142dfb774ecd5d0f29d5778e7)

pallets/parachain-staking/src/lib.rs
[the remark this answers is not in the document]
wischli: Yes.

runtimes/clone/src/lib.rs
weichweich: aha interesting! So it's actually a const function. Didn't look close enough then.

ntn-x2: It is a constant hence using a simple `/` should be enough, as any invalid case would be caught at compilation time (when compiled in release mode, any overflow or division by zero is also caught). A saturating division would, potentially, saturate without giving any notice. I don't see why we would want to replace a simple division with a saturating one. Or am I missing something else?

runtimes/peregrine/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: 👍

runtimes/clone/src/lib.rs
weichweich: I also don't see a strong case for the saturating_div here.

wischli: > [... the compiler] throws `cannot call non-const operator in constants` for `constants::MAXIMUM_BLOCK_WEIGHT / 4`.

runtimes/standalone/src/lib.rs
weichweich: It's standalone so keep it in. We might find a use for it someday. 😂

wischli: Its also in the other runtimes 🙈

weichweich: In that case i wouldn't want something in our spiritnet runtime if we don't have a use case for it.

PR #417, opened 2022-09-29
fix: add force_batch to allowed DID calls

ntn-x2:
@weichweich @wischli maybe we can squeeze it in with the batch origin hotfix.

NOTE this is targeting `master`.


weichweich (approved):
LGMT

PR #419, opened 2022-10-07
feat: cleanup duplicate dependencies

wischli:
## fixes KILTProtocol/ticket#2225
* Removes duplicate dependencies by applying `cargo machete`
* NOTE: `cargo machete --fix` removed more deps than necessary, e.g. `scale-info` and `codec` and did not apply removed deps to features

## Checklist:


ntn-x2 (approved):
LGTM! ✈️✈️✈️

crates/assets/Cargo.toml
```diff
@@ -9,7 +9,7 @@ version = "1.7.2"
 [dependencies]
 # External dependencies
 base58 = {version = "0.2.0", default-features = false}
-hex = {version = "0.4.3", default-features = false}
+hex = {version = "0.4.3", default-features = false, features = ["alloc"]}
```
weichweich, line 12: We need the alloc feature?

weichweich: let me guess, the feature was enabled by the did pallet. Since the dependency was removed there, the feature was also not enabled?

weichweich (approved):
LGTM

crates/assets/Cargo.toml
wischli: Yes

PR #421, opened 2022-10-12
chore: bump development spec_version to 1.8.0

ntn-x2:
Right now, it is very very **very** hard for the SDK to try to make sense of how to encode and decode stuff to and from the blockchain. This is due to the fact that we have same runtime, same `spec_version`, same `transaction_version`, and same `everything_else` both in our `master` branch and in our `develop` branch.
This normally would not be an issue, but right now we have a big breaking change between the two, where `master` only deals with `AccountId32` stuff, while `develop` deals with `LinkableAccountId`.

Types in the SDK can be cleverly defined as starting from a given spec_version number, hence I am proposing the following:

_always make sure that the `spec_version` number on `develop` is higher than the `spec_version` number on `master` and any released binaries._

This also follows the release flow of the SDK, since anything that has not been released in a version **x**, will be released not earlier than the next version **y**. It also assumes that patch releases will be pushed directly to `master`, and then merged back into `develop`. There might be other solutions, and I don't have strong preference as long as it makes it easier for the SDK to decorate and augment types. So here I am suggesting to **always** merge back from `master` into `develop` and to bump the `spec_version` on `develop` to the next minor version. In this case, I am proposing to bump from 1.7.4 to 1.8.0. It now becomes easier for the SDK to say that versions until 1.7.4 do not contain said breaking change. In case we would switch to a release branch model, I am proposing `master` to always have a `spec_version` highest than the latest released version, following the same logic.

Right now we are in an exceptional situation where we have this breaking change hanging in the codebase that is requiring the SDK to implement few hacks here and there to make it work, since we have to rely on type definitions for our RPC stuff. But if we make sure that `develop` is always ahead of `master` (as it should logically be), then this is easier to handle in the SDK, especially now that the [type definitions have been merged inside the SDK codebase](https://github.com/KILTprotocol/sdk-js/pull/658), making it very easy to spot if there is any breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -339,7 +339,7 @@ checksum = "065374052e7df7ee4047b1160cca5e1467a12351a40b3da123c870ba0b8eda2a"
 
 [[package]]
 name = "attestation"
-version = "1.7.4"
+version = "1.8.0"
 dependencies = [
  "ctype",
  "frame-benchmarking",
@@ -963,7 +963,7 @@ dependencies = [
 
 [[package]]
 name = "clone-runtime"
-version = "1.7.4"
+version = "1.8.0"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-dmp-queue",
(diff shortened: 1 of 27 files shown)


wischli (approved):
LGTM and I support your proposal. If the current imposes issues on the SDK, we need to change that.

PR #422, opened 2022-10-12
feat: remove client RPC for DIDs

ntn-x2:
This PR follows the direction that Parity is taking of removing RPC calls that do nothing but forwarding it to the runtime. Plus, RPC calls do not support versioning, while runtime calls do, and this makes it very easy for the SDK to use the right types and functions depending on the runtime API version (tested on a local script using the same type definitions to connect both to Spiritnet and to a node running this PR, and both were successful).

If this PR is merged, https://github.com/KILTprotocol/kilt-node/pull/421 becomes less urgent, although we should have a strategy for breaking changes at the RPC level anyway.

## How it's used in the SDK

Please refer to the following commit: https://github.com/KILTprotocol/sdk-js/pull/643/commits/ba1c023a054d6e3726109ccf42c2fa117ac2dd0e.


rpc/did/runtime-api/src/lib.rs
```diff
@@ -125,13 +126,16 @@ sp_api::decl_runtime_apis! {
...
+		fn query_did_by_account_id(account: AccountId) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, AccountId, Balance, Key, BlockNumber>>;
 		fn query_did_by_account_id(account: LinkableAccountId) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId, Balance, Key, BlockNumber>>;
 		/// Given a did this returns:
 		/// * the DID
 		/// * public keys stored for the did
 		/// * the web3name (optional)
 		/// * associated accounts
 		/// * service endpoints
-		fn query_did(did: DidIdentifier) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId,Balance, Key, BlockNumber>>;
+		fn query_did(did: DidIdentifier) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId, Balance, Key, BlockNumber>>;
```
wischli, line 139: Shouldn't we have a `changed_in` flag for this call as well then?

wischli (commented):
Removal LGTM. However, I think you missed flagging the other DID runtime API call

Already looking forward to rebasing #413 🥲

rpc/did/runtime-api/src/lib.rs
ntn-x2: That's true! I think all of them must be marked as deprecated. Good catch!

wischli (approved):
LGTM!

PR #423, opened 2022-10-14
chore: runtime API renaming

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2243.

This PR adds a second runtime API for DIDs, which is meant to replace the `DidApi` runtime API in the future.

When this PR is deployed, the SDK can make DID runtime calls following the syntax specified in the issue linked above.
Because I would like to get this merged and released in 1.7.5 (before the SDK 0.29 is released), the node will support both versions (a minor change is needed to merge into `master` since it does not have the Ethereum linking support yet). We can then remove the old version as part of the 1.8.0 release, but that will not require changes to the SDK since we would already use the new API in it.

## How to test

Take the types and runtime call definitions as defined in https://github.com/KILTprotocol/sdk-js/pull/662 and use them in a simple node script. Try to call both the old API `api.call.didApi.queryDid(...)` and the new one `api.call.did.query(...)`, and they should yield the same result.


weichweich (approved):

wischli (approved):
LGTM

ntn-x2:
@wischli this can not be cherry-picked as part of the 1.7.5 release 🎉

PR #424, opened 2022-10-17
feat: release v1.7.5

wischli:
## RC for v1.7.5

* Release ticket: https://github.com/KILTprotocol/ticket/issues/2248
* Based on `master` v1.7.4-1 with the following additions:
  * https://github.com/KILTprotocol/kilt-node/pull/416
  * https://github.com/KILTprotocol/kilt-node/pull/422

## Checklist:


nodes/common/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
ntn-x2, line 1: I think we don't need this at this point.

rpc/did/runtime-api/src/lib.rs
```diff
@@ -69,7 +72,7 @@ pub struct DidLinkedInfo<
 	BlockNumber: MaxEncodedLen,
 > {
 	pub identifier: DidIdentifier,
-	pub accounts: Vec<AccountId>,
+	pub accounts: Vec<LinkableAccountId>,
```
ntn-x2, line 75: I think this is also wrong, since we don't have it yet on `master`.

rpc/did/runtime-api/src/lib.rs
```diff
@@ -86,14 +89,65 @@ pub struct DidLinkedInfo<
...
+		Key: Codec,
+		Balance: Codec,
+	{
+		/// Given a web3name this returns:
+		/// * the DID
+		/// * public keys stored for the did
+		/// * the web3name (optional)
+		/// * associated accounts
+		/// * service endpoints
+		#[changed_in(2)]
```
ntn-x2, line 123: We don't need API versioning either, since we don't have Ethereum linking support yet.

rpc/did/runtime-api/src/lib.rs
```diff
@@ -104,20 +158,20 @@ sp_api::decl_runtime_apis! {
...
-		fn query_did_by_account_id(account: AccountId) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, Balance, Key, BlockNumber>>;
+		fn query_by_account(account: LinkableAccountId) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId, Balance, Key, BlockNumber>>;
 		/// Given a did this returns:
 		/// * the DID
 		/// * public keys stored for the did
 		/// * the web3name (optional)
 		/// * associated accounts
 		/// * service endpoints
-		fn query_did(did: DidIdentifier) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, Balance, Key, BlockNumber>>;
+		fn query(did: DidIdentifier) -> Option<RawDidLinkedInfo<DidIdentifier, AccountId, LinkableAccountId, Balance, Key, BlockNumber>>;
```
ntn-x2, line 175: Same here, it would be `AccountId`.

ntn-x2 (commented):
Didn't look into details everything, but I should have covered all the relevant parts.

PR #426, opened 2022-10-20
chore: remove RPC for public credentials and remove `Api` suffix from runtime APIs

ntn-x2:
Since runtime APIs are easier to support in the SDK and to version, and because we did not do any crazy logic at the RPC layer, I think removing the RPC is a good idea. It also removes the `node-common` folder since there is nothing in there anymore.

The runtime API for `get_credentials` has also been modified from taking already a formed AssetDid (i.e., an enum), to accept a `Vec<u8>`, and internally it tries to decode it to a valid `AssetDid`, failing in case the input is invalid.

We probably also don't need to derive `serde` anymore for most of the types, since runtime API use SCALE encoding, but I left that out and to be tackled in a separate PR, maybe even https://github.com/KILTprotocol/kilt-node/pull/419.

### DIFF
--- a/rpc/public-credentials/runtime-api/src/lib.rs
+++ b/rpc/public-credentials/runtime-api/src/lib.rs
@@ -33,7 +33,11 @@ sp_api::decl_runtime_apis! {
 		Filter: Codec + ItemFilter<CredentialEntry>,
 		Error: Codec,
 	{
+		/// Return the public credential with the specified ID, if found.
 		fn get_credential(credential_id: CredentialId) -> Option<CredentialEntry>;
+		/// Return all the public credentials linked to the specified subject.
+		/// An optional filter can be passed to be applied to the result before being returned to the client.
+		/// It returns an error if the provided specified subject ID is not valid.
 		fn get_credentials(subject: SubjectId, filter: Option<Filter>) -> Result<Vec<(CredentialId, CredentialEntry)>, Error>;
 	}
 }


support/src/traits.rs
```diff
@@ -75,3 +75,8 @@ pub trait GenerateBenchmarkOrigin<OuterOrigin, AccountId, SubjectId> {
 pub trait GetWorstCase {
 	fn worst_case() -> Self;
 }
+
+/// Generic filter.
+pub trait ItemFilter<Item> {
+	fn should_include(&self, credential: &Item) -> bool;
+}
```
weichweich, line 82: there is a similar trait in Frame support(?).

weichweich (approved):

runtimes/spiritnet/src/lib.rs
```diff
@@ -1283,14 +1285,20 @@ impl_runtime_apis! {
...
-		fn get_credentials(subject: <Runtime as public_credentials::Config>::SubjectId) -> Vec<(Hash, public_credentials::CredentialEntry<Hash, DidIdentifier, BlockNumber, AccountId, Balance, AuthorizationId<<Runtime as delegation::Config>::DelegationNodeId>>)> {
-			public_credentials::Credentials::<Runtime>::iter_prefix(&subject).collect()
+		fn get_credentials(subject: Vec<u8>, filter: Option<PublicCredentialsFilter<Hash, AccountId>>) -> Result<Vec<(Hash, public_credentials::CredentialEntry<Hash, DidIdentifier, BlockNumber, AccountId, Balance, AuthorizationId<<Runtime as delegation::Config>::DelegationNodeId>>)>, PublicCredentialsApiError> {
+			let asset_did = AssetDid::try_from(subject).map_err(|_| PublicCredentialsApiError::InvalidSubjectId)?;
+			let credentials_prefix = public_credentials::Credentials::<Runtime>::iter_prefix(&asset_did);
+			if let Some(filter) = filter {
+				Ok(credentials_prefix.filter(|(_, entry)| filter.should_include(entry)).collect())
+			} else {
+				Ok(credentials_prefix.collect())
+			}
```
wischli, line 1301: C: Simplify. Not sure if my quick suggestion is more readable though 😅

```suggestion
			Ok(filter.map(|f| 
				credentials_prefix.filter(|(_, entry| f.should_include(entry)).collect())
			).unwrap_or(credentials_prefix.collect()))
```

wischli (approved):
LGTM!

support/src/traits.rs
wischli: Are you talking about [InstanceFilter](https://paritytech.github.io/substrate/master/frame_support/traits/trait.InstanceFilter.html) ?

ntn-x2: Yes, but in the last conversation we had (https://github.com/KILTprotocol/kilt-node/pull/378#discussion_r958170843)
we agreed we don't need any superset logic, so I would rather not use that. I would be happy to use any other trait that would work, otherwise we can keep it as it is, and get rid of it in the future if needed.

PR #427, opened 2022-10-21
refactor: deposits

weichweich:
## fixes KILTprotocol/ticket/issues/2198

* Make the deposit calls less confusing.
* refactor the code to reduce duplication

## Checklist:


support/src/traits.rs
```diff
@@ -80,3 +85,50 @@ pub trait GetWorstCase {
...
+	) -> Result<Deposit<AccountId, <Self::Currency as Currency<AccountId>>::Balance>, DispatchError>;
+
+	fn deposit_amount(key: &Key) -> <Self::Currency as Currency<AccountId>>::Balance;
+
+	fn store_deposit(
+		key: &Key,
+		deposit: Deposit<AccountId, <Self::Currency as Currency<AccountId>>::Balance>,
+	) -> Result<(), DispatchError>;
+
+	fn change_deposit_owner(key: &Key, new_owner: AccountId) -> Result<(), DispatchError> {
```
ntn-x2, line 103: This does not simply change the deposit owner tho, right? It also potentially updates the deposit amount. I think that should be reflected in the name somehow, as right now it's kinda misleading.

support/src/traits.rs
```diff
@@ -80,3 +85,50 @@ pub trait GetWorstCase {
...
+pub trait StorageItemMeter<AccountId, Key> {
+	type Currency: ReservableCurrency<AccountId>;
+
+	fn deposit(
+		key: &Key,
+	) -> Result<Deposit<AccountId, <Self::Currency as Currency<AccountId>>::Balance>, DispatchError>;
+
+	fn deposit_amount(key: &Key) -> <Self::Currency as Currency<AccountId>>::Balance;
+
+	fn store_deposit(
```
ntn-x2, line 98: Since we have knowledge of the `AccountId`, I think it would be cleaner if the `store_deposit` function would take the deposit owner, and does everything internally, returning the deposit in the result. Returning the new deposit could also be something that `change_deposit_owner` and `update_deposit` could do.

ntn-x2 (commented):
I like the idea. I would also not exclude migrate the deposits under a single pallet piece by piece, and not all in the same migration. Each release might migrate one or few different pallets, for instance, so that eventually we achieve the goal of having all the deposits in the same storage space, which would be much nicer and cleaner.

support/src/traits.rs
weichweich: the goal of the trait is to reduce code duplication. That's done by moving the parts that change to the pallet, while the parts that stay the same into the default trait implementation. The code that never changes are actually only three lines of code

```rust
free_deposit::<AccountId, Self::Currency>(&deposit);

let deposit = Deposit {
	owner: new_owner,
	..deposit
};
Self::Currency::reserve(&deposit.owner, deposit.amount)?;
```

If we would now move parts of this to the `store_deposit` function, we would duplicate code again. I guess with "cleaner" you mean that the deposit amount and reserve is set in the `store_deposit` call and therefor an invalid state is much harder to achieve?

ntn-x2: I mean that the caller of the trait does not have to have knowledge of how to build a deposit, since that would become responsibility of the trait implementer.
I guess I don't fully understand your argument about code duplication. Now you would have to create the deposit type outside and give it to the `store_deposit()` function every time a deposit is to be reserved. If it goes inside the `store_deposit()` function, it is only implemented once by the trait implementer, and the client could access it as the function return value or, if it does not care, could simply ignore the result. Does it make sense what I am saying?

ntn-x2: The `update_deposit()` could then simply call `free_deposit()` and then return whatever `store_deposit()` returns, aka the new deposit or web3name ownership stuff.

ntn-x2: I guess what I am saying is that while `store_deposit()` takes a deposit as a parameter, `update_deposit()` does everything internally. I think the latter is a better approach.

support/src/traits.rs
weichweich: the easiest to understand would be that there is one call to update the owner and one to update the amount. I think that's how it should be. If you want both, you batch them together... [da0bdff](https://github.com/KILTprotocol/kilt-node/pull/427/commits/da0bdff5d2d433f2b8e1b849f7a0caebbb9722bb)

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -294,31 +297,35 @@ pub mod pallet {
...
-		#[pallet::weight(<T as Config>::WeightInfo::transfer_deposit())]
-		pub fn transfer_deposit(origin: OriginFor<T>, account: LinkableAccountId) -> DispatchResult {
+		#[pallet::weight(<T as Config>::WeightInfo::change_deposit_owner())]
+		pub fn change_deposit_owner(origin: OriginFor<T>, account: LinkableAccountId) -> DispatchResult {
 			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
+			let subject = source.subject();
 
 			let record = ConnectedDids::<T>::get(&account).ok_or(Error::<T>::AssociationNotFound)?;
-			ensure!(record.did == source.subject(), Error::<T>::NotAuthorized);
+			ensure!(record.did == subject, Error::<T>::NotAuthorized);
```
wischli, line 313: Food for thought for a potential later PR: Every pallet has their own version of `AccessDenied`. Maybe we want to normalize this in the future.

pallets/public-credentials/src/lib.rs
```diff
@@ -478,6 +481,42 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Changes the deposit owner.
+		///
+		/// The balance that is reserved by the current deposit owner will be
+		/// freed and balance of the new deposit owner will get reserved.
+		///
+		/// The subject of the call must be the owner of the web3name.
```
wischli, line 490: CP strikes 🎳
```suggestion
		/// The subject of the call must be the owner of the credential.
```

support/src/traits.rs
wischli: I'm in favour of the current approach! I also don't see what we win by removing the deposit input arg from `store_deposit`. You would just move the creation of the deposit into `store_deposit` and thus create duplicate code for all implementations of the  `StorageDepositCollector` trait.

IMO, it's super clean as is! E.g., the implementing pallets only call `change_deposit_owner` and `update_deposit` which generically create the deposit and forward that to that custom implementations of `store_deposit`.

wischli (approved):
LGTM!

support/src/traits.rs
```diff
@@ -80,3 +85,50 @@ pub trait GetWorstCase {
 pub trait ItemFilter<Item> {
 	fn should_include(&self, credential: &Item) -> bool;
 }
+
+pub trait StorageDepositCollector<AccountId, Key> {
+	type Currency: ReservableCurrency<AccountId>;
+
+	fn deposit(
```
wischli, line 92: S: Add docs for all exposed functions

pallets/pallet-did-lookup/src/lib.rs
weichweich: yes agree 100%

support/src/traits.rs
weichweich: [f1ef9ca](https://github.com/KILTprotocol/kilt-node/pull/427/commits/f1ef9ca483f68deb53716b38cb95482b00ae764b)

pallets/public-credentials/src/lib.rs
weichweich: [f1ef9ca](https://github.com/KILTprotocol/kilt-node/pull/427/commits/f1ef9ca483f68deb53716b38cb95482b00ae764b)

pallets/public-credentials/src/benchmarking.rs
```diff
@@ -177,6 +177,73 @@ benchmarks! {
...
+
+		ctype::Ctypes::<T>::insert(&ctype_hash, attester);
+		Pallet::<T>::add(origin, creation_op).expect("Pallet::add should not fail");
+		let credential_id_clone = credential_id.clone();
+
+		let origin = RawOrigin::Signed(deposit_owner);
+	}: _(origin, credential_id_clone)
+	verify {
+		assert_eq!(
+			Credentials::<T>::get(subject_id, &credential_id)
```
ntn-x2, line 240: Am I missing something or this test is not really testing? The verify condition would be met even without calling `update_deposit`.

pallets/pallet-web3-names/src/benchmarking.rs
```diff
@@ -161,6 +161,28 @@ benchmarks! {
...
+		let origin_create = T::OwnerOrigin::generate_origin(deposit_owner.clone(), owner);
+
+		make_free_for_did::<T>(&deposit_owner);
+		Pallet::<T>::claim(origin_create, web3_name_input.clone()).expect("Should register the claimed web3 name.");
+
+		let origin = RawOrigin::Signed(deposit_owner.clone());
+	}: _(origin, web3_name_input_clone)
+	verify {
+		let web3_name = Web3NameOf::<T>::try_from(web3_name_input.to_vec()).unwrap();
+		assert_eq!(Owner::<T>::get(&web3_name).expect("w3n should exists").deposit, Deposit {
```
ntn-x2, line 181: See point in public-credentials benchmarks.

pallets/delegation/src/lib.rs
```diff
@@ -650,32 +651,40 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::change_deposit_owner())]
+		pub fn change_deposit_owner(origin: OriginFor<T>, delegation_id: DelegationNodeIdOf<T>) -> DispatchResult {
 			let source = <T as Config>::EnsureOrigin::ensure_origin(origin)?;
 
-			let mut delegation = DelegationNodes::<T>::get(&delegation_id).ok_or(Error::<T>::DelegationNotFound)?;
+			let delegation = DelegationNodes::<T>::get(&delegation_id).ok_or(Error::<T>::DelegationNotFound)?;
 
 			// Deposit can only be swapped by the owner of the delegation node, not the
 			// parent or another ancestor.
 			ensure!(delegation.details.owner == source.subject(), Error::<T>::AccessDenied);
```
ntn-x2, line 669: Not touched in this PR, but should we change this error to be consistent with the other pallets?

pallets/delegation/src/lib.rs
```diff
@@ -650,32 +651,40 @@ pub mod pallet {
...
-			let deposit = Deposit {
-				owner: source.sender(),
-				amount: <T as Config>::Deposit::get(),
-			};
-			CurrencyOf::<T>::reserve(&deposit.owner, deposit.amount)?;
-			delegation.deposit = deposit;
-			DelegationNodes::<T>::insert(&delegation_id, delegation);
+			// Deposit can only be swapped by the owner of the delegation node, not the
+			// parent or another ancestor.
+			ensure!(delegation.deposit.owner == sender, Error::<T>::AccessDenied);
```
ntn-x2, line 685: Same here perhaps?

ntn-x2 (approved):
Few nitpicks, otherwise LGTM!

pallets/delegation/src/lib.rs
weichweich: yes will be done in a separate PR

pallets/delegation/src/lib.rs
weichweich: yes will be done in a separate PR

pallets/public-credentials/src/mock.rs
```diff
@@ -153,16 +153,17 @@ pub(crate) mod runtime {
 		block_number: <T as frame_system::Config>::BlockNumber,
 		attester: T::AttesterId,
 		ctype_hash: Option<CtypeHashOf<T>>,
+		deposit: Option<Deposit<T::AccountId, BalanceOf<T>>>,
```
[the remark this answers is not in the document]
weichweich: This method only builds the struct. So i think it's fine. `insert_public_credentials` takes care of inserting the credential correctly into storage.

support/src/traits.rs
```diff
@@ -80,3 +85,67 @@ pub trait GetWorstCase {
...
+	/// Store the new deposit information in the storage entry behind the key.
+	fn store_deposit(
+		key: &Key,
+		deposit: Deposit<AccountId, <Self::Currency as Currency<AccountId>>::Balance>,
+	) -> Result<(), DispatchError>;
+
+	/// Change the deposit owner.
+	///
+	/// The deposit balance of the current owner will be freed, while the
+	/// deposit balance of the new owner will get reserved.
```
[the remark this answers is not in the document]
weichweich: [a0a0644](https://github.com/KILTprotocol/kilt-node/pull/427/commits/a0a06448824a3eba5d61ddc32f01f7d3b596f979)

pallets/pallet-web3-names/src/benchmarking.rs
weichweich: good catch [0ba4e65](https://github.com/KILTprotocol/kilt-node/pull/427/commits/0ba4e65c87228e5fbcdbcf00d661c4dfef29fc62)

pallets/public-credentials/src/benchmarking.rs
weichweich: [0ba4e65](https://github.com/KILTprotocol/kilt-node/pull/427/commits/0ba4e65c87228e5fbcdbcf00d661c4dfef29fc62) good catch

ntn-x2 (approved):
💰💰💰

PR #428, opened 2022-10-24
refactor(delegation): move impl Pallet into pallet mod

weichweich:
## No Ticket

Move the Pallet impl block into the pallet module like we do it for all other pallets as well.

## Checklist:


wischli (approved):
LGTM

PR #430, opened 2022-11-03
feat: workspace inheritance

wischli:
## fixes KILTProtocol/ticket#2217
* Applies the [`workspace inheritance`](https://doc.rust-lang.org/cargo/reference/workspaces.html) feature of Rust 1.64.0: All dependencies sources are declared exactly once in our root `Cargo.toml` and downstreamed via `workspace = true` 
  * Version, authors, website, repo, license, edition, readme are also declared once and downstreamed
* Since **features in the root definition are additive**, I decided to not enforce any features (except for disabling `std` for all crates which are used outside of our client).
  * IMO, these features should rather be declared in the corresponding crates such that it is clear which ones are used where.
  * In the case of `serde`, `codec` and `scale-info` we could import features into root, because these are always used.
 * Normalizes structure of all of our crate tomls
 * Adds description to crates which didn't have that yet
 * Exposes more endpoints for all pallets, e.g. Documentation, Website, ...

## How to review efficiently

You should focus on the [root Cargo.toml](https://github.com/KILTprotocol/kilt-node/pull/430/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) and my text above, e.g. what is our policy in regards to default crate features.

## Checklist:


weichweich (approved):
Needs some "getting used to" i would say. The "workspace=true" everywhere is interesting.

Who does the workflow goes if we add a new dependency? We always add it in the crate with workspace=true and add it in the root toml?

wischli:
> Who does the workflow goes if we add a new dependency? We always add it in the crate with workspace=true and add it in the root toml?
That would be my take, yes.

An unfortunate side effect is that `cargo machete` did not work with the current rust version and `workspace inheritance`. Currently, we have no unused dependencies in any of our crates. By following the above rule, I would hope this to stick. In the end, machete should work with a newer nightly version soon again, hopefully.

Cargo.toml
```diff
@@ -10,3 +10,170 @@ members = [
...
+try-runtime-cli = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29"}
+ 
+# Polkadot (with default disabled)
+pallet-xcm = {git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.29"}
+polkadot-parachain = {git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.29"}
+xcm = {git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.29"}
+xcm-builder = {git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.29"}
+xcm-executor = {git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.29"}
+
+# Client-only (with default enabled)
```
ntn-x2, line 140: Is default disabled?

ntn-x2 (approved):

Cargo.toml
wischli: As discussed, it says `# Client-only (with default enabled)` as desired 😅

PR #431, opened 2022-11-04
[DO NOT MERGE] Release 1.8.0

wischli:
The only purpose of this PR is to ensure the release branch is correct and that the reversion of the Ethereum Migration and LinkableAccountId type was applied correctly.


runtimes/spiritnet/src/lib.rs
```diff
@@ -668,6 +670,23 @@ impl pallet_utility::Config for Runtime {
 
 impl pallet_randomness_collective_flip::Config for Runtime {}
 
+impl public_credentials::Config for Runtime {
```
wischli, line 673: We are not constructing the pallet. IMO, we should still remove all `public_credentials` code right?

weichweich: 9aa8086eaad7298a1a56b361fc1f93cdd98dfd74

PR #439, opened 2022-11-29
chore: update toolchain versions

ntn-x2:
Since rust-analyzer has been merged into the main rustlang repo, it is tested for breaking changes. This means that we don't need to pin the nightly to a specific version anymore, as using the latest nightly is always guaranteed to work with the latest version of the rust-analyzer extension (try it yourselves, for me it finally just works).

This PR then does the following:

- update the toolchain file to use the latest nightly, which gives us access to potentially new clippy warnings, and to better rust-analyzer performances
- update the runtime builder to use Rust 1.64 as also nowadays done by Polkadot
- remove the deprecated `cargo-features = ["workspace-inheritance"]`, since from 1.64 it has been stabilised (this could be caught only by using a recent enough nightly version)
- Let the CI job use the same version as in the toolchain file, by removing the `+nightly` flag. This will ensure we have a consistent behaviour between local machine and CI server.

The only thing to consider is that the nightly will differ from the version that srtool uses to build the runtime WASM. That was the case also before, and it will always be until substrate will support compilation with the stable toolchain.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,4 @@
 [toolchain]
-channel = "nightly-2022-07-24"
+channel = "nightly"
+components = [ "rustfmt", "clippy" ]
 targets = [ "wasm32-unknown-unknown" ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2022-11-29"
 components = [ "rustfmt", "clippy" ]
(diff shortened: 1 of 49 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,4 @@
 [toolchain]
-channel = "nightly-2022-07-24"
+channel = "nightly-2022-11-29"
```
wischli, line 2: Im not in favor of using a newer version than in the last Polkadot release. But we could try until we run into some weird compilation errors.

wischli (approved):
LGTM but I have a remark, see below.

PR #434, opened 2022-11-14
feat: upgrade to Polkadot v0.9.32

wischli:
## fixes KILTProtocol/ticket#2289 and KILTProtocol/ticket#2296
* Upgrades from Polkadot v0.9.29 to v0.9.32
* Adds missing feature implementations for all tomls (checked via `subalfred check features` in all crates)
  * Actually necessary for application of https://github.com/paritytech/substrate/pull/10592 (see runtime changes in [dd81eac](https://github.com/KILTprotocol/kilt-node/pull/434/commits/dd81eac1c44c9309ba49265908435412eb1c0999))
* Migrates Democracy, Preimage and Scheduler pallets to use bounded Calls, see below

## Summary of changes (Polkadot v0.9.30-0.9.32)

* Weights v2 are not fully there yet, but the struct now includes the [second field for storage size](https://github.com/paritytech/substrate/pull/12277) (tracking issue: https://github.com/paritytech/substrate/issues/12176)

### Breaking Changes
* Breaking: Outer enums (https://github.com/paritytech/substrate/pull/11981) 
  * `Origin` --> `RuntimeOrigin`
  * `Call` --> `RuntimeCall`
  * `Event` --> `RuntimeEvent`
  * ~Convention seems to be to keep `Event`, `Call`, `Origin` for inner pallet usage, e.g. `Did::Origin`~ Update: We use `Runtime` prefix internally as well

### Noteworthy PRs
* https://github.com/paritytech/substrate/pull/12109
* https://github.com/paritytech/substrate/pull/12328
* https://github.com/paritytech/cumulus/pull/1585
  * Following the effort of decoupling collators and full relay nodes, this PR adds the possibility of pointing the collator to an “external” (non in-process) relay node. This is still considered experimental, and the **relay node is suggested to run on the same machine than the collator for the moment**.
  * To specify the relay full node rpc: `polkadot-parachain --alice --collator --relay-chain-rpc-url <rpc-websocket-url>`
* https://github.com/paritytech/substrate/pull/12486
  * Before this change only the interpreted WASM executor was included in per default compilations. Making the compiled executor opt-in, now, compiled WASM executor is set by default and an opt-out instead. This could lead to **big performance difference** between using these two, as more recent versions of the interpreter see a regression in performance.

### Scheduler, Preimage, Democracy Migration

* https://github.com/paritytech/substrate/pull/11649
  * Referenda, Democracy, Scheduler and Preimage pallets are all now bounded in storage access footprint
[...]

PR #440, opened 2022-11-29
feat: add block number for CTypes

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2327.

### DIFF
--- a/pallets/ctype/src/benchmarking.rs
+++ b/pallets/ctype/src/benchmarking.rs
@@ -57,7 +57,7 @@ benchmarks! {
 
 	}: _<T::Origin>(origin, ctype)
 	verify {
-		let stored_ctype_creator: T::CtypeCreatorId = Ctypes::<T>::get(&ctype_hash).expect("CType hash should be present on chain.");
+		let stored_ctype_creator: T::CtypeCreatorId = Ctypes::<T>::get(ctype_hash).expect("CType hash should be present on chain.");
 
 		// Verify the CType has the right owner
 		assert_eq!(stored_ctype_creator, did);
--- a/pallets/ctype/src/ctype_entry.rs
+++ b/pallets/ctype/src/ctype_entry.rs
@@ -1,0 +1,12 @@
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::RuntimeDebug;
+use scale_info::TypeInfo;
+
+/// Creation details of a CType.
+#[derive(Encode, Decode, RuntimeDebug, MaxEncodedLen, Eq, PartialEq, TypeInfo)]
+pub struct CtypeEntry<Creator, BlockNumber> {
+	/// Identifier of the creator.
+	pub creator: Creator,
+	/// Block number in which the creation tx was dispatched.
+	pub creation_block_number: BlockNumber,
+}
--- a/pallets/ctype/src/lib.rs
+++ b/pallets/ctype/src/lib.rs
@@ -160,7 +160,7 @@ pub mod pallet {
 
 			let hash = <T as frame_system::Config>::Hashing::hash(&ctype[..]);
 
-			ensure!(!Ctypes::<T>::contains_key(&hash), Error::<T>::CTypeAlreadyExists);
+			ensure!(!Ctypes::<T>::contains_key(hash), Error::<T>::CTypeAlreadyExists);
 
 			// *** No Fail except during withdraw beyond this point  ***
 
@@ -176,7 +176,7 @@ pub mod pallet {
 
 			T::FeeCollector::on_unbalanced(imbalance);
 			log::debug!("Creating CType with hash {:?} and creator {:?}", hash, creator);
-			Ctypes::<T>::insert(&hash, creator.clone());
+			Ctypes::<T>::insert(hash, creator.clone());
 
 			Self::deposit_event(Event::CTypeCreated(creator, hash));
 
--- a/pallets/ctype/src/lib.rs
+++ b/pallets/ctype/src/lib.rs
@@ -73,7 +73,7 @@ pub mod pallet {
 	use sp_std::vec::Vec;
 
 	/// The current storage version.
-	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
 
 	/// Type of a CType hash.
 	pub type CtypeHashOf<T> = <T as frame_system::Config>::Hash;
--- a/pallets/ctype/src/lib.rs
+++ b/pallets/ctype/src/lib.rs
@@ -42,6 +42,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+pub mod ctype_entry;
 pub mod default_weights;
 
 #[cfg(any(feature = "mock", test))]
@@ -58,6 +59,8 @@ pub use crate::{default_weights::WeightInfo, pallet::*};
 
 #[frame_support::pallet]
 pub mod pallet {
+	use crate::ctype_entry::CtypeEntry;
+
 	use super::*;
 	use frame_support::{
 		pallet_prelude::*,
@@ -74,6 +77,8 @@ pub mod pallet {
 
 	/// Type of a CType hash.
 	pub type CtypeHashOf<T> = <T as frame_system::Config>::Hash;
+
+	pub type CtypeEntryOf<T> = CtypeEntry<<T as Config>::CtypeCreatorId, BlockNumberFor<T>>;
 
 	/// Type of a CType creator.
 	pub type CtypeCreatorOf<T> = <T as Config>::CtypeCreatorId;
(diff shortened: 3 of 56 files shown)


runtimes/spiritnet/src/lib.rs
```diff
@@ -553,6 +553,8 @@ impl delegation::Config for Runtime {
 	type Deposit = constants::delegation::DelegationDeposit;
 }
 
+type CTypeBlockNumberSetOrigin = pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCollective, 3, 5>;
```
weichweich, line 556: i would argue that we should do that using a general vote and not the TC.

weichweich (commented):
Looks good. Just a few points that I would like to get clarified.

runtimes/spiritnet/src/lib.rs
wischli: By general vote you are referring to a referendum or the Council collective? I agree we shouldnt give special power to special entities.

weichweich: referendum

pallets/ctype/src/benchmarking.rs
```diff
@@ -57,10 +58,32 @@ benchmarks! {
...
 		// Verify the CType has the right owner
-		assert_eq!(stored_ctype_creator, did);
+		assert_eq!(stored_ctype_entry.creator, did);
+	}
+
+	set_block_number {
+		let caller = account("caller", 0, SEED);
+		let did: T::CtypeCreatorId = account("did", 0, SEED);
+
+		let ctype: Vec<u8> = (0u8..u8::MAX).cycle().take(MAX_CTYPE_SIZE.try_into().unwrap()).collect();
```
wischli, line 71: I wonder if worst case behaviour would rather be having all bytes at max value?

pallets/ctype/src/lib.rs
```diff
@@ -176,11 +186,41 @@ pub mod pallet {
...
+		///
+		/// Emits `CTypeUpdated`.
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::set_block_number())]
+		pub fn set_block_number(
+			origin: OriginFor<T>,
+			ctype_hash: CtypeHashOf<T>,
+			block_number: BlockNumberFor<T>,
+		) -> DispatchResult {
+			T::OverarchingOrigin::ensure_origin(origin)?;
+			Ctypes::<T>::try_mutate(ctype_hash, |ctype_entry| -> Result<(), Error<T>> {
```
wischli, line 212: Unnecessary declaration of return function.

```suggestion
			Ctypes::<T>::try_mutate(ctype_hash, |ctype_entry|  {
```

runtimes/common/src/migrations.rs
```diff
@@ -16,67 +16,77 @@
...
+	storage_alias,
+	traits::{GetStorageVersion, OnRuntimeUpgrade},
 };
+use sp_runtime::traits::{Get, Zero};
+use sp_std::marker::PhantomData;
+
+use ctype::{CtypeCreatorOf, CtypeEntryOf};
+
+#[storage_alias]
+type MigrationCounter<T: ctype::Config> = StorageValue<ctype::Pallet<T>, u32, ValueQuery>;
```
wischli, line 30: This is not the correct way of using temp storage in a migration.

In our current version of Polkadot, the pre-hook would use `Self::set_temp_storage(migration_counter, "ctype_migration_counter");` and post `Self::get_temp_storage("ctype_migration_counter")`.

After #434 in Polkadot v0.9.32+ the pre-hook actually returns any temp storage and the post one takes this as argument.

runtimes/common/src/migrations.rs
```diff
@@ -16,67 +16,77 @@
...
+			num_translations = num_translations.saturating_add(1);
+			Some(CtypeEntryOf::<T> {
+				creator: old,
+				creation_block_number: default_block_number,
+			})
+		});
+		current.put::<ctype::Pallet<T>>();
+
+		// Num translations + old version read and new version write
+		T::DbWeight::get().reads_writes(num_translations.saturating_add(1), num_translations.saturating_add(1))
```
wischli, line 72: Don't we have 2 reads from the storage versions?

```suggestion
		T::DbWeight::get().reads_writes(num_translations.saturating_add(2), num_translations.saturating_add(1))
```

runtimes/common/src/migrations.rs
```diff
@@ -16,67 +16,77 @@
...
 
-		match (has_migration_storage, has_filter_storage) {
-			(false, false) => Ok(()),
-			(true, false) => Err("RelayMigration still present"),
-			(false, true) => Err("DynFilter still present"),
-			(true, true) => Err("Pallets still present"),
+		// Use iter() on new storage so it also checks that the new values can be
+		// decoded after the migration.
+		assert_eq!(MigrationCounter::<T>::get(), ctype::Ctypes::<T>::iter().count() as u32);
+		if let Some(ctype_entry) = ctype::Ctypes::<T>::iter_values().last() {
```
wischli, line 82: To be 100% safe, we should check every element and not only the last one.

wischli (commented):
Adding to Albi's remarks, I would do some minor changes to the migration.

runtimes/spiritnet/src/lib.rs
ntn-x2: I don't see why the general public would be concerned about setting the preimage for a CType hash. Yes, it will probably be a one-off operation, but I don't see why we want to wait 2+ weeks to do something that only has technical motivations?

pallets/ctype/src/benchmarking.rs
ntn-x2: You mean all bytes with value 255? What would that change?

runtimes/common/src/migrations.rs
ntn-x2: I took this from the [Parachain staking migration](https://github.com/KILTprotocol/kilt-node/blob/621997ed4bed76a30aa9a53fdc21ef0217e3496e/pallets/parachain-staking/src/migration.rs#L65), and I thought it was a nice approach. Plus, this PR was worked on before the 0.9.32 upgrade was in, which is now, so I will probably use the new approach.

runtimes/common/src/migrations.rs
ntn-x2: Why?

wischli: Ah, didn' see it is `const`. Then forget about my remark.

runtimes/common/src/migrations.rs
wischli: How ironic the inspiration was taken by me, the one who complains 😅 Maybe the temp storage pre `0.9.32` was deprecated then. In any case, since it is only used in the `try-runtime` hooks, it would be completely fine to leave as is. Sorry about the noise.

pallets/ctype/src/benchmarking.rs
wischli: Not sure, it could have a tiny tiny tiny impact on decoding/encoding times?  😅 Regard this as irrelevant.

ntn-x2: I am fairly certain that decoding a boolean takes the same time a decoding a byte, and that the value of the bits do not matter 😊

pallets/ctype/src/ctype_entry.rs
```diff
@@ -0,0 +1,12 @@
...
+use frame_support::RuntimeDebug;
+use scale_info::TypeInfo;
+
+/// Creation details of a CType.
+#[derive(Encode, Decode, RuntimeDebug, MaxEncodedLen, Eq, PartialEq, TypeInfo)]
+pub struct CtypeEntry<Creator, BlockNumber> {
+	/// Identifier of the creator.
+	pub creator: Creator,
+	/// Block number in which the creation tx was dispatched.
+	pub creation_block_number: BlockNumber,
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/440/commits/ffc79c2e46e7bf91dae27058bb993de7ddbe10c2

pallets/ctype/src/lib.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/440/commits/147536f03f84b27c823a8296379d56fb99a2b5f7

runtimes/common/src/migrations.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/440/commits/4d92056e7b37a14655f461da57441795c1a89605

runtimes/common/src/migrations.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/440/commits/8b4c9dcc68390383f61372d428cdcbbddf6bcdcc

runtimes/spiritnet/src/lib.rs
weichweich: Since the block number is set automatically, you would only need to trust the blockchain, that the number is correct. So trust the chain == trust that polkadot checks everything and the KILT governance(all token holders) is honest and fair.
If you allow  the TC to set the block number you would need to trust them, that they set the correct number. It feels weaker.
Also since this should only be necessary once after the upgrade, i would say it's fine to wait 2 weeks.

weichweich (approved):
LGTM

I don't have a strong opinion for governance issue. If you are not convinced by the arguments, I'm also fine with leaving it like it is.

ntn-x2:
> If you are not convinced by the arguments, I'm also fine with leaving it like it is.
I understood your point, and I will go over your concern again later today or tomorrow. It might be right to do what you are suggesting.

William was suggesting we want to have a migration strategy for CTypes that mirrors what the Ethereum migration does. Do we still want to do that? I think in this case is probably unnecessary, but it is also true that this feature is not super high priority and could wait.

weichweich:
> > If you are not convinced by the arguments, I'm also fine with leaving it like it is.
> > I understood your point, and I will go over your concern again later today or tomorrow. It might be right to do what you are suggesting.
> 
> William was suggesting we want to have a migration strategy for CTypes that mirrors what the Ethereum migration does. Do we still want to do that? I think in this case is probably unnecessary, but it is also true that this feature is not super high priority and could wait.

I think it's not necessary. The current strategy is fine.

PR #441, opened 2022-12-08
feat: harmonize dispatch error names

trusch:
## fixes KILTProtocol/ticket#2290

This harmonizes the instance names in the Error enums of our pallets.

Before this change the names were kind of randomly choosen, so we had for example CTypeNotFound and DidNotPresent. Those names were bad for thwo reasons:

1) They mean the same thing but use different wording
2) They contain the module name

After this change both instances would be named NotFound.

All the other changes are something along those lines.

## How to test:
If necessary provide information about dependencies (specific configuration, branches, database dumps, etc.)

- Step 1
- Step 2
- etc.

## Checklist:


pallets/pallet-web3-names/src/lib.rs
```diff
@@ -360,8 +358,8 @@ pub mod pallet {
 		pub fn update_deposit(origin: OriginFor<T>, name_input: Web3NameInput<T>) -> DispatchResult {
 			let source = ensure_signed(origin)?;
 			let name = Web3NameOf::<T>::try_from(name_input.into_inner()).map_err(DispatchError::from)?;
-			let w3n_entry = Owner::<T>::get(&name).ok_or(Error::<T>::Web3NameNotFound)?;
-			ensure!(w3n_entry.deposit.owner == source, Error::<T>::Unauthorized);
+			let w3n_entry = Owner::<T>::get(&name).ok_or(Error::<T>::NotFound)?;
+			ensure!(w3n_entry.deposit.owner == source, Error::<T>::NotAuthorized);
```
wischli, line 362: There still exists _Unauthorized_ in the attestation pallet which hasn't been touched in terms of authorization errors. Please check all pallets again for similar occurrences.

pallets/did/src/lib.rs
```diff
@@ -409,20 +409,20 @@ pub mod pallet {
 				DidError::StorageError(storage_error) => Self::from(storage_error),
 				DidError::SignatureError(operation_error) => Self::from(operation_error),
 				DidError::InputError(input_error) => Self::from(input_error),
-				DidError::InternalError => Self::InternalError,
```
wischli, line 412: Why are we not refactoring the above ones? Eg getting rid of the _Error_ suffix for ServiceError, StorageError etc.

pallets/did/src/lib.rs
```diff
@@ -391,16 +391,16 @@ pub mod pallet {
...
-		ServiceAlreadyPresent,
+		ServiceAlreadyExists,
 		/// A service with the provided ID is not present under the given DID.
-		ServiceNotPresent,
+		ServiceNotFound,
 		/// One of the service endpoint details contains non-ASCII characters.
 		InvalidServiceEncoding,
 		/// The number of service endpoints stored under the DID is larger than
 		/// the number of endpoints to delete.
 		StoredEndpointsCountTooLarge,
```
wischli, line 401: Could this be refactored to match the _Max...Exceeded_ style?

wischli (changes requested):
More harmonization required. For a final statement, the document with all runtime existing KILT runtime errors is mandatory (assuming post PR state) as its too easy to oversee stuff.

pallets/did/src/errors.rs
```diff
@@ -22,54 +22,54 @@ use crate::did_details::DidVerificationKeyRelationship;
...
+impl From<Input> for Error {
+	fn from(err: Input) -> Self {
+		Error::Input(err)
 	}
 }
 
 /// Error involving the pallet's storage.
 #[derive(Debug, Eq, PartialEq, TypeInfo)]
-pub enum StorageError {
+pub enum Storage {
```
weichweich, line 50: What's the reason to remove the `Error` suffix here? If I see an enum called `Storage` I'm not expected error codes.

pallets/did/src/errors.rs
```diff
@@ -22,54 +22,54 @@ use crate::did_details::DidVerificationKeyRelationship;
...
 	/// The DID being created is already present on chain.
-	DidAlreadyPresent,
+	AlreadyExists,
 	/// The expected DID cannot be found on chain.
-	DidNotPresent,
+	NotFound,
 	/// The given DID does not contain the right key to verify the signature
 	/// of a DID operation.
-	DidKeyNotPresent(DidVerificationKeyRelationship),
+	DidKeyNotFound(DidVerificationKeyRelationship),
```
weichweich, line 57: We have `NotFound`, `KeyNotFound` and `DidKeyNotFound`. They all sound very similar.

In case of `DidKeyNotFound`, we can't execute the call since the DID doesn't have the required `VerificationKeyRelationship`? I would interpret that as disallowing the call and rename it to `CallNotAllowed` or similar.
In case of `NotFound` and `KeyNotFound` I would keep `DidNotFound` as it reflects that this is specific for the Did. Best would probably be to have a `NotFound(Resource)` error, where the Resource is an enum that contains all the resources that could be missing.

pallets/did/src/errors.rs
```diff
@@ -84,7 +84,7 @@ pub enum SignatureError {
 /// Error generated when some extrinsic input does not respect the pallet's
 /// constraints.
 #[derive(Debug, Eq, PartialEq, TypeInfo)]
-pub enum InputError {
+pub enum Input {
```
weichweich, line 87: Name `Input` doesn't reflect that it's error codes.

pallets/did/src/errors.rs
```diff
@@ -22,54 +22,54 @@ use crate::did_details::DidVerificationKeyRelationship;
...
 	MaxTotalKeyAgreementKeysExceeded,
 	/// The DID has already been previously deleted.
-	DidAlreadyDeleted,
+	AlreadyDeleted,
 }
 
 /// Error generated when validating a DID operation.
 #[derive(Debug, Eq, PartialEq, TypeInfo)]
-pub enum SignatureError {
+pub enum Signature {
```
weichweich, line 72: Name `Signature` doesn't reflect that it's error codes

pallets/did/src/errors.rs
```diff
@@ -22,54 +22,54 @@ use crate::did_details::DidVerificationKeyRelationship;
...
 	/// The signature is not in the format the verification key expects.
 	InvalidSignatureFormat,
```
weichweich, line 74: Since we know it's a signature error, we don't require the Signature in the error name? Same with `InvalidSignature` even though we might want to explicitly say that it's the signature part that is wrong and not the nonce part (we also have `InvalidNonce`).

```suggestion
	InvalidFormat,
```

pallets/did/src/service_endpoints.rs
```diff
@@ -59,48 +59,45 @@ pub struct DidEndpoint<T: Config> {
 impl<T: Config> DidEndpoint<T> {
 	/// Validates a given [DidEndpoint] instance against the constraint
 	/// set in the pallet's [Config].
-	pub(crate) fn validate_against_constraints(&self) -> Result<(), InputError> {
+	pub(crate) fn validate_against_constraints(&self) -> Result<(), Input> {
```
weichweich, line 62: If we would use the `errors::Input` path everywhere I would think that this change would make more sense. Seeing only `Input` there makes you think that the input is returned in case of an error.

```suggestion
	pub(crate) fn validate_against_constraints(&self) -> Result<(), errors::Input> {
```

pallets/pallet-web3-names/src/lib.rs
```diff
@@ -150,30 +150,28 @@ pub mod pallet {
...
 		/// A name that is too short is being claimed.
-		Web3NameTooShort,
+		TooShort,
 		/// A name that is too long is being claimed.
-		Web3NameTooLong,
+		TooLong,
 		/// A name that contains not allowed characters is being claimed.
-		InvalidWeb3NameCharacter,
-		/// The origin was not authorized to perform that action
-		Unauthorized,
```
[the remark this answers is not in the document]
weichweich: @trusch can you write a comment for these when you resolve them? Something like:

> This is a duplicate of the error in line 168.

weichweich (changes requested):
Can you adjust the PR description? You can remove the parts of the template that don't apply? Also there should be a doc with a description of the naming and conventions? Maybe you could commit that here as a md file?

pallets/did/src/errors.rs
trusch: Its part of the errors module, so its qualified name is did::errors::Storage which is beautiful. With the suffix it would be did::errors::StorageError.

pallets/did/src/errors.rs
```diff
@@ -22,54 +22,54 @@ use crate::did_details::DidVerificationKeyRelationship;
...
+pub enum Error {
 	/// See [StorageError].
-	StorageError(StorageError),
+	Storage(Storage),
 	/// See [SignatureError].
-	SignatureError(SignatureError),
+	Signature(Signature),
 	/// See [InputError].
-	InputError(InputError),
+	Input(Input),
```
weichweich, line 31: The comments need updating

pallets/pallet-web3-names/src/lib.rs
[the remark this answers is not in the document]
trusch: NotAuthorized was already there from the beginning, I only unified the usage of those two very similar errors (one was about a not authorized DID and the other about a not authorized account for deposit reclaiming).

pallets/pallet-web3-names/src/lib.rs
trusch: I checked, its all gone now

pallets/did/src/errors.rs
trusch: updated

pallets/did/src/errors.rs
trusch: I think now its completely refactored to be in that NotFound(Resource) shape

pallets/did/src/errors.rs
trusch: Same as with the name of did::errors::Storage, the module indicates that its all about error types

pallets/did/src/errors.rs
trusch: same as above

pallets/did/src/errors.rs
trusch: I removed the Signature part from the variants and changed from InvalidSignature case to InvalidData

pallets/did/src/service_endpoints.rs
trusch: done

pallets/did/src/errors.rs
weichweich: `did::errors::StorageError` would only be used once to bring `StorageError` into scope. After that it's always only `StorageError`. So your proposal is `errors::Storage` vs `StorageError`. I personally think `StorageError` is better since you can't forget to have the error path. If you don't use `errors::Storage` but `Storage` this becomes very misleading.

pallets/did/src/errors.rs
```diff
@@ -22,59 +22,93 @@ use crate::did_details::DidVerificationKeyRelationship;
...
+pub enum NotFoundKind {
+	/// The expected DID cannot be found on chain.
+	Did,
+	/// At least one key referenced is not stored under the given DID.
+	Key(KeyType),
+}
+
+/// Enum describing the different did key types.
+#[derive(Debug, Eq, PartialEq, TypeInfo)]
+pub enum KeyType {
```
weichweich, line 76: Even tho this introduces another nesting level i would maybe reuse the `DidVerificationKeyRelationship` here.

```
pub enum KeyType {
 Verification(DidVerificationKeyRelationship),
 KeyAgreement,
}
```

You can than implement `From<DidVerificationKeyRelationship> for KeyType` or maybe even derive that and `From<KeyType> for NotFoundKind`

When creating an error gets too deeply nested you can also add a helper function to `StorageError`.

weichweich (commented):
@ntn-x2 What do you think about `StorageError` vs `errors::Storage`? Which one do you prefer?

ntn-x2:
> @ntn-x2 What do you think about StorageError vs errors::Storage? Which one do you prefer?

My two cents are that types can also be imported directly in Rust. This means that I could do `use did::errors::Storage`, and then use `Storage` directly. This could be confusing. That's why I would leave error names with the `Error` prefix, so that even by fully importing the type and not its enclosing module, we don't have any ambiguities.

pallets/did/src/errors.rs
```diff
@@ -22,59 +22,93 @@ use crate::did_details::DidVerificationKeyRelationship;
...
 /// All the errors that can be generated when validating a DID operation.
 #[derive(Debug, Eq, PartialEq, TypeInfo)]
-pub enum DidError {
-	/// See [StorageError].
-	StorageError(StorageError),
-	/// See [SignatureError].
-	SignatureError(SignatureError),
-	/// See [InputError].
-	InputError(InputError),
+pub enum Error {
```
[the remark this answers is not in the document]
trusch: That's the top-level Error type for DIDs, so calling this `Error` in the context of the did::error module kind of makes sense. I agree that this also violates the "no error suffix" rule since technically "Error" is a suffix of "Error", but this is really an edgecase. Naming it "Did" could make sense, but it feels counter-intuitive to me. Perhaps we should modify the no-error suffix rule somehow?

trusch:
I reverted the removal of the Error suffix for the error types in the did::errors module.

ntn-x2:
@trusch please fix the CI errors, thanks!

pallet_error_naming_convention.md
```diff
@@ -0,0 +1,11 @@
+# Pallet Error Naming Conventions
+
+1) Use capitalized camel case in the variant names. For example, instead of using "NOT_FOUND" you should use "NotFound".
+
+2) Avoid using the word "error" as a suffix for the variant names. For example, instead of using "NotFoundError" you should use "NotFound". It's clear from the caller's context that this is an error.
+
+3) Avoid using the pallet name in the variant names. For example instead of "Web3NameNotFound" you should use "NotFound". 
+
+4) Try to take words from the vocabulary already defined in the code base. For example instead of introducing a new variant "NotExisting" you should use, once again, "NotFound"
```
weichweich, line 9: could you list the vocabulary here?
e.g. NotAuthorized, NotFound, AlreadyExists, MaxXYZExceeded (thats all? i guess...)

weichweich (approved):

PR #442, opened 2022-12-15
chore: add remaining bootnodes

ggera:


ntn-x2 (commented):
How can I check that this is actually correct?

weichweich:
outdated

PR #447, opened 2023-01-11
fix: toolchain version

ntn-x2:


weichweich (approved):
Thanks for cleaning my mess. ❤️

PR #448, opened 2023-01-11
feat: add merkelization of DID documents for cross-chain identity

ntn-x2:
* Adds two new traits to the DID pallets: `DidDocumentHasher`, which takes a DID document and produces a (merkle) hash, and `DidRootDispatcher`, which takes an action (DID updated or DID created) and a `MultiLocation` and does something, returning the weight consumed
* The DID pallet exposes a new extrinsic, `propagate_root`, which can be submitted by any signed origin, which will also pay for the fees. The extrinsic allows the submitter to generate _and_ propagate the root of a DID Document to a specified `MultiLocation` destination. The KILT chain will then check whether the destination has an open XCM channel and supports this operation. This part is TBD. If there is an advantage in sending a message to multiple chains at once (something like multicasting), then we could also offer an `All` option where the change is pushed to all chains that support this functionality. If not, then this extrinsic can be batched as many times as needed for the different target chains.

By making the origin signed but not restricted to anyone specifically, few use cases are possible. The target blockchain might want to call extrinsic for its users. The user itself might want to call the extrinsic to push the changes (kind of like refreshing an access token and invalidating the previous one). 

How the Merkle root is generated is also subject to change, since I am still looking for better ways to do it. This way works, although I have not copy-pasted the test cases from my previous test run. This solution, when refined, allows to prove the presence of any number of keys within a DID document.

Nevertheless, I would like to get some feedback on the approach used. 

## Alternative solutions

Before arriving to this solution, I evaluated different ones:

1. As Albi suggested, whole-pallet hashing, although would solve some of the issues I faced with other solutions, would make proofs quite large and users to have knowledge of the whole content of the DID pallet. I personally don't see any upside in following this solution, but I'm all ears.
2. Creation of a queue where to push changed DIDs, and create and dispatch root hashes in the `on_idle` or `on_finalize`: while this sounds like the most viable approach, one problem is that the queue must be bound, meaning that, in the case it gets full, new updates would not eventually be pushed to the target chains. Making the structure unbound results in the loss of the FIFO property, which is also not good. Plus, automatically dispatching XCM operations would involved more complex fee schemes than the cases where the dispatcher is well known.
3. There might be more advanced solutions where a user would be able to submit a proof taking into account the parachain header and its relay chain parent, and from there deriving the proof of the state of a DID, but I feel this is both more complicated and would suffer from the same issues as solution 1.


pallets/did/src/traits.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use crate::{did_details::DidDetails, AccountIdOf, Config, DidIdentifierOf};
+
+pub trait DidDocumentHasher<DidIdentifier, DidDetails, Output> {
+	const MAX_WEIGHT: Weight;
+
+	fn calculate_root(did: &DidIdentifier, details: &DidDetails) -> Result<(Output, Weight), DispatchError>;
+}
+
+impl<T: Config> DidDocumentHasher<DidIdentifierOf<T>, DidDetails<T>, T::Hash> for () {
```
weichweich, line 33: Just a minor comment. Maybe we could start to use structs instead of `()`. THis would make sure that it's clear what the implementation does.

runtimes/peregrine/src/lib.rs
```diff
@@ -1456,3 +1461,52 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+			WithdrawAsset(asset.clone().into()),
+			BuyExecution {
+				fees: asset,
+				weight_limit: Unlimited,
+			},
+			Transact {
+				origin_type: OriginKind::Xcm,
+				require_weight_at_most: 1_000_000_000,
+				// TODO: Replace with actual call on target chain
+				call: action.encode().into(),
```
weichweich, line 1496: this could potentially encode differently on every chain? So we would need to distinguish the target chain?

weichweich (commented):
Looks promising. The interesting part is on the other chain tho right?

pallets/did/src/traits.rs
ntn-x2: Wasn't kind of known that a trait implemented by `()` does nothing? But sure, I will change that.

runtimes/peregrine/src/lib.rs
ntn-x2: The action is always encoded/decoded in the same way, if the two chains use the same struct definition. We will have to support versioning, of course, which is not yet here. But I also not yet 100& if a transact can contain a generic payload that the other chain can parse and interpret, or if it always has to be an call encoded in the target chain's logic.

weichweich: AFAIK it always has to be a call. But would be good to check this assumption.

ntn-x2: We will have our own pallet deployed on the target chain, so no problem anyway.

weichweich: But the pallet will not have the same pallet index on every chain. So the encoding of the call will be different.

PR #449, opened 2023-01-13
chore: update relay chain references

ntn-x2:
After the move to Polkadot, Spiritnet XCM config was still referring to Kusama as the relay chain. This should be updated now.


weichweich (approved):
Good Catch. LGTM

PR #457, opened 2023-01-25
chore: refactor some more XCM config stuff

ntn-x2:
There were few things that, as far as I have understood, were not 100% consistent with what the chain does and especially does NOT do, right now. More details are provided as comments in the files.


runtimes/common/src/xcm_config.rs
```diff
@@ -146,23 +146,23 @@ parameter_types! {
 /// `AccountId`. This is used when determining ownership of accounts for asset
 /// transacting and when attempting to use XCM `Transact` in order to determine
 /// the dispatch Origin.
-pub type LocationToAccountId<RelayNetwork> = (
-	// The parent (Relay-chain) origin converts to the parent `AccountId`.
+pub type LocationToAccountId<NetworkId> = (
```
ntn-x2, line 149: The generic refers to a network ID. So especially for the future it is clearer to give the generic the proper name.

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -92,7 +89,7 @@ impl xcm_executor::Config for XcmConfig {
 	// How weight is transformed into fees. The fees are not taken out of the
 	// Balances pallet here. Balances is only used if fees are dropped without being
 	// used. In that case they are put into the treasury.
-	type Trader = UsingComponents<WeightToFee<Runtime>, RelayLocation, AccountId, Balances, Treasury>;
+	type Trader = UsingComponents<WeightToFee<Runtime>, HereLocation, AccountId, Balances, Treasury>;
```
ntn-x2, line 92: We allow unpaid execution from the relay chain. Plus, here we were saying that relay chain funds could be used to pay for the fees, when nothing else was allowed. I think keeping it local for now is the most logical choice. We will open up once we open up to other chains.

weichweich (approved):
LGTM... the seem to make sense.

I guess this was copied from statemine or similar when this was added.

PR #458, opened 2023-01-26
fix: on-chain storage version for pallets where it is missing

ntn-x2:
A while ago, we have migrated to using the `StorageVersion` way of declaring the storage version for our pallets. Nevertheless, we were not aware that such a change would only be written to storage either explicitly in a storage migration for live chains, or at genesis for new chains.

This means that we have few places where the declared storage version, which is a `const` inside each pallet, does not reflect what is written on chain in the relative entry.

This means that, for the future, **every time we add a new pallet we must also include the storage migration to write the value of the storage version on chain**.

### Peregrine

The `attestation` and `publicCredentials` pallet expose a `StorageVersion` of 1, but the on-chain value is 0.

### Spiritnet

It has the same issue as Peregrine, with the addition of the `web3Names` pallet as well.

This PR then exposes a pre-runtime hook that logs with a warning all pallets that suffer from this inconsistency, and a post-runtime check that verifies that everything has been fixed.

#### Note

The migration that updates the storage version has to be run last, so that pre- and post- runtime hooks don't trigger any unexpected behaviour.

## How to test

Compile the `kilt-parachain` binary with `cargo build -p kilt-parachain --features try-runtime`.

For Peregrine, run:

```bash
./target/debug/kilt-parachain try-runtime \
--runtime target/debug/wbuild/peregrine-runtime/peregrine_runtime.compact.compressed.wasm \
-lruntime=info \
on-runtime-upgrade --checks \
live --uri wss://peregrine.kilt.io:443/parachain-public-ws
```

For Spiritnet, run:

```bash
./target/debug/kilt-parachain try-runtime \
--runtime target/debug/wbuild/spiritnet-runtime/spiritnet_runtime.compact.compressed.wasm \
-lruntime=info \
on-runtime-upgrade --checks \
live --uri wss://spiritnet.kilt.io:443
```


runtimes/common/src/migrations.rs
```diff
@@ -81,3 +83,209 @@ impl<T: ctype::Config> OnRuntimeUpgrade for AddCTypeBlockNumber<T> {
...
+		AttestationPallet::<R>::current_storage_version().put::<AttestationPallet<R>>();
+		// Not an issue with Peregrine, but it is with Spiritnet.
+		Web3NamesPallet::<R>::current_storage_version().put::<Web3NamesPallet<R>>();
+		PublicCredentialsPallet::<R>::current_storage_version().put::<PublicCredentialsPallet<R>>();
+
+		<R as frame_system::Config>::DbWeight::get().writes(4)
+	}
+}
+
+#[cfg(feature = "try-runtime")]
```
ntn-x2, line 107: Did not find a better way to make the dependencies that are only used with `try-runtime` optional.

weichweich: Can you not make the `pre_` and `post_` implementation optional?

ntn-x2: Yes but you can't import crates only for that. The point of this solution is that the crates for the pallets are imported only when the feature is enabled.

ntn-x2: Hence the dependency for those can be made optional.

weichweich: ah you are referring to the `where` clause

weichweich (approved):

PR #460, opened 2023-02-01
[DO NOT MERGE] feat: 1.9.0 release

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2396.

PR shows the diff with the `master` branch.

For better insights into what this release changes, please refer to the [diff with develop](https://github.com/KILTprotocol/kilt-node/compare/develop...release-1.9.0?expand=1) - to see if we included all the latest features and fixes - and the [diff with 1.8.0](https://github.com/KILTprotocol/kilt-node/compare/release-1.8.0...release-1.9.0?expand=1) to check exactly what we have changed

## Edit

After going through past releases and other chain upgrade paths, I found out two migrations we did not consider:

- https://github.com/paritytech/substrate/pull/12989: no biggie for us, it is just executed after the v3 migration for the same pallet
- https://github.com/paritytech/substrate/pull/12813: this is a tricky one. We should run the migration since we need to bump the storage version of the pallet and also because the treasury starts tracking deactivated funds anyway. All the other chains I've seen, they use the XCM checking account as the account in the migration (which deactivates funds on `check_out` and reactivates them on `check_in`). We do not have that, and beyond the treasury have nothing else to be considered as inactive. Apparently Polkadot does not consider deposits as inactive. _We could consider removing the tokens sent to the pure proxy as deactivated, perhaps in the next migration. It is not critical that we do it in the same migration._

## TODO

- [x] Review
- [ ] Benchmarks


runtimes/spiritnet/src/lib.rs
```diff
@@ -1018,10 +1056,58 @@ pub type Executive = frame_executive::Executive<
...
+	(
+		// "Bound uses of call" <https://github.com/paritytech/polkadot/pull/5729>
+		pallet_preimage::migration::v1::Migration<Runtime>,
+		pallet_scheduler::migration::v3::MigrateToV4<Runtime>,
+		pallet_democracy::migrations::v1::Migration<Runtime>,
+		runtime_common::migrations::AddCTypeBlockNumber<Runtime>,
+		// The migration above must be run as last since it checks that all pallets are using the new StorageVersion
+		// properly.
+		runtime_common::migrations::MigrateToNewStorageVersion<Runtime>,
+	),
```
trusch, line 1068: we might be missing some migrations here as discussed on slack, this comment is just as a reminder

ntn-x2:
cc @weichweich once you're back. We will deploy on staging by end of the week but wait for your pair of eyes before merging and releasing on Peregrine.

trusch (approved):
To me this looks good, all important stuff seems to be there. I'd say we create an rc from this and continue testing on staging :tada:

weichweich (approved):
LGTM

Some calls are heavier now, for no apparent reason.

PR #461, opened 2023-02-03
chore: update rules for image push

ntn-x2:
Partially fixes https://github.com/KILTprotocol/ticket/issues/2405.

Right now, it's not possible to test SDK compatibility because we don't have a consistent naming for rc tags. I would suggest to keep pushing `latest-master`, `latest-develop` and `latest`, and pushing an additional `latest-rc` on rc tags, so that the SDK can pick up on those and run integration tests before we release the new node version.

A companion PR on the SDK will have to be opened if this is merged.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,9 +61,15 @@ build:
     - ./.maintain/docker-auth-config.sh
 
     - ./.maintain/build-image.sh build
+      # commit hash on every branch commit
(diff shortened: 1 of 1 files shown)


weichweich (approved):

ntn-x2:
Closing for now. We will think about a better strategy and potentially get rid of the `master` branch altogether.

PR #462, opened 2023-02-09
fix: set right env variables for srtool and subwasm

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2409.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -78,6 +78,8 @@ build-wasm-peregrine:
   script:
     - export PACKAGE=peregrine-runtime
     - export RUNTIME_DIR=runtimes/peregrine
+    - export PARACHAIN_PALLET_ID=0x50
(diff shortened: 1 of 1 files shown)


weichweich (approved):

wischli:
Would you mind sharing how there were derived?

ntn-x2:
@wischli the `PARACHAIN_PALLET_ID` is the HEX-converted index of the `parachain_system` pallet in the runtime, and `AUTHORIZE_UPGRADE_PREFIX` is the index of the [`authorize_upgrade` call inside the parachain_system pallet](https://github.com/paritytech/cumulus/blob/1faed8b0f0ac2d3b49cc74d13846a0c04b5c61b5/pallets/parachain-system/src/lib.rs#L447). It is set to 3 by default since in the past that used to be the index. Now it's changed, and a PR was merged in srtool to fix the defaults.

PR #463, opened 2023-02-16
chore: add manual trigger for Polkadot releases dependencies

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2403.

I was playing around with `subalfred` and found out this nice feature, which we might want to integrate into our update flow. The action is configured to be run manually by providing the versions we are updating from and to.

~There is one thing that would break if we keep changing the weekly board project name to include the sprint goal, I think we could start tracking that separately.~ -> replaced `project-name` with `project-number`, thanks @ggera 

What do you guys think? Would it make sense?

### DIFF
--- a/.github/workflows/polkadot-release-list.yml
+++ b/.github/workflows/polkadot-release-list.yml
@@ -23,8 +23,8 @@ on:
         - Ad96el
 
 jobs:
-  runs-on: ubuntu-latest
   create-ticket:
+    runs-on: ubuntu-latest
     steps:
       - name: Install cargo
         run: curl https://sh.rustup.rs -sSf | sh
--- a/.github/workflows/polkadot-release-list.yml
+++ b/.github/workflows/polkadot-release-list.yml
@@ -1,0 +1,49 @@
+name: create-polkadot-release-ticket
+
+on:
+  workflow_dispatch:
+    inputs:
+      current-version:
+        description: Current version of the Polkadot/Cumulus dependency
+        required: true
+        type: string
+      target-version:
+        description: Target version of the Polkadot/Cumulus dependency
+        required: true
+        type: string
+jobs:
+  runs-on: ubuntu-latest
+  create-ticket:
+    steps:
+      - name: Install cargo
+        run: curl https://sh.rustup.rs -sSf | sh
+      - name: Install subalfred
+        env:
+          - SUBALFRED_VERSION: 0.9.1
+        run: cargo install --version $SUBALFRED_VERSION --git https://github.com/hack-ink/subalfred.git subalfred
+      - name: Run subalfred for Substrate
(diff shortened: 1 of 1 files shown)


.github/workflows/polkadot-release-list.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          title: "chore: update Polkadot dependencies from ${{ inputs.current-version }} to ${{ inputs.target-version }}"
+          repository: KILTProtocol/ticket
+          token: ${{ secrets.REPO_ACCESS_TOKEN }}
+          content-filepath: out.md
+      - name: Put issue into weekly board
+        uses: peter-evans/create-or-update-project-card@v2
+        with:
+          # TODO: This will break if we change the name of the project
+          project-name: "Weekly Board 📅 Test & Release public credentials Chain 1.9.0 SDK 0.31 / CType Draft V2 / Documentation Engagement"
```
ggera, line 46:
```suggestion
          project-number: 24
```

ggera:
> Fixes [KILTprotocol/ticket#2403](https://github.com/KILTprotocol/ticket/issues/2403).
> 
> I was playing around with `subalfred` and found out this nice feature, which we might want to integrate into our update flow. The action is configured to be run manually by providing the versions we are updating from and to.
> 
> There is one thing that would break if we keep changing the weekly board project name to include the sprint goal, I think we could start tracking that separately.
> 
> What do you guys think? Would it make sense?

Looks awesome,, can we take a learning from [ticket-version-bump](https://github.com/KILTprotocol/ticket/blob/master/.github/workflows/version-bump-ticket.yml) and if possible put both in a single repo?

ggera:
> Fixes [KILTprotocol/ticket#2403](https://github.com/KILTprotocol/ticket/issues/2403).
> 
> I was playing around with `subalfred` and found out this nice feature, which we might want to integrate into our update flow. The action is configured to be run manually by providing the versions we are updating from and to.
> 
> There is one thing that would break if we keep changing the weekly board project name to include the sprint goal, I think we could start tracking that separately.
> 
> What do you guys think? Would it make sense?

Looks awesome,, can we take a learning from [ticket-version-bump](https://github.com/KILTprotocol/ticket/blob/master/.github/workflows/version-bump-ticket.yml) and if possible put both in a single repo?

ggera (approved):

ntn-x2:
There you go! I remember we had an action somewhere for this sort of ticket, but I could not find it!

Do you think it should live in the ticket repo or in the node repo? I personally think it should live in the node repo, which is also easier to access for most people.
Thanks for the suggestion, that's exactly what I was looking for!

ntn-x2:
Since there's a new Polkadot 0.9.38 release, what do you think about reviewing this PR and test it with the new release? @weichweich @trusch

weichweich (approved):

PR #465, opened 2023-02-17
feat: create a tipper committee

Ad96el:
## fixes https://github.com/KILTprotocol/ticket/issues/2424
created a new collectives for tippers. 

## Checklist:

### DIFF
--- a/runtimes/common/src/constants.rs
+++ b/runtimes/common/src/constants.rs
@@ -299,7 +299,7 @@ pub mod governance {
 		pub const TechnicalMaxProposals: u32 = 100;
 		pub const TechnicalMaxMembers: u32 = 100;
 		// Tipper Group
-		pub const TipperMaxMembers: u32 = 7;
+		pub const TipperMaxMembers: u32 = 21;
 	}
 }
 
--- a/runtimes/common/src/constants.rs
+++ b/runtimes/common/src/constants.rs
@@ -298,6 +298,8 @@ pub mod governance {
 		pub const TechnicalMotionDuration: BlockNumber = TECHNICAL_MOTION_DURATION;
 		pub const TechnicalMaxProposals: u32 = 100;
 		pub const TechnicalMaxMembers: u32 = 100;
+		// Tipper Group
+		pub const TipperMaxMembers: u32 = 7;
 	}
 }
 
@@ -418,7 +420,7 @@ pub mod tips {
 	parameter_types! {
 		pub const MaximumReasonLength: u32 = 16384;
 		pub const TipCountdown: BlockNumber = DAYS;
-		pub const TipFindersFee: Percent = Percent::from_percent(20);
+		pub const TipFindersFee: Percent = Percent::from_percent(0);
 		pub const TipReportDepositBase: Balance = deposit(1, 1);
 	}
 }
--- a/runtimes/spiritnet/src/lib.rs
+++ b/runtimes/spiritnet/src/lib.rs
@@ -510,7 +510,7 @@ impl pallet_membership::Config<TipsMembershipProvider> for Runtime {
 	type PrimeOrigin = MoreThanHalfCouncil;
 	type MembershipInitialized = ();
 	type MembershipChanged = ();
-	type MaxMembers = constants::governance::TechnicalMaxMembers;
+	type MaxMembers = constants::governance::TipperMaxMembers;
 	type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,9 +205,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.69"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "224afbd727c3d6e4b90103ece64b8d1b67fbb1973b1046c2281eed3f3803f800"
+checksum = "7de8ce5e0f9f8d88245311066a578d72b7af3e7088f32783804676302df237e4"
 
 [[package]]
 name = "approx"
@@ -267,7 +267,7 @@ dependencies = [
  "num-traits",
  "rusticata-macros",
  "thiserror",
- "time 0.3.17",
+ "time 0.3.20",
 ]
 
 [[package]]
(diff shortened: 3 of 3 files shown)


runtimes/spiritnet/src/lib.rs
```diff
@@ -945,7 +959,7 @@ construct_runtime! {
 		Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 44,
 
 		// Tips module to reward contributions to the ecosystem with small amount of KILTs.
-		TipsMembership: pallet_membership::<Instance2> = 45,
+		TipsMembership: pallet_membership::<Instance3> = 45,
```
Ad96el, line 962: For consistency, I changed the instance from `Instance2` to `Instance3`

runtimes/common/src/constants.rs
```diff
@@ -280,6 +280,8 @@ pub mod governance {
 	pub const TECHNICAL_MOTION_DURATION: BlockNumber = 4 * MINUTES;
 	#[cfg(not(feature = "fast-gov"))]
 	pub const TECHNICAL_MOTION_DURATION: BlockNumber = 3 * DAYS;
+	#[cfg(not(feature = "fast-gov"))]
+	pub const TIPPER_MOTION_DURATION: BlockNumber = 3 * DAYS;
```
Ad96el, line 284: Is the duration of three days fine?

runtimes/spiritnet/src/lib.rs
```diff
@@ -485,7 +497,7 @@ impl pallet_membership::Config<TechnicalMembershipProvider> for Runtime {
 	type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
 }
 
-type TipsMembershipProvider = pallet_membership::Instance2;
+type TipsMembershipProvider = pallet_membership::Instance3;
 impl pallet_membership::Config<TipsMembershipProvider> for Runtime {
```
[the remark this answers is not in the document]
ntn-x2: And also the callbacks for `MembershipInitialized` and `MembershipChanged` would have to be set to `TipperCommittee` so that changes are reflected in the underlying storage.

runtimes/spiritnet/src/lib.rs
```diff
@@ -485,7 +497,7 @@ impl pallet_membership::Config<TechnicalMembershipProvider> for Runtime {
 	type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
 }
 
-type TipsMembershipProvider = pallet_membership::Instance2;
+type TipsMembershipProvider = pallet_membership::Instance3;
```
ntn-x2, line 500: This would usually be a breaking change, but here I think we are good since we did not implement `MembershipInitialized` and `MembershipChanged`, so no underlying storage changes were made.

ntn-x2 (changes requested):
I left a couple of comments, but it looks very nice! Please apply the same changes also to the Peregrine runtime, since it is supposed to be as close as possible to the Spiritnet one :)

runtimes/spiritnet/src/lib.rs
weichweich: AFAIK the tippers can vote but the tippers origin has no power. So they can't do much.

Tips are not voted on, but have their separate mechanism.

runtimes/spiritnet/src/lib.rs
```diff
@@ -471,6 +471,18 @@ impl pallet_collective::Config<TechnicalCollective> for Runtime {
...
 
+type TipperCollective = pallet_collective::Instance3;
+impl pallet_collective::Config<TipperCollective> for Runtime {
+	type RuntimeOrigin = RuntimeOrigin;
+	type Proposal = RuntimeCall;
+	type RuntimeEvent = RuntimeEvent;
+	type MotionDuration = constants::governance::TipperMotionDuration;
+	type MaxProposals = constants::governance::TipperMaxProposals;
+	type MaxMembers = constants::governance::TipperMaxMembers;
+	type DefaultVote = pallet_collective::PrimeDefaultVote;
```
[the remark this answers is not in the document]
weichweich: AFAIK the tippers can vote but the tippers origin has no power. So they can't do much.

Tips are not voted on, but have their separate mechanism.

runtimes/spiritnet/src/lib.rs
```diff
@@ -471,6 +471,18 @@ impl pallet_collective::Config<TechnicalCollective> for Runtime {
 	type WeightInfo = weights::pallet_collective::WeightInfo<Runtime>;
 }
 
+type TipperCollective = pallet_collective::Instance3;
+impl pallet_collective::Config<TipperCollective> for Runtime {
```
weichweich, line 475: I actually think, that we don't need the collective pallet here. It only adds voting power. Tippers don't need that.

weichweich (commented):
I'm not convinced that we need the Collective pallet at all.

Dudleyneedham:
I would only add this information this can be closed:

- Set a maximum tip amount 1000 KILT
- Set the finders fee to 0%

Ad96el:
I just changed some variable values (TipFindersFee, Max Members), since the basic functionality is already implemented. Now, the council has to select the tippers. @Dudleyneedham Limiting a tip to 1000 Kilt is not possible in the pallet. From my point of view, it would be over engineered to make a fork from the pallet_tips to implement this feature. Since only the median is selected, the payed out tip is robust against malicious tippers.

weichweich (approved):

runtimes/common/src/constants.rs
```diff
@@ -297,6 +297,8 @@ pub mod governance {
 		pub const TechnicalMotionDuration: BlockNumber = TECHNICAL_MOTION_DURATION;
 		pub const TechnicalMaxProposals: u32 = 100;
 		pub const TechnicalMaxMembers: u32 = 100;
+		// Tipper Group
+		pub const TipperMaxMembers: u32 = 21;
```
ntn-x2, line 301: Didn't we say this would be 7? Has anything changed since then?

ntn-x2 (approved):
Good job! I have a minor comment.

runtimes/common/src/constants.rs
Ad96el: The variable `TipperMaxMembers` is just defining the max amount of tippers. The council can still just select 7 tippers. I changed it to 21, since in the near or far future, we want to select more tippers, and then we do not have to change the source code again.

PR #476, opened 2023-02-23
feat: genesis config for linking pallet

weichweich:
## No Ticket

* adds accounts for doc tests to endowed accounts
* Linking Pallet
    * MigrationState is Done in genesis
    * allow to configure pallet in genesis

I added the option to configure the DidLookupPallet in genesis, so that we can start a test chain with already populated DID-Account links. We could start adding genesis configs to the other pallets as well. Some tests require well known DIDs or w3n to exist. We could add them in the genesis config, so that the tests work out of the box with dev chains.

## Checklist:


nodes/parachain/src/chain_spec/peregrine.rs
```diff
@@ -83,6 +83,10 @@ pub fn make_dev_spec() -> Result<ChainSpec, String> {
 					get_account_id_from_seed::<sr25519::Public>("Eve"),
 					get_account_id_from_seed::<sr25519::Public>("Ferdie"),
 					get_account_id_from_seed::<sr25519::Public>("Alice//stash"),
+					get_account_id_from_seed::<sr25519::Public>("Alice//core"),
```
ntn-x2, line 86: Do you think this is more scalable and maintainable than having a fallback on the faucet seed for the docs? So if you run them locally on a mashnet-node instance you would set a `FAUCET_SEED` env variable. Otherwise we would have to change this when we change something in the docs, which might not be an immediate dependency.

weichweich: It's only for convenience. If it is not up to date anymore, no harm is done.
We could also have `FAUCET_SEED` as default and FAUCET_SEED_{CORE,...}  that overwrites the `FAUCET_SEED`.

ntn-x2: `BASE_MNEMONIC` already implies that derivation paths should be used. Else, `FAUCET_SEED` would imply that the faucet seed is used directly, and no derivation should be applied (as we did until 2 days ago).

I agree that no harm is done, it's just that a new docs section that would need a different derivation path would require a new dev version of the node for local testing. I would personally just set a `FAUCET_SEED` and use the old way of endowing accounts, if we support that in the docs repo. So my point is that we don't necessarily need to reflect that here.

weichweich: Ok. Let's change it in the docs then.

weichweich: [02ae16a](https://github.com/KILTprotocol/kilt-node/pull/476/commits/02ae16a8d530aa707421f5ef436534075035f2f7)

ntn-x2 (approved):
Sehr gut! 🚀

PR #482, opened 2023-02-27
chore: add polkadot output to Subalfred

ntn-x2:
Polkadot release branch naming has a different convention since it starts with `release`. I thought it was simply not supported by subalfred, but I was simply too tired to notice that 😄

Just launched a test run from this branch: https://github.com/KILTprotocol/kilt-node/actions/runs/4280433199/jobs/7452196855, which created the new ticket https://github.com/KILTprotocol/kilt-node/issues/481.


ggera:
Nice , now it becomes complete

ggera (approved):

PR #487, opened 2023-03-10
feat: PoV size benchmarks

weichweich:

Add support for PoV size benchmarks

## Checklist:


runtimes/peregrine/src/weights/pallet_scheduler.rs
```diff
@@ -16,123 +18,265 @@
...
+	#[test]
+	fn test_service_task_fetched() {
+		// Since this is called on on initialize we can use more than the max_extrinsic proof size.
+		assert!(
+			<crate::Runtime as frame_system::Config>::BlockWeights::get()
+				.max_block
+				.proof_size()
+				> 4199620
+		);
+	}
```
weichweich, line 233: Note that this is quite big. (bigger than the max extrinsic weight)

.maintain/runtime-weight-template.hbs
```diff
@@ -30,40 +15,70 @@
...
+			.saturating_add(Weight::from_parts(0, {{cp.slope}}).saturating_mul({{cp.name}}.into()))
+			{{/each}}
+	}
+	{{/each}}
+}
+
+#[cfg(test)]
+mod tests {
+	{{#each benchmarks as |benchmark|}}
+	{{#if (ne benchmark.base_calculated_proof_size "0")}}
```
ntn-x2, line 69: Do we need this line?

weichweich: Yeah if the size is `0` I don't add a test since it triggered a clippy warning. `u32 < 0` can never be true since there is no value in `u32` that is less than 0.

PR #489, opened 2023-03-15
feat: basic DIP structure

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2550.

**This PR will merge the changes into the `aa/dip` branch, which will be the base branch for all the future DIP improvements until we reach a version that we can merge into `develop` and deploy on some testnets in the Peregrine runtime.**

## Base structure for DIP

This PR contains the following components:

* A pallet to send identity information to one or more destinations (DIP sender)
* A pallet to receive such identity information from a source (DIP receiver)
* A support crate that contains common (versioned) types, in future also traits and auxiliary functions, if needed
* A template project consisting of 1. a template runtime for the dip-sender, with its node counterpart, 2. a template runtime for the dip-receiver, with its node counterpart, 3. an example XCM integration test using [XCM emulator](https://github.com/shaunxw/xcm-simulator) (not simulator) which connects the two template runtimes with a local Rococo relay chain.

### Design

The current version assumes that anyone on the sender chain can dispatch identity information on some target chain (with which there exists an open HRMP channel). Ultimately, all the fees for dispatching the message and process it on the receiver side will be paid by the extrinsic submitter on the target chain. Right now, the submitter only pays for the extrinsic dispatch on the source chain, and the chain itself uses its own balance (liked to its sovereign account) on the target chain to pay for the `Transact` execution. The `Transact` operation that interacts with the `dip-receiver` pallet on the target chain has an `OriginKind::Native`, so that there is no way for any user on the source chain to dispatch such messages.

Right now, the basic structure and the templates use defaults for everything, from proof generation to proof verification. One step at the time, all these defaults will be replaced with the actual features. Nevertheless, for two chains that have the same concept of `Identifier` and `AccountId`, it is already possible to demo the complete e2e flow. For the sake of our demonstration, the flow would go as follow, **assuming that both the source and target chains are registered as parachain on the same relaychain**:

1. The source chain requests to open an HRMP channel with the target chain
2. The target chain accepts the request and the channel is open
3. The target chain issues some funds (enough to cover the demo use case) to the sender chain's sovereign account
4. A user on the source chain creates a DID with a single authentication key
5. The same user (but could be anyone else), calls the extrinsic on the `dip-sender` pallet to send the merkle root of the DID to the target chain, paying for the dispatch fee
6. The source chain sends the information to the target chain
7. The user calls the `dispatch_as` extrinsic of the `dip-receiver` pallet which is deployed on the target chain, providing a merkle proof of their DID document + details on the source chain. For now this check always returns true.
8. Inside, the `dispatch_as` has a nested call which is dispatched with the new `DipOrigin`, which pallets can now use if they expect extrinsics to be DID-authorized.

## What's coming

I created tickets for each of the next steps, under the DIP milestone. The tickets will take care of:

* Adding benchmarks and weights
* Introducing proper error handling
* Introducing proper DID signature verificaiton on the target template chain (right now it's just a default behaviour)
* Introducing proper merkle proof generation (I tested that it works, but removed those pieces from this PR for the ease of understanding)
* Introducing proper merkle proof verification (same as above)
* Adding an emergency brake in the sender pallet to stop dispatching messages to a specific destination, e.g., if the sender sovereign account on the destination chain does not have any more or if the `Transact` encoding would fail
* Adding support for metadata beyond the simple DID details, e.g., proof expiration time, etc.
* Replacing user-provided weights and assets with per-chain configurations stored in the sender pallet
* More comments (I can work on this first if the code as it is it's too hard to follow)
* Adding support for version negotiation. There is a basic concept of versioning right now, but it can and should be expanded and improved.
* More unit and integration tests.

Probably more tickets will come out of this work, and will be tracked separately in the same or in future milestones.

## How to test

A basic integration test of the two template runtimes is present in the `dip-templates-xcm-tests` crate, inside the `dip-template/runtimes/xcm-tests`. It can be run as usual with `cargo test -p dip-templates-xcm-tests`.


pallets/pallet-dip-receiver/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+		pub fn dispatch_as(
+			origin: OriginFor<T>,
+			identifier: T::Identifier,
+			proof: VersionedIdentityProofOf<T>,
+			call: Box<<T as Config>::RuntimeCall>,
+		) -> DispatchResult {
+			let submitter = ensure_signed(origin)?;
+			let proof_digest = IdentityProofs::<T>::get(&identifier).ok_or(Error::<T>::IdentityNotFound)?;
+			let _ = T::ProofVerifier::verify_proof_against_digest(proof, proof_digest)
+				.map_err(|_| Error::<T>::InvalidProof)?;
```
wilwade, line 132: I don't see the call as part of the proof verification. Shouldn't the proof or something check to make sure that the call is not someone front running and actually requested by the DID owner?

ntn-x2: Hey, yes, that's in the making! This is the skeleton PR. There's no logic yet in here, but there will be a trait implemented for the `RuntimeCall` type which dictates which calls require the Dip origin, and if so, which key relationship must be used in the signature.
We already use it in our blockchain to define which calls require a DID origin and which don't -> https://github.com/KILTprotocol/kilt-node/blob/e8626944e58585e9ee26ad4e68ddc30413af717f/runtimes/peregrine/src/lib.rs#L996.

ntn-x2: Also DID signature verification is not yet in there, but it would work similarly to what we have on our chain in the `submit_did_call` extrinsic: https://github.com/KILTprotocol/kilt-node/blob/e8626944e58585e9ee26ad4e68ddc30413af717f/pallets/did/src/lib.rs#L1037.

So the receiver chain would keep track of the last used nonce, so that it's not possible to replay actions nor front-run anything if one action is dependent on another (i.e., the first nonce has to be 1 less than the next one).

wilwade: I see. Cool! Thanks!

pallets/pallet-dip-receiver/src/origin.rs
```diff
@@ -0,0 +1,89 @@
...
+where
+	OuterOrigin: From<DipOrigin<Identifier, AccountId>> + Into<Result<DipOrigin<Identifier, AccountId>, OuterOrigin>>,
+	Identifier: From<[u8; 32]>,
+	AccountId: From<[u8; 32]>,
+{
+	Ok(OuterOrigin::from(DipOrigin {
+		identifier: Identifier::from([0u8; 32]),
+		account_address: AccountId::from([0u8; 32]),
+	}))
+}
```
weichweich, line 55: why not move this into the trait?

pallets/pallet-dip-receiver/src/traits.rs
```diff
@@ -0,0 +1,52 @@
...
+	type Error;
+
+	fn verify_proof_against_digest(
+		proof: VersionedIdentityProof<Self::LeafKey, Self::LeafValue>,
+		digest: Self::ProofDigest,
+	) -> Result<Self::VerificationResult, Self::Error>;
+}
+
+// Always returns success.
+pub struct SuccessfulProofVerifier<ProofDigest, LeafKey, LeafValue>(PhantomData<(ProofDigest, LeafKey, LeafValue)>);
```
weichweich, line 36: this should probably only be used in tests and benchmarks?

weichweich (commented):
Nice work.

My thoughts on this:

* there are a few unrelated changes, would be great to move them to a separate PR
* I don't understand the `ReceiverParachainDipReceiverCalls` and `ReceiverParachainCalls`
* I feel like Receiver and Sender are not the best wordings here?
  * Isn't Provider and relying-party what openID uses? I feel like this would describe it better? In the end sending and receiving is only done to read the storage. If we use another strategy to access the parachains storage, this wouldn't be sending/receiving anymore
* Is this future proof for new technology?
  * The consensus in the community seems to be, that storage proofs should be used to read another chains storage. Here we use XCM. We should at least make sure we can switch later to the other mechanism?

dip-template/runtimes/dip-sender/src/dip.rs
```diff
@@ -0,0 +1,70 @@
...
+use pallet_dip_sender::traits::{
+	DefaultIdentityProofGenerator, DefaultIdentityProvider, TxBuilder, XcmRouterDispatcher,
+};
+use xcm::{latest::MultiLocation, DoubleEncoded};
+
+use crate::{DidIdentifier, Runtime, RuntimeEvent, XcmRouter};
+
+#[derive(Encode, Decode)]
+enum ReceiverParachainCalls {
+	#[codec(index = 50)]
```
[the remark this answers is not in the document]
ntn-x2: That's the index of the dip-receiver pallet on the receiver chain. I think we can make this dynamic in the future, but right now the information has to be statically known to compose the `Transact` call.

dip-template/runtimes/dip-sender/src/dip.rs
```diff
@@ -0,0 +1,70 @@
...
+	DipReceiver(ReceiverParachainDipReceiverCalls),
+}
+
+#[derive(Encode, Decode)]
+enum ReceiverParachainDipReceiverCalls {
+	#[codec(index = 0)]
```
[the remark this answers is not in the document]
ntn-x2: True, but I think it's better to make it explicit, in case this changes in the future. It does not hurt really 😆

ntn-x2:
> Isn't Provider and relying-party what openID uses? I feel like this would describe it better? In the end sending and receiving is only done to read the storage. If we use another strategy to access the parachains storage, this wouldn't be sending/receiving anymore

Gotcha. I will rename them everywhere. In order to avoid merge hell, I [made a ticket](https://github.com/KILTprotocol/ticket/issues/2588) and will tackle this in the future. But your is a valid point, as other people have had the same confusion.

> Is this future proof for new technology?

As mentioned elsewhere, versioning is not completely baked in here, and will be part of a different ticket, most likely after the other big feature milestone have been reached, e.g., call-based verification and actual DID signature verification. But we already have a ticket for versioning, where we can discuss how future-proof the implementation will be, at that point.

dip-template/runtimes/dip-sender/src/dip.rs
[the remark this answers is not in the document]
weichweich: ah so you basically recreate the RuntimeCall enum?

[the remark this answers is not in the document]
ntn-x2: Yes, only for the relevant pallet(s). There are talks on how this can be improved, e.g., [here](https://github.com/paritytech/polkadot/issues/5631). But for now this seems to be the only way.

pallets/pallet-dip-receiver/src/traits.rs
weichweich: in some pallets with have a mock feature for that

pallets/pallet-dip-receiver/src/origin.rs
weichweich: but you have the same trait bounds on the trait? I don't see the difference.

pallets/pallet-dip-receiver/src/traits.rs
ntn-x2: It is used in the actual runtime template, so we can't use any features. It will be deprecated with the new merkle proof PR, but here it's used here: https://github.com/KILTprotocol/kilt-node/pull/489/files#diff-6c40e72378ea467c86621567df431cc2dabb74d5b76ef2a25fa43a16cf568e00R32

pallets/pallet-dip-receiver/src/origin.rs
ntn-x2: Yes, you are right! 😁 I used a similar but apparently for a different reason in the past. In this case it does not make sense. Fixed in https://github.com/KILTprotocol/kilt-node/pull/489/commits/4d8f14d46af5fc40a164471809dee81cc87c6a99.

weichweich (approved):

PR #497, opened 2023-04-04
ci: update action in gh wf that was still using node12

rflechtner:
## re/ KILTProtocol/ticket#2264

See ticket.

## How to test:
 
If tests pass we're good!

## Checklist:


weichweich (approved):
Very well done. love it

Issue #495, opened 2023-04-03
[AUTOMATIC] Update Polkadot dependencies from 0.9.38 to 0.9.39


kiltbot:
### Track updates
```
repository: paritytech/substrate
commits   : 83
command   : subalfred track-updates paritytech/substrate --from polkadot-v0.9.38 --to polkadot-v0.9.39
```
> https://github.com/paritytech/substrate/compare/polkadot-v0.9.38...polkadot-v0.9.39
### All
- [Fee and tip represented as asset ID inside `AssetTxFeePaid`](https://github.com/paritytech/substrate/pull/13083) - `A0-please_review`, `C1-low`, `D3-trivial 🧸`
- [Move beefy-merkle-tree to utils/binary-merkle-tree and make it generic](https://github.com/paritytech/substrate/pull/13076) - `A0-please_review`, `B0-silent`, `C1-low`, `D2-notlive 💤`
- [Metadata V15: Derive `TypeInfo` for describing runtime types](https://github.com/paritytech/substrate/pull/13272) - `A0-please_review`, `B0-silent`, `C1-low`, `D3-trivial 🧸`
- [benchmarks: `EnsureRankedMember` must add ranked members](https://github.com/paritytech/substrate/pull/13297) - `A0-please_review`, `B0-silent`, `C1-low`, `D3-trivial 🧸`
- [Minor: Update output validity tests](https://github.com/paritytech/substrate/pull/13190) - `A0-please_review`, `B0-silent`, `C1-low`, `D3-trivial 🧸`
- [BEEFY: define on-chain beefy-genesis and use it to coordinate voter initialization](https://github.com/paritytech/substrate/pull/13215) - `A0-please_review`, `B0-silent`, `C1-low`, `D2-notlive 💤`
- [feat: add event SkepticsChosen event in society](https://github.com/paritytech/substrate/pull/13291) - `A0-please_review`, `C1-low`, `D3-trivial 🧸`
- [feat(client): significantly increase wasm instance limits](https://github.com/paritytech/substrate/pull/13298) - `A0-please_review`, `B0-silent`, `C1-low`, `D3-trivial 🧸`
- [Remove in-tree bounded types and use bounded-collections crate](https://github.com/paritytech/substrate/pull/13243) - `A0-please_review`, `B0-silent`, `C1-low`, `D3-trivial 🧸`
[...]

PR #500, opened 2023-04-05
chore: update Polkadot dependencies to 0.9.39

ntn-x2:
Fixes https://github.com/KILTprotocol/kilt-node/issues/495.

Since the `wasmtime` dependency has been bumped from 1.x to 6.x, the old nightly was not good anymore since some inline stuff the new crate is using was still not stable back then. Hence, I bumped the nightly to a more recent version. Nevertheless, it cannot be TOO recent because of [this issue](https://github.com/paritytech/substrate/issues/13636) (which maybe has been fixed in 0.9.40 @weichweich?).

Anyway, updating to the new toolchain version added a whole bunch of Clippy warnings which I also addressed in this PR, among which there was one about the wrong declaration of the `parity-scale-codec` package, which I have now used with its original name everywhere, instead of aliasing it to `codec`.


runtimes/peregrine/Cargo.toml
```diff
@@ -70,7 +70,7 @@ pallet-indices.workspace = true
 pallet-membership.workspace = true
 pallet-preimage.workspace = true
 pallet-proxy.workspace = true
-pallet-randomness-collective-flip.workspace = true
+pallet-insecure-randomness-collective-flip.workspace = true
```
weichweich, line 73: This pallet is not needed here. I think the renaming should also make it clear that we should remove it. :D

runtimes/spiritnet/src/lib.rs
```diff
@@ -916,7 +919,7 @@ construct_runtime! {
 		UncheckedExtrinsic = UncheckedExtrinsic,
 	{
 		System: frame_system = 0,
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip = 1,
+		RandomnessCollectiveFlip: pallet_insecure_randomness_collective_flip = 1,
```
weichweich, line 922: Let's remove this please

crates/assets/src/asset.rs
```diff
@@ -456,7 +456,7 @@ pub mod v1 {
 			check_identifier_length_bounds(input)?;
 
 			input.iter().try_for_each(|c| {
-				if !(b'0'..=b'9').contains(c) {
+				if !c.is_ascii_digit() {
```
weichweich, line 459: love it!

weichweich (changes requested):
Let's remove the insecure pallet

crates/assets/src/asset.rs
```diff
@@ -34,32 +34,32 @@ pub mod v1 {
 
 	/// The minimum length, including separator symbols, an asset ID can have
 	/// according to the minimum values defined by the CAIP-19 definition.
-	pub const MINIMUM_ASSET_ID_LENGTH: usize = MINIMUM_NAMESPACE_LENGTH + 1 + MINIMUM_REFERENCE_LENGTH;
+	pub const MINIMUM_ASSET_ID_LENGTH: usize = MINIMUM_ASSET_NAMESPACE_LENGTH + 1 + MINIMUM_ASSET_REFERENCE_LENGTH;
```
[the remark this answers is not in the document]
ntn-x2: Clippy was complaining that the `asset` and `chain` modules are both exporting a constant with the same name.

runtimes/peregrine/Cargo.toml
ntn-x2: Removed it with a migration: https://github.com/KILTprotocol/kilt-node/pull/500/commits/6187dc92d833bc35c144b9603eb8e46b28e83bca

runtimes/spiritnet/src/lib.rs
ntn-x2: Same as above, plus a cleanup PR: https://github.com/KILTprotocol/kilt-node/pull/500/commits/ac9ab1fed82794c685a47a16e26ebd4ddbcc7002

runtimes/peregrine/src/weights/pallet_democracy.rs
```diff
@@ -390,6 +390,24 @@ impl<T: frame_system::Config> pallet_democracy::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().reads(2))
 			.saturating_add(T::DbWeight::get().writes(2))
 	}
+	fn clear_external_metadata() -> Weight {
+		unimplemented!()
```
weichweich, line 394: can we not do this? There was a team that had a stalled chain because they unintentionally pushed code like this to the chain. 😓

ntn-x2: Oh yeah? Which one was that? I changed that to return `Weight::MAX` so that if we forget to benchmark it for some reason, none of the new txs will be included in a block and the chain won't stall -> https://github.com/KILTprotocol/kilt-node/pull/500/commits/ac590deb38eb0f0f34cde55b586a9d725eb7a331.

weichweich: I would simply run the benchmarks on the compile server for this specific pallet. IMHO that should be fine?

ntn-x2: Ok will do. I just don't see how is it different to give completely off but benchmarked weights or give it a weight that would prevent this tx from being included. If anything, I think the latter option is safer 😄

weichweich (approved):

PR #501, opened 2023-04-11
refactor: Improve comments 

Ad96el:
## fixes [KILTProtocol/ticket#312](https://github.com/KILTprotocol/ticket/issues/2621)
Clean up comments. 
Rearrange construct_runtime macro.


runtimes/peregrine/src/lib.rs
```diff
@@ -936,29 +936,27 @@ construct_runtime! {
...
 		TechnicalCommittee: pallet_collective::<Instance2> = 32,
-		// placeholder: parachain council election = 33,
+		// Placeholder: parachain council election = 33,
 		TechnicalMembership: pallet_membership::<Instance1> = 34,
 		Treasury: pallet_treasury = 35,
 		// DELETED: RelayMigration: pallet_relay_migration::{Pallet, Call, Storage, Event<T>} = 36,
 		// DELETED: DynFilter: pallet_dyn_filter = 37,
 
-		// Utility module.
+		// A stateless module with helpers, to allow transmitting batch of calls
```
ntn-x2, line 948: I think we can get rid of this comment altogether. Runtime developers know what this module does.

runtimes/peregrine/src/lib.rs
```diff
@@ -936,29 +936,27 @@ construct_runtime! {
 		Authorship: pallet_authorship::{Pallet, Storage} = 20,
 		AuraExt: cumulus_pallet_aura_ext = 24,
 
-		// Governance stuff
 		Democracy: pallet_democracy = 30,
 		Council: pallet_collective::<Instance1> = 31,
 		TechnicalCommittee: pallet_collective::<Instance2> = 32,
-		// placeholder: parachain council election = 33,
+		// Placeholder: parachain council election = 33,
```
ntn-x2, line 942: Do we need this here? Cc @weichweich

runtimes/peregrine/src/lib.rs
```diff
@@ -936,29 +936,27 @@ construct_runtime! {
...
 		Utility: pallet_utility = 40,
 
 		// Vesting. Usable initially, but removed once all vesting is finished.
 		Vesting: pallet_vesting = 41,
 
-		// System scheduler.
 		Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>} = 42,
 
-		// Proxy pallet.
+		// Allowing accounts to give permission to other accounts to dispatch types of calls from their signed origin
```
ntn-x2, line 956: Same for this. I would say we keep comments only for pallets outside of FRAME, if any.

runtimes/peregrine/src/lib.rs
```diff
@@ -936,29 +936,27 @@ construct_runtime! {
...
 		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>} = 43,
 
-		// Preimage registrar
+		// Preimage pallet allows the storage of large bytes blob
```
ntn-x2, line 959: Same.

runtimes/peregrine/src/lib.rs
weichweich: yeah at some point we want to have another election strategy. index 33 is reserved for that. So maybe replace `placeholder` with `reserved`.

runtimes/peregrine/src/lib.rs
weichweich: I think i like it. 😁 Might also be interesting for not so experienced devs.
 
```suggestion
		// A stateless pallet with helper extrinsics (batch extrinsics, send from different origins, ...)
```

weichweich (commented):
I think the more verbose comments don't hurt here.

weichweich (approved):

PR #502, opened 2023-04-12
feat: add DID key verification relationship logic

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2587.

This PR adds a new trait `DipCallOriginFilter` that exposes a `fn check_proof(call: Call, proof: Self::Proof) -> Result<Self::Success, Self::Error>` function, called by the runtime to verify if a call can be dispatched with the provided proof or not.

### DIFF
--- a/dip-template/runtimes/dip-receiver/src/dip.rs
+++ b/dip-template/runtimes/dip-receiver/src/dip.rs
@@ -47,7 +47,10 @@ impl DipCallOriginFilter for DipCallFilter {
 	type Proof = VerificationResult<BlockNumber>;
 	type Success = ();
 
+	// Accepts only a DipOrigin for the DidLookup pallet calls.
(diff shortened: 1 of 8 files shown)


dip-template/runtimes/dip-receiver/src/dip.rs
```diff
@@ -31,3 +37,90 @@ impl pallet_dip_receiver::Config for Runtime {
...
+		.iter()
+		.skip(1)
+		.map(derive_verification_key_relationship)
+		.try_fold(first_call_relationship, |acc, next| {
+			if next == Some(acc) {
+				Ok(acc)
+			} else {
+				Err(())
+			}
+		})
```
weichweich, line 66: FYI This could get simplified with reduce once it's not experimental... https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_reduce

weichweich (approved):

PR #506, opened 2023-04-18
fix: deposit wrong calculated

Ad96el:


weichweich (approved):
Good catch. 😓

ntn-x2 (approved):
This makes much more sense, so we can probably bump the service endpoints to 2 then 😄 Good to fix it before it is included in a release.

PR #507, opened 2023-04-19
feat: dynamic deposit for did creation

Ad96el:
## fixes [2636](https://github.com/KILTprotocol/ticket/issues/2636)

The basic design idea can be found [here](https://docs.google.com/document/d/1vPRr3E7AzwRBuQJ54TJh48U-ZxXJkZGp2rGBdjxUWoU/edit). 

The runtimeAPI call to get an approximation for the deposit will be developed in a different PR.


pallets/did/src/lib.rs
```diff
@@ -846,14 +862,24 @@ pub mod pallet {
 		#[pallet::call_index(8)]
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::add_service_endpoint())]
 		pub fn add_service_endpoint(origin: OriginFor<T>, service_endpoint: DidEndpoint<T>) -> DispatchResult {
+			let sender = ensure_signed(origin.clone())?;
 			let did_subject = T::EnsureOrigin::ensure_origin(origin)?.subject();
 
+			// Check the free balance before we do any work
```
Ad96el, line 868: Should I add this query everywhere possible?

runtimes/common/src/constants.rs
```diff
@@ -324,7 +324,16 @@ pub mod did {
...
 	pub const MAX_DID_BYTE_LENGTH: u32 = 9918;
 
-	pub const DID_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, MAX_DID_BYTE_LENGTH);
+	/// The length of all used Keys.
+	pub const KEY_DID_BYTE_LENGTH: u32 = 32;
+
+	pub const MAX_SERVICE_ENDPOINT_BYTE_LENGTH: u32 =
+		MAX_SERVICE_URL_LENGTH + MAX_SERVICE_ID_LENGTH + MAX_SERVICE_TYPE_LENGTH;
+
+	pub const DID_BASE_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, MAX_DID_BYTE_LENGTH);
```
weichweich, line 333: Not sure if we should take the MAX_DID_BYTE_LENGHT here. Maybe we should calculate the `MIN_DID_DEPOSIT` and 2 KILT and always take the maximum for the `DID_BASE_DEPOSIT`

weichweich (changes requested):
Looks good! But there might be some bugs and I'm not 100% sure how to do the base deposit.

pallets/did/src/lib.rs
```diff
@@ -556,12 +570,17 @@ pub mod pallet {
...
+			DidEndpointsCount::<T>::insert(&did_identifier, input_service_endpoints.len().saturated_into::<u32>());
+
+			let mut did_entry =
 				DidDetails::from_creation_details(*details, account_did_auth_key).map_err(Error::<T>::from)?;
 
-			// *** No Fail beyond this call ***
+			Did::<T>::insert(&did_identifier, did_entry.clone());
 
-			CurrencyOf::<T>::reserve(&did_entry.deposit.owner, did_entry.deposit.amount)?;
+			did_entry.update_deposit(&did_identifier)?;
```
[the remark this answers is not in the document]
Ad96el: True. It is [resolved](https://github.com/KILTprotocol/kilt-node/pull/507/commits/426e9b220ad5568ac2b19c53950a5e89022fbdfe). 

I also added a test for checking, how the deposit has to change by adding a new service endpoint. I think the rest of the implemented logic are implicitly tested by the old tests.

pallets/did/src/lib.rs
```diff
@@ -915,6 +934,8 @@ pub mod pallet {
 				}
 			});
 
+			did_details.update_deposit(&did_subject)?;
```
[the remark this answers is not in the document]
Ad96el: [Resolved](https://github.com/KILTprotocol/kilt-node/pull/507/commits/426e9b220ad5568ac2b19c53950a5e89022fbdfe)

runtimes/common/src/constants.rs
Ad96el: Could you provide more information, please? The `MIN_DID_DEPOSIT` would be around 0.05 KILTS. I do not fully understand that point.

pallets/did/src/mock.rs
```diff
@@ -141,10 +146,12 @@ impl Config for Test {
 	type RuntimeOrigin = RuntimeOrigin;
 	type RuntimeCall = RuntimeCall;
 	type EnsureOrigin = EnsureSigned<DidIdentifier>;
+	type DepositKey = DepositKey;
+	type DepositServiceEndpoint = DepositKey;
```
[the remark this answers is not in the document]
Ad96el: Since these are mock data, I wanted to make it easy for me :laughing:. I will make it as accurate as in the runtimes.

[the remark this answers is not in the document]
weichweich: ah no

pallets/did/src/lib.rs
```diff
@@ -218,10 +217,23 @@ pub mod pallet {
...
+		/// deposit can be reclaimed when the DID is deleted.
+		#[pallet::constant]
+		type BaseDeposit: Get<BalanceOf<Self>>;
+
+		/// The amount of balance that will be taken for each service endpoint
+		/// as a deposit to incentivise fair use of the on chain storage. The
+		/// deposit can be reclaimed when the service endpoint is removed or the
+		/// DID deleted.
+		#[pallet::constant]
+		type DepositServiceEndpoint: Get<BalanceOf<Self>>;
```
ntn-x2, line 231: Can we call it in the other way? `ServiceEndpointDeposit`?

pallets/did/src/lib.rs
```diff
@@ -218,10 +217,23 @@ pub mod pallet {
...
+
+		/// The amount of balance that will be taken for each added key as a
+		/// deposit to incentivise fair use of the on chain storage.
 		#[pallet::constant]
-		type Deposit: Get<BalanceOf<Self>>;
+		type DepositKey: Get<BalanceOf<Self>>;
```
ntn-x2, line 236: Also here, `KeyDeposit`.

pallets/did/src/utils.rs
```diff
@@ -16,17 +16,45 @@
...
+	did_details::{DidCreationDetails, DidPublicKey},
+	BalanceOf, Config, KeyIdOf,
+};
 use parity_scale_codec::Encode;
-use sp_runtime::traits::Hash;
+use sp_core::Get;
+use sp_runtime::traits::{Hash, Zero};
 use sp_std::vec::Vec;
 
-use crate::{did_details::DidPublicKey, Config, KeyIdOf};
```
ntn-x2, line 23: We usually keep the internal imports at the bottom, separated from the rest by a newline. I think we should keep that.

pallets/did/src/utils.rs
```diff
@@ -16,17 +16,45 @@
...
 	T::Hashing::hash(&hashed_values)
 }
 
+pub fn calculate_deposit<T: Config>(details: &DidCreationDetails<T>) -> BalanceOf<T>
+where
+	BalanceOf<T>: From<u32>,
+{
+	let mut deposit: BalanceOf<T> = T::BaseDeposit::get();
+
+	let count_service_endpoint: BalanceOf<T> = (details.new_service_details.len() as u32).into();
```
ntn-x2, line 39:
```suggestion
	let service_endpoint_count: BalanceOf<T> = (details.new_service_details.len() as u32).into();
```

pallets/did/src/utils.rs
```diff
@@ -16,17 +16,45 @@
...
+	deposit += count_service_endpoint * T::DepositServiceEndpoint::get();
+
+	let count_key_agreements: BalanceOf<T> = (details.new_key_agreement_keys.len() as u32).into();
```
ntn-x2, line 42:
```suggestion
	let key_agreements_count: BalanceOf<T> = (details.new_key_agreement_keys.len() as u32).into();
```

pallets/did/src/utils.rs
```diff
@@ -16,17 +16,45 @@
...
 }
 
+pub fn calculate_deposit<T: Config>(details: &DidCreationDetails<T>) -> BalanceOf<T>
+where
+	BalanceOf<T>: From<u32>,
+{
+	let mut deposit: BalanceOf<T> = T::BaseDeposit::get();
+
+	let count_service_endpoint: BalanceOf<T> = (details.new_service_details.len() as u32).into();
+	deposit += count_service_endpoint * T::DepositServiceEndpoint::get();
```
ntn-x2, line 40: Here also we use `saturating_add`, and `saturating_mul`.

pallets/did/src/utils.rs
```diff
@@ -16,17 +16,45 @@
...
 use sp_std::vec::Vec;
 
-use crate::{did_details::DidPublicKey, Config, KeyIdOf};
-
 pub fn calculate_key_id<T: Config>(key: &DidPublicKey) -> KeyIdOf<T> {
 	let hashed_values: Vec<u8> = key.encode();
 	T::Hashing::hash(&hashed_values)
 }
 
+pub fn calculate_deposit<T: Config>(details: &DidCreationDetails<T>) -> BalanceOf<T>
```
ntn-x2, line 33: Because we still create `DidDetails` from the creation details, perhaps it makes more sense to have this function only on `DidDetails` and, for creations, calculate the deposit after craeting the `DidDetails` instance, so we avoid code duplication.

ntn-x2: TL;DR: I would remove this function from here and only keep it for `DidCreationDetails`.

pallets/did/src/mock.rs
```diff
@@ -23,6 +23,7 @@ use frame_support::{
 };
 use frame_system::EnsureSigned;
 use pallet_balances::NegativeImbalance;
+
```
ntn-x2, line 26: Why these new lines that I see often? Are you following some sort of convention?

pallets/did/src/did_details.rs
```diff
@@ -344,7 +413,6 @@ impl<T: Config> DidDetails<T> {
 		if let Some(delegation_key) = details.new_delegation_key {
 			new_did_details.update_delegation_key(delegation_key, current_block_number)?;
 		}
```
ntn-x2, line 412: Can we maybe update the deposit inside here, once we set all the keys by calling `calculate_deposit`?

pallets/did/src/lib.rs
```diff
@@ -556,12 +570,17 @@ pub mod pallet {
...
+			input_service_endpoints.iter().for_each(|service| {
+				ServiceEndpoints::<T>::insert(&did_identifier, &service.id, service.clone());
+			});
+			DidEndpointsCount::<T>::insert(&did_identifier, input_service_endpoints.len().saturated_into::<u32>());
+
+			let mut did_entry =
 				DidDetails::from_creation_details(*details, account_did_auth_key).map_err(Error::<T>::from)?;
 
-			// *** No Fail beyond this call ***
+			did_entry.update_deposit(&did_identifier)?;
```
ntn-x2, line 585: This could be done inside the function already, it's probably safer.

ntn-x2 (changes requested):
I like the approach, which minimally changes only what needs to change 😄
Nevertheless there are few points I highlighted:

- There could be less code duplication when it comes to deposit calculation
- We should aim to have self-contained functions that take care of everything needed, instead of letting part of the task be taken care of by the caller
- Replacing all occurrences of `+`, `-`, `*`, and `/` with their saturating alternatives (most places), or checked ones, where it makes sense to fail instead of saturate.

pallets/did/src/utils.rs
weichweich: We should consider this lint: https://rust-lang.github.io/rust-clippy/rust-1.68.0/index.html#as_conversions

weichweich: Or this: https://rust-lang.github.io/rust-clippy/rust-1.68.0/index.html#cast_possible_wrap

pallets/did/src/utils.rs
Ad96el: May I ask what kind of formatter do you use?

Ad96el:
All requested changes should be resolved now

pallets/did/src/utils.rs
ntn-x2: This is not a formatter rule, it's just a convention we use. Crate imports always come after external imports, and to avoid the formatter to shuffle them around alphabetically, we put a new line.

weichweich (changes requested):
I think the benchmarks might not be accurate.

pallets/did/src/did_details.rs
```diff
@@ -642,6 +639,30 @@ pub struct DidCreationDetails<T: Config> {
 	pub new_service_details: Vec<DidEndpoint<T>>,
 }
 
+impl<T: Config> DidCreationDetails<T> {
+	pub fn calculate_deposit(&self) -> BalanceOf<T> {
```
ntn-x2, line 642: Do we still need this if we have the same function for `DidDetails`?

pallets/did/src/did_details.rs
```diff
@@ -318,44 +318,38 @@ impl<T: Config> DidDetails<T> {
...
+			Some(_) => T::KeyDeposit::get(),
 			_ => Zero::zero(),
-		};
+		});
 
 		deposit
 	}
 
-	fn reserve_deposit(&self, _deposit_to_reserve: BalanceOf<T>) -> Result<(), DispatchError> {
+	fn reserve_deposit(_deposit_to_reserve: BalanceOf<T>, who: AccountIdOf<T>) -> Result<(), DispatchError> {
```
ntn-x2: I guess we don't need the leading `_` for `_deposit_to_reserve`?

pallets/did/src/lib.rs
```diff
@@ -427,6 +427,7 @@ pub mod pallet {
 				DidError::Signature(operation_error) => Self::from(operation_error),
 				DidError::Input(input_error) => Self::from(input_error),
 				DidError::Internal => Self::Internal,
+				DidError::DepositError(_) => Self::UnableToPayFees,
```
ntn-x2: Should it only be `DidError::Deposit`?

pallets/did/src/did_details.rs
ntn-x2: Is the heavy work you are referring to the one that checks the validity of the input? If that's the case, it does not matter much, since we take fees for that anyway, based on the size of the input, and we never return any surplus in case we finish early (e.g., if the user has not enough funds to pay for the deposit). Hence I would say it makes sense to avoid duplication and remove the code to calculate the deposit from the creation details, and simply call it on the created `DidDetails` inside the creation operation.
Alternatively, you could also simply create the details from the creation, calculate and check for the deposit, then run all the checks.
But as I said, the extrinsic is benchmarked depending on the input, so it does not really matter if we return early from the function or not.

runtimes/common/src/constants.rs
```diff
@@ -324,7 +324,18 @@ pub mod did {
...
+	/// The length of all used Keys.
+	pub const KEY_DID_BYTE_LENGTH: u32 = 32;
+
+	pub const MAX_SERVICE_ENDPOINT_BYTE_LENGTH: u32 =
+		MAX_SERVICE_URL_LENGTH + MAX_SERVICE_ID_LENGTH + MAX_SERVICE_TYPE_LENGTH;
+
+	pub const DID_BASE_DEPOSIT: Balance = 2 * KILT;
+	pub const KEY_DEPOSIT: Balance = deposit(0, KEY_DID_BYTE_LENGTH);
+	pub const SERVICE_ENDPOINT_DEPOSIT: Balance = deposit(1, MAX_SERVICE_ENDPOINT_BYTE_LENGTH);
+	pub const MAX_DEPOSIT_DID: Balance =
```
ntn-x2, line 336: Is this accurate? Do we actually have 2 + number of services elements, each of size `MAX_DID_BYTE_LENGTH`?

pallets/did/src/lib.rs
```diff
@@ -857,6 +865,9 @@ pub mod pallet {
 				},
 			)?;
 			DidEndpointsCount::<T>::insert(&did_subject, currently_stored_endpoints_count.saturating_add(1));
+			did.update_deposit(&did_subject)?;
```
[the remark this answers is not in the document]
Ad96el: In general, the `update_deposit` function must be called first, and then the record has to be saved. Otherwise, the deposit attribute will not be updated, and the state is not correct.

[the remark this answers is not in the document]
Ad96el: In general, the `update_deposit` function must be called first. Otherwise, the deposit is not updated in the `did_details`.

pallets/did/src/lib.rs
```diff
@@ -556,10 +561,15 @@ pub mod pallet {
 			service_endpoints_utils::validate_new_service_endpoints(&input_service_endpoints)
 				.map_err(Error::<T>::from)?;
 
-			let did_entry =
-				DidDetails::from_creation_details(*details, account_did_auth_key).map_err(Error::<T>::from)?;
+			input_service_endpoints.iter().for_each(|service| {
```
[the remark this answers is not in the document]
Ad96el: The deposit is based on the saved service endpoints in the storage. If I calculate the deposit before the endpoints are saved, the deposit is wrong. What I can do, is updating the deposit not inside the `from_creation_details` and calling the `update_deposit` function before saving the record.

pallets/did/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Is this not currently the case? In each call, the `update_deposit` function is executed, then the mutated record is saved. :thinking:

runtimes/common/src/constants.rs
Ad96el: The value MAX_DEPOSIT_DID consists of the old did deposit, which was the worst case, plus the new base deposit. 

The max byte length and the amount of storage entries is still the same.

runtimes/common/src/constants.rs
```diff
@@ -324,7 +324,18 @@ pub mod did {
 	/// The size is checked in the runtime by a test.
 	pub const MAX_DID_BYTE_LENGTH: u32 = 9918;
 
-	pub const DID_DEPOSIT: Balance = deposit(2 + MAX_NUMBER_OF_SERVICES_PER_DID, MAX_DID_BYTE_LENGTH);
+	/// The length of all used Keys.
+	pub const KEY_DID_BYTE_LENGTH: u32 = 32;
```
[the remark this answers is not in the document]
Ad96el: Yes exactly 👯

runtimes/common/src/constants.rs
Ad96el: It should be accurate, since it is the old deposit + the new base deposit. MAX_DID_BYTE_LENGTH is the worst case of a single DID. We have 27 different storage entries. 1 for the did_details and 1 for the service endpoint count, if I am not wrong 😃

pallets/did/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Same as above. The new service endpoint has to be in the storage, otherwise the update_deposit function would calculate the old deposit value.

runtimes/common/src/constants.rs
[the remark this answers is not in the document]
ntn-x2: Could you maybe rename the const as well? 😁

runtimes/common/src/constants.rs
ntn-x2: Yes, I see that now only the base deposit is added, which looks correct. I am just wondering about the nature of the `deposit` function:

```
/// Calculate the storage deposit based on the number of storage items and the
/// combined byte size of those items.
```

I understood it that for each item type, we would count how much space they take and how many entries we write in storage. But reading it again now, it says **combined**, so I guess items are to be counted separately, but the grand total is the total size. In that case, this is correct ☺️

pallets/did/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: Oh right, cause `DidDetails` does not have info about service endpoints!

ntn-x2 (approved):
One last rename, and then it's good to go for me! 👌

PR #512, opened 2023-04-26
chore: add sanity tests

Ad96el:

### DIFF
--- a/pallets/did/src/lib.rs
+++ b/pallets/did/src/lib.rs
@@ -463,6 +463,15 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		#[cfg(feature = "try-runtime")]
+		fn try_state(_n: BlockNumberFor<T>) -> Result<(), &'static str> {
+			do_try_state()?;
+			Ok(())
+		}
+	}
+
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// Store a new DID on chain, after verifying that the creation
@@ -1224,6 +1233,52 @@ pub mod pallet {
 			log::debug!("Deleting DID {:?}", did_subject);
 
 			Self::deposit_event(Event::DidDeleted(did_subject));
+
+			Ok(())
+		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			Did::<T>::iter().try_for_each(
+				|(did_subject, did_details): (DidIdentifierOf<T>, DidDetails<T>)| -> DispatchResult {
+					let service_endpoints_count: usize = ServiceEndpoints::<T>::iter_prefix(&did_subject).count();
+
+					ensure!(
+						service_endpoints_count == DidEndpointsCount::<T>::get(&did_subject).saturated_into::<usize>(),
+						DispatchError::Other("Storage Error: amount of endpoints are not matching!")
+					);
+
+					ensure!(
+						did_details.key_agreement_keys.len()
+							> <T as Config>::MaxTotalKeyAgreementKeys::get().saturated_into::<usize>(),
+						DispatchError::Other("Storage Error: amount of endpoints are not matching!")
+					);
+
+					ensure!(
+						service_endpoints_count
+							< <T as Config>::MaxNumberOfServicesPerDid::get().saturated_into::<usize>(),
+						DispatchError::Other("Stoarge Error: amount of endpoints is to high!")
+					);
+
+					ensure!(
+						!DidBlacklist::<T>::contains_key(did_subject),
+						DispatchError::Other("Deleted DID is still in storage!")
+					);
+
+					Ok(())
+				},
+			)?;
+
+			DidBlacklist::<T>::iter_keys().try_for_each(|deleted_did_subject| -> DispatchResult {
+				let service_endpoints_count = ServiceEndpoints::<T>::iter_prefix(&deleted_did_subject).count();
+
+				ensure!(
+					service_endpoints_count == 0,
+					DispatchError::Other("Deleted service enpoints are still in storage!")
+				);
+
+				Ok(())
+			})?;
 
 			Ok(())
 		}
--- a/pallets/did/src/mock.rs
+++ b/pallets/did/src/mock.rs
@@ -40,7 +40,8 @@ use crate::{
 		RelationshipDeriveError,
 	},
 	service_endpoints::DidEndpoint,
-	utils as crate_utils, AccountIdOf, Config, CurrencyOf, DidBlacklist, DidEndpointsCount, KeyIdOf, ServiceEndpoints,
+	utils as crate_utils, AccountIdOf, Config, CurrencyOf, DidBlacklist, DidEndpointsCount, KeyIdOf, Pallet,
+	ServiceEndpoints,
 };
 #[cfg(not(feature = "runtime-benchmarks"))]
 use crate::{DidRawOrigin, EnsureDidOrigin};
@@ -497,6 +498,13 @@ impl ExtBuilder {
 		});
 
 		ext
+	}
+
+	pub fn build_and_execute_with_sanity_tests(self, ext: Option<sp_io::TestExternalities>, test: impl FnOnce() -> ()) {
+		self.build(ext).execute_with(|| {
+			test();
+			Pallet::<Test>::do_try_state().unwrap();
+		})
 	}
 
 	#[cfg(feature = "runtime-benchmarks")]
--- a/pallets/did/src/tests.rs
+++ b/pallets/did/src/tests.rs
@@ -697,8 +697,7 @@ fn check_invalid_service_id_character_did_creation() {
 		+ <<Test as did::Config>::Currency as Currency<did::AccountIdOf<Test>>>::minimum_balance();
 	ExtBuilder::default()
 		.with_balances(vec![(ACCOUNT_00, balance)])
-		.build(None)
-		.execute_with(|| {
+		.build_and_execute_with_sanity_tests(None, || {
 			assert_noop!(
 				Did::create(
 					RuntimeOrigin::signed(ACCOUNT_00),


pallets/attestation/src/lib.rs
```diff
@@ -465,6 +471,22 @@ pub mod pallet {
 				ExternalAttestations::<T>::remove(authorization_id, claim_hash);
 			}
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			Attestations::<T>::iter().try_for_each(|(claim_hash, attestation_details)| -> DispatchResult {
```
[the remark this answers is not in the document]
ntn-x2: We could probably return the error string that the function signature expects. So we know which line failed and which storage entry failed.

pallets/delegation/src/lib.rs
```diff
@@ -87,7 +87,7 @@ use frame_support::{
 use kilt_support::traits::StorageDepositCollector;
 use parity_scale_codec::Encode;
 use sp_runtime::{traits::Hash, DispatchError};
-use sp_std::{marker::PhantomData, vec::Vec};
+use sp_std::{marker::PhantomData, vec, vec::Vec};
```
ntn-x2, line 90: This dependendency is only used in `try-runtime`, right? In this case it should not be pulled in regardless, and also it could either be used directly with `sp_std::vec`, or imported inside the function that needs it.

pallets/delegation/src/lib.rs
```diff
@@ -965,6 +974,58 @@ pub mod pallet {
...
 			Ok((removals, consumed_weight))
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			fn get_merged_subtree<T: Config>(node: DelegationNode<T>) -> Vec<DelegationNode<T>> {
+				let mut nodes_to_explore = vec![node];
+				let mut children: Vec<DelegationNode<T>> = vec![];
+				while !nodes_to_explore.is_empty() {
+					let current_node = nodes_to_explore.pop().unwrap();
```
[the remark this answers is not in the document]
Ad96el: Thats cool :heavy_check_mark:

pallets/delegation/src/lib.rs
```diff
@@ -965,6 +974,58 @@ pub mod pallet {
...
+			}
+
+			DelegationNodes::<T>::iter().try_for_each(
+				|(delegation_node_id, delegation_details): (DelegationNodeIdOf<T>, DelegationNode<T>)| -> DispatchResult {
+					let hierachy_id = delegation_details.hierarchy_root_id;
+
+					// check if node is in part of a delegation hierachy.
+					ensure!(
+						DelegationHierarchies::<T>::contains_key(hierachy_id),
+						DispatchError::Other("Test1")
```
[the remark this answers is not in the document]
ntn-x2: We can simply return a `&'static str`, instead of returning a full error. This applies to all integrity tests.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -421,6 +431,19 @@ pub mod pallet {
...
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			ConnectedDids::<T>::iter().try_for_each(|(account, record)| -> DispatchResult {
+				let is_in_connected_accounts = ConnectedAccounts::<T>::contains_key(record.did, account);
+
+				ensure!(is_in_connected_accounts, DispatchError::Other("Test"));
+
+				Ok(())
+			})?;
```
ntn-x2, line 443: We can also simply return the result of the `try_for_each` without writing the additional `Ok(())` line at the end.

pallets/pallet-web3-names/src/lib.rs
```diff
@@ -496,6 +502,32 @@ pub mod pallet {
 		fn unban_name(name: &Web3NameOf<T>) {
 			Banned::<T>::remove(name);
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			// check if for each owner there is a name stored.
+			Owner::<T>::iter().try_for_each(
```
ntn-x2, line 509: Can we also test the reverse direction? For each entry here, there has to be a reverse entry in `Names`.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1902,14 +2053,20 @@ pub mod pallet {
 					}
 					(false, true) => {
 						// candidate pushed out the least staked collator which is now at position
-						// min(max_selected_top_candidates, top_candidates - 1) in TopCandidates
-						let old_col_idx = max_selected_candidates.min(top_candidates.len().saturating_sub(1));
+						let (drop_self, drop_delegators) = match max_selected_candidates.cmp(&top_candidates.len()) {
```
ntn-x2, line 2056: Why these changes here?

pallets/public-credentials/src/lib.rs
```diff
@@ -585,6 +594,31 @@ pub mod pallet {
 				}
 			})
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> DispatchResult {
+			// check if for each owner there is a name stored.
```
ntn-x2, line 600: Can we also check the reverse?

pallets/public-credentials/src/lib.rs
```diff
@@ -585,6 +594,31 @@ pub mod pallet {
...
+			Credentials::<T>::iter().try_for_each(|(subject_id, credential_id, entry)| -> DispatchResult {
+				ensure!(
+					CredentialSubjects::<T>::contains_key(&credential_id),
+					DispatchError::Other("Test")
+				);
+
+				ensure!(
+					CredentialSubjects::<T>::get(&credential_id).unwrap() == subject_id,
```
ntn-x2, line 608:
```suggestion
					CredentialSubjects::<T>::get(&credential_id) == Some(subject_id),
```

ntn-x2: This is enough, and allows you to remove the check above. This should be applied to all the other places as well.

pallets/parachain-staking/src/lib.rs
Ad96el: This is a bug, which the sanity tests revealed. If the top candidates are not full, we have not to decrease the total stake.

pallets/delegation/src/lib.rs
```diff
@@ -965,6 +974,58 @@ pub mod pallet {
...
+					);
+
+					let children_count = DelegationNodes::<T>::iter_values()
+						.filter(|delegation_node : &DelegationNode<T> | delegation_node.children.contains(&delegation_node_id))
+						.count();
```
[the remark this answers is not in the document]
Ad96el: With this filter, I count the parents of my current delegation node, not its child's. The variable is named terribly. I count how many parents my node has. With this information, I can determine if the graph contains any cycles.

I will also add a test to see if all child nodes exist.

pallets/delegation/src/lib.rs
```diff
@@ -965,6 +974,56 @@ pub mod pallet {
...
+					);
+
+					let parent_count = DelegationNodes::<T>::iter_values()
+						.filter(|delegation_node: &DelegationNode<T>| {
+							delegation_node.children.contains(&delegation_node_id)
+						})
+						.count();
+
+					match delegation_details.parent {
+						// If node is a leaf or intermediate, check if it occurs only once. Otherwise we have cycles.
```
ntn-x2, line 1008: Would an `if let Some(...)` be better here?

pallets/did/src/lib.rs
```diff
@@ -1227,6 +1236,42 @@ pub mod pallet {
...
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> Result<(), &'static str> {
+			Did::<T>::iter().try_for_each(
+				|(did_subject, did_details): (DidIdentifierOf<T>, DidDetails<T>)| -> Result<(), &'static str> {
+					let service_endpoints_count = ServiceEndpoints::<T>::iter_prefix(&did_subject).count();
+
+					ensure!(
+						service_endpoints_count == DidEndpointsCount::<T>::get(&did_subject).saturated_into::<usize>(),
+						"Unequal service endpoint"
```
ntn-x2, line 1248: Count, maybe?

```suggestion
						"Unequal service endpoint count"
```

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -421,6 +430,26 @@ pub mod pallet {
...
+					ConnectedAccounts::<T>::contains_key(record.did, account),
+					"Unknown account"
+				);
+				Ok(())
+			})?;
+
+			ConnectedAccounts::<T>::iter().try_for_each(
+				|(did_identifier, linked_account_id, _)| -> Result<(), &'static str> {
+					let connected_did = ConnectedDids::<T>::get(linked_account_id);
+					ensure!(connected_did.is_some(), "Unknown did");
```
ntn-x2, line 447: This can still be simplified by checking against `Some(...)`.

pallets/pallet-web3-names/src/lib.rs
```diff
@@ -496,6 +502,32 @@ pub mod pallet {
 		fn unban_name(name: &Web3NameOf<T>) {
 			Banned::<T>::remove(name);
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> Result<(), &'static str> {
+			// check if for each owner there is a name stored.
+			Owner::<T>::iter().try_for_each(
+				|(w3n, ownership): (Web3NameOf<T>, Web3OwnershipOf<T>)| -> Result<(), &'static str> {
+					ensure!(Names::<T>::contains_key(ownership.owner.clone()), "W3n unknown");
```
ntn-x2, line 511: This check seems redundant, given the line below, no?

pallets/pallet-web3-names/src/lib.rs
```diff
@@ -496,6 +502,32 @@ pub mod pallet {
...
+					ensure!(Names::<T>::get(ownership.owner) == Some(w3n), "Unequal w3n");
+					Ok(())
+				},
+			)?;
+
+			// check for each name there is an owner.
+			Names::<T>::iter().try_for_each(
+				|(w3n_owner, w3n): (Web3NameOwnerOf<T>, Web3NameOf<T>)| -> Result<(), &'static str> {
+					ensure!(Owner::<T>::contains_key(&w3n), "Owner unknown");
```
ntn-x2, line 520: Same as above.

pallets/attestation/src/lib.rs
```diff
@@ -148,7 +148,13 @@ pub mod pallet {
 	pub struct Pallet<T>(_);
 
 	#[pallet::hooks]
-	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		#[cfg(feature = "try-runtime")]
+		fn try_state(_n: BlockNumberFor<T>) -> Result<(), &'static str> {
+			Self::do_try_state()?;
+			Ok(())
```
ntn-x2, line 155: We don't need the `Ok(())` line if `do_try_state` already returns what we need. Same for other places.

pallets/attestation/src/lib.rs
```diff
@@ -465,6 +471,19 @@ pub mod pallet {
 				ExternalAttestations::<T>::remove(authorization_id, claim_hash);
 			}
 		}
+
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> Result<(), &'static str> {
```
ntn-x2, line 476: These do not need to be public, so making them all crate-visible is the right thing.

```suggestion
		pub(crate) fn do_try_state() -> Result<(), &'static str> {
```

pallets/attestation/src/lib.rs
```diff
@@ -465,6 +471,19 @@ pub mod pallet {
...
+			Attestations::<T>::iter().try_for_each(|(claim_hash, attestation_details)| -> Result<(), &'static str> {
+				if let Some(authorization_id) = attestation_details.authorization_id {
+					ensure!(
+						ExternalAttestations::<T>::get(authorization_id, claim_hash),
+						"Unknown external attestation"
```
ntn-x2, line 481: What about including some more information in the error strings? Something like `format!("External attestation entry does not exist for attestation with authorization ID {:?}", authorization_id)`. This will save us a lot of time in case something does actually break in the future. Probably the same could be applied to most of the error messages.

pallets/attestation/src/mock.rs
```diff
@@ -385,6 +387,13 @@ pub(crate) mod runtime {
 			ext
 		}
 
+		pub fn build_and_execute_with_sanity_tests(self, test: impl FnOnce()) {
+			self.build().execute_with(|| {
+				test();
+				Pallet::<Test>::do_try_state().unwrap();
```
ntn-x2, line 393: Better to use `expect(error_string)` then `unwrap()` in almost all circumstances 😃

pallets/delegation/src/lib.rs
ntn-x2: Or if we don't need the value in the `Some` variant, even just `if delegation_details.parent.is_some()` is enough.

pallets/public-credentials/src/lib.rs
```diff
@@ -585,6 +594,35 @@ pub mod pallet {
...
+		pub fn do_try_state() -> Result<(), &'static str> {
+			Credentials::<T>::iter().try_for_each(
+				|(subject_id, credential_id, entry)| -> Result<(), &'static str> {
+					ensure!(
+						CredentialSubjects::<T>::contains_key(&credential_id),
+						"Unknown credential subject"
+					);
+
+					ensure!(
+						CredentialSubjects::<T>::get(&credential_id) == Some(subject_id),
```
ntn-x2, line 608: This check also is sufficient, and you can remove the previous one since they check for the same thing.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1757,6 +1767,13 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
+		#[cfg(any(feature = "try-runtime", test))]
+		pub fn do_try_state() -> Result<(), &'static str> {
+			crate::try_state::validate_candiate_pool::<T>()?;
```
ntn-x2, line 1772: I would rather like to see a single `try_state()` function exposed, which then calls the other auxiliary functions. It's less burden for someone that is going over this pallet's code to see a single call in here.

ntn-x2 (commented):
A second round of comments! I love the solution that runs integrity tests after every unit test case. Nevertheless, I would still like to see the following:

* Clearer error messages when tests fail, especially (but not only) when inside loops, since otherwise it becomes much harder to find out when the test failed
* A separate `try_state.rs` module for each pallet, with a single `pub(crate) try_state()` function, which then internally calls several different functions, if needed.

Other than that, this looks like a very very very nice improvement, which will be completed once we have cross-pallet checks in place, especially regarding deposits and locked balances.

pallets/pallet-did-lookup/src/lib.rs
Ad96el: Are you sure? I want to check a property of `connected_did`. 

Doing something like: 
```
connected_did == Some(ConnectionRecord {
					did: did_identifier,
					deposit: deposit
				})
```

looks to me over-engineered.

Ad96el: I also do not know the deposit – I needed to copy it from the `connected_did` variable -> I have to unwrap it...

pallets/pallet-web3-names/src/lib.rs
Ad96el: Well, `Names` and `Owner` should have a 1to1 relation. It can happen that a name record is pointing to an owner, which does not exist. The `check_claiming_preconditions` ensures, that the relation is correct, but I think testing it does not hurt.

Ad96el: Owner and names should have a 1 to 1 relation. It might happen, that an owner is pointing to a name which does not exist. Iterating over only one of these storage, would not reveal these issues.

Ad96el:
All requested changes should be resolved now.

pallets/pallet-did-lookup/src/lib.rs
ntn-x2: I see, then probably calling `.expect(...)` on ConnectedDids::<T>::get()` and then matching the DID is a better option, in my opinion. All cases panic anyway.

pallets/pallet-web3-names/src/try_state.rs
```diff
@@ -0,0 +1,45 @@
...
+			ensure!(Names::<T>::get(ownership.owner) == Some(w3n), "Unequal w3n");
+			Ok(())
+		},
+	)?;
+
+	// check for each name there is an owner.
+	Names::<T>::iter().try_for_each(
+		|(w3n_owner, w3n): (Web3NameOwnerOf<T>, Web3NameOf<T>)| -> Result<(), &'static str> {
+			ensure!(Owner::<T>::contains_key(&w3n), "Unknown Owner");
+			ensure!(Owner::<T>::get(w3n).unwrap().owner == w3n_owner, "Unequal owner");
```
ntn-x2, line 36: Here you could use `expect` the web3name, and matching the owner with the expected value.

ntn-x2 (commented):
Apart from the offline discussion about better errors, I just have a couple more suggestions, then it looks good for me!

Ad96el:
@ntn-x2 I am always happy to see your suggestions!

support/src/test.rs
```diff
@@ -0,0 +1,23 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use scale_info::prelude::{boxed::Box, string::String};
```
ntn-x2, line 19: I think taking them from `sp_std` would be cleaner, as that's supposed to be the replacement of the standard library in Substrate.

support/src/lib.rs
```diff
@@ -20,6 +20,9 @@
 pub mod deposit;
 pub use deposit::{free_deposit, reserve_deposit};
 
+#[cfg(any(feature = "try-runtime", test))]
+pub mod test;
```
ntn-x2, line 24: Not sure this is the best nemaing. This is a testing utility, right? Maybe `test_utils` or `test_utilities` would be better? We usually use `test` for actual tests.

pallets/public-credentials/src/try_state.rs
```diff
@@ -0,0 +1,50 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use frame_support::ensure;
+use kilt_support::test::convert_error_message;
+use scale_info::prelude::format;
```
ntn-x2, line 21: Do you have to import this? Is it a different macro than what's used in other places?

pallets/parachain-staking/src/try_state.rs
```diff
@@ -0,0 +1,202 @@
...
+use sp_runtime::{traits::Zero, SaturatedConversion, Saturating};
+
+use crate::{
+	set::OrderedSet,
+	types::{BalanceOf, Candidate, Stake},
+	CandidatePool, Config, DelegatorState, LastDelegation, MaxCollatorCandidateStake, MaxSelectedCandidates, Pallet,
+	Round, TopCandidates, TotalCollatorStake,
+};
+
+pub(crate) fn do_try_state<T: Config>() -> Result<(), &'static str> {
```
weichweich, line 31: Maybe I didn't see that but do we check that all `DelegatorState`s have a collator where they stake?

weichweich (changes requested):
still some minor requests that need changing. I'm not a fan of `convert_error_message`. Is this the result of a discussion with @ntn-x2?

pallets/parachain-staking/src/try_state.rs
```diff
@@ -0,0 +1,202 @@
...
+			let sum_delegations: BalanceOf<T> = candidate
+				.delegators
+				.iter()
+				.fold(Zero::zero(), |acc, stake| acc.saturating_add(stake.amount));
+
+			// total stake should be the sum of delegators stake + colator stake.
+			let stake_total = sum_delegations.saturating_add(candidate.stake);
+			ensure!(
+				stake_total == candidate.total,
+				convert_error_message(format!("Total stake of collator {:?} does not match", candidate.id))
```
ntn-x2, line 59: Should we add what the expected and returned values are? In general if we are testing a condition, we might want to also want to output, in the case of a failure, what the expected condition was and what was actually returned.

support/src/lib.rs
ntn-x2: Also the function name itself is not very indicative, and a comment would not hurt about its expected usage.

ntn-x2 (commented):
Few more suggestions 😄 The bigger ones are about 1. the new utility function to return static strings for testing, 2. better error messages, that also include the failed condition, where it makes sense, so that it's easier to see what failed and for which reason. Other minor points include the way of importing the required types for this new feature.

support/src/test.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub fn convert_error_message(error_message: String) -> &'static str {
+	Box::leak(error_message.into_boxed_str())
```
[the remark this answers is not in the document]
ntn-x2: I agree the naming of the function is not ideal, as is not using `test` as the module name. Nevertheless, this is feature-gated, hence it's not possible to leak data in the runtime itself. Or did I not fully understand your point?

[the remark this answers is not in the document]
weichweich: Even leaking memory in tests is not something i would like to introduce so lightly. How many testcases do we have? how many iterations inside a test case do we do? Do we really want to end up trying to debug why our tests use so much memory?

My point is that we add a pitfall to our code base. If you are not careful and call `convert_error_message` too often we might get memory issues. The test program doesn't exit once an error is reported. And this method might get called even if no error is reported and the test was successful.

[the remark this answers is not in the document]
ntn-x2: Between an error message that just tells you something went wrong and requires you to gather more information yourself, and one that gives you already enough information to know where to look for the problem, but that could leak some memory, I would lean for the second option, which is also what I suggested Adel. `Box::leak` was one solution I found to have `&static str` that actually contain useful information. If there's a better alternative, or if you think this should not be in the codebase at all, or something else, is eventually up to you, I guess.
I just liked to have a `Delegator 4q4aTGAhfa43R1G4i3DaYMMDsDkgu2CfqAiwUup87KjFTV6M is expected to have 3 KILTs staked, but 2 KILTs were found` over `Delegator stake does not match the expected one`.

support/src/test.rs
Ad96el: I import the Box struct from `sp_std` now. I think `sp_std` has no String struct…

pallets/parachain-staking/src/try_state.rs
Ad96el: It's added now

pallets/parachain-staking/src/try_state.rs
Ad96el: I updated all possible error messages

support/src/test.rs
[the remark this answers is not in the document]
Ad96el: I log the detailed error message with `log::error!` and return for all sanity test a generic error string `"Sanity test error"`. Otherwise, I have to do something like: 

`pub fn log_and_return_error_message(detailed_error_message: String, error_message: &'static str) -> &'static str {
	log::error!("{}", detailed_error_message);
        error_message
}
`

Which does not feel right.

weichweich (approved):

PR #518, opened 2023-05-16
feat: extend delegation rewards into Y3

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2406.

Tests have been extended to check for one more year of rewards.


weichweich (approved):
The test are really hard to read.

ntn-x2:
> The test are really hard to read.

Agree, but it is what it is, and I would not touch anything right now 😄

Ad96el (approved):

PR #545, opened 2023-08-04
fix: DID endpoint validation logic

ntn-x2:
Service endpoints (URLs), were mistakenly restricted to be valid URI fragments, while they are supposed to be valid URIs. This PR does that. Plus, it replaces the custom validation logic with a third-party library with full support for the URI specification.

Tests and weights have been updated to reflect the changes.

Will have to be cherry-picked in the 1.11 release.


pallets/did/src/utils.rs
```diff
@@ -98,6 +79,33 @@ fn check_is_valid_ascii_string() {
 	});
 }
 
+#[test]
+fn check_is_valid_uri() {
+	let test_cases = [
+		("kilt.io", true),
+		("super.long.domain.com:12345/path/to/directory#fragment?arg=value", true),
+		("super.long.domain.com:12345/path/to/directory/file.txt", true),
+		("domain.with.only.valid.characters.:/?#[]@!$&'()*+,;=-._~", false),
```
Ad96el, line 88: Why is this URI wrong? Maybe a comment for the false cases can help to quicker understand.

Ad96el (approved):
LGTM just one minor comment

weichweich (approved):

pallets/did/src/utils.rs
ntn-x2: Clearer like this? 😊 https://github.com/KILTprotocol/kilt-node/pull/545/commits/cc9c79146fd26e206e2bbda64f4ff74f62a5557e

PR #547, opened 2023-08-09
fix: potential circular dependency

ntn-x2:
The lookup pallet introduced a dependency on the `runtime-common` crate, which should not really be there. While working on DIP, where the reverse relationship is introduced, we have a circular dependency.

This PR abstracts in the lookup pallet over the definition of `Balance` by re-using the trait constraints that the `pallet-balances` defines. It also resolves one lint warning regarding an unused function.

### DIFF
--- a/pallets/pallet-did-lookup/src/lib.rs
+++ b/pallets/pallet-did-lookup/src/lib.rs
@@ -105,7 +105,6 @@ pub mod pallet {
 
 		type RuntimeHoldReason: From<HoldReason>;
 
-		// type Balance: frame_support::traits::tokens::Balance;
 		/// The currency that is used to reserve funds for each did.
 		type Currency: MutateHold<AccountIdOf<Self>, Reason = Self::RuntimeHoldReason>;
 
--- a/pallets/pallet-did-lookup/src/lib.rs
+++ b/pallets/pallet-did-lookup/src/lib.rs
@@ -105,7 +105,7 @@ pub mod pallet {
 
 		type RuntimeHoldReason: From<HoldReason>;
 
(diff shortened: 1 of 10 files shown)


weichweich (commented):
This shouldn't have happened. Good catch

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -107,8 +109,19 @@ pub mod pallet {
...
+			+ Member
+			+ AtLeast32BitUnsigned
+			+ Codec
+			+ Default
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ sp_std::fmt::Debug
+			+ MaxEncodedLen
+			+ TypeInfo
+			+ FixedPointOperand;
```
weichweich, line 122: Can you not simply use MutateHold::Balance? Do you need the additional trait bounds?

ntn-x2: I was able to bring it down to [this](https://github.com/KILTprotocol/kilt-node/pull/547/commits/f57c084f26949ca0d95dc76b5c97d89f65e88700). If I remove the trait bounds completely, everything becomes red.

ntn-x2: Ok maybe this works... https://github.com/KILTprotocol/kilt-node/pull/547/commits/609a41b18776d2fc0304823ce894bd783afaccd0

weichweich (approved):

Ad96el (approved):

Ad96el:
Closed by accident :D

PR #551, opened 2023-08-21
feat: single signature DID calls/creation

weichweich:
## fix https://github.com/KILTprotocol/ticket/issues/2849

This PR enables blockchain accounts to manage DIDs on the KILT chain.

Sending DID calls and creating DIDs will be possible with any substrate enabled wallet (polkadotJS, Nova, ...).

Creating a DID will also be possible but still requires two signatures. By supporting wrapped payloads as a special signature type, wallets that support `signRaw` or similar will be able to create DIDs. With this wrapped type the default way of sending DIDCalls would also work, but still require a second signature.

<details>

```
!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
[≠] pallet 64: Did -> 5 change(s)
  - calls changes:
    [≠]  1: set_authentication_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [≠]  2: set_delegation_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [≠]  4: set_attestation_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [+] CallDesc { index: 15, name: "dispatch_as", signature: SignatureDesc { args: [ArgDesc { name: "did_identifier", ty: "DidIdentifierOf<T>" }, ArgDesc { name: "call", ty: "Box<DidCallableOf<T>>" }] } }
    [+] CallDesc { index: 16, name: "create_from_account", signature: SignatureDesc { args: [ArgDesc { name: "authentication_key", ty: "DidVerificationKey<AccountIdOf<T>>" }] } }

SUMMARY:
- Compatible.......................: false
- Require transaction_version bump.: false

!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
```

</details>

<details>

```
!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
[≠] pallet 64: Did -> 5 change(s)
  - calls changes:
    [≠]  1: set_authentication_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [≠]  2: set_delegation_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [≠]  4: set_attestation_key ( new_key: DidVerificationKey, )  )
        [Signature(SignatureChange { args: [Changed(0, [Ty(StringChange("DidVerificationKey", "DidVerificationKey<AccountIdOf<T>>"))])] })]
    [+] CallDesc { index: 15, name: "dispatch_as", signature: SignatureDesc { args: [ArgDesc { name: "did_identifier", ty: "DidIdentifierOf<T>" }, ArgDesc { name: "call", ty: "Box<DidCallableOf<T>>" }] } }
    [+] CallDesc { index: 16, name: "create_from_account", signature: SignatureDesc { args: [ArgDesc { name: "authentication_key", ty: "DidVerificationKey<AccountIdOf<T>>" }] } }

SUMMARY:
- Compatible.......................: false
- Require transaction_version bump.: false

!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!

```

</details>

## Checklist:

    * Link the PR/Ticket here


pallets/did/src/did_details.rs
```diff
@@ -141,6 +173,12 @@ pub enum DidSignature {
 	Sr25519(sr25519::Signature),
 	/// An Ecdsa signature.
 	Ecdsa(ecdsa::Signature),
+	/// A Ed25519 signature.
+	WrappedEd25519(ed25519::Signature),
+	/// A Sr25519 signature.
+	WrappedSr25519(sr25519::Signature),
+	/// An Ecdsa signature.
+	WrappedEcdsa(ecdsa::Signature),
```
weichweich, line 181: Not needed anymore

pallets/did/src/tests/submit.rs
```diff
@@ -28,6 +28,33 @@ use crate::{
 	mock_utils::*,
 };
 
+#[test]
```
weichweich, line 31: i pulled the "successful-case" test to the top

runtimes/peregrine/src/weights/did.rs
```diff
@@ -723,6 +722,38 @@ impl<T: frame_system::Config> did::WeightInfo for WeightInfo<T> {
...
 	}
+
+	/// Storage: Did Did (r:1 w:0)
+	/// Proof: Did Did (max_values: None, max_size: Some(2312), added: 4787, mode: MaxEncodedLen)
+	fn dispatch_as() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `353`
+		//  Estimated: `5777`
+		// Minimum execution time: 186_836_000 picoseconds.
+		Weight::from_parts(189_377_000, 0)
```
weichweich, line 733: weights will be calculated on our benchmark server. There were done with debug builds

pallets/did/src/did_details.rs
```diff
@@ -402,14 +429,39 @@ impl<T: Config> DidDetails<T> {
...
+			amount: Zero::zero(),
+		};
+
+		// Creates a new DID with the given authentication key.
+		let mut new_did_details = DidDetails::new(new_auth_key, current_block_number, deposit)?;
+
+		let deposit_amount = new_did_details.calculate_deposit(did_subject);
+		new_did_details.deposit.amount = deposit_amount;
+
+		DidDepositCollector::<T>::create_deposit(submitter, deposit_amount)?;
```
Ad96el, line 453: Should this function have side effects?

Ad96el (commented):
3 minor comments. I like the modular test setup. :partying_face:

pallets/did/src/did_details.rs
```diff
@@ -70,21 +72,46 @@ impl DidVerificationKey {
...
 	}
+
+	pub fn into_account(&self) -> AccountId
+	where
+		AccountId: From<[u8; 32]> + AsRef<[u8; 32]>,
+	{
+		let bytes = match self {
+			DidVerificationKey::Ed25519(pub_key) => pub_key.0,
+			DidVerificationKey::Sr25519(pub_key) => pub_key.0,
+			DidVerificationKey::Ecdsa(pub_key) => sp_io::hashing::blake2_256(pub_key.as_ref()),
```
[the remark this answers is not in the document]
weichweich: The ECDSA key is 33 bytes long. An accountID only 32. In substrate, ecdsa keys are therefore hashed to get the accountId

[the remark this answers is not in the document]
weichweich: [5a6a650](https://github.com/KILTprotocol/kilt-node/pull/551/commits/5a6a65016573b105963852ef3879dc37466356f0)

pallets/did/src/did_details.rs
```diff
@@ -402,14 +430,39 @@ impl<T: Config> DidDetails<T> {
 		Ok(new_did_details)
 	}
 
+	// Creates a new DID entry from some [DidCreationDetails] and a given
+	// authentication key.
+	pub fn from_account(
+		submitter: AccountIdOf<T>,
```
ntn-x2, line 436: I think also here it would be more correct to name this `origin`, right? It might not be a tx submitter.

pallets/did/src/did_details.rs
```diff
@@ -70,21 +72,47 @@ impl DidVerificationKey {
...
+			// Hash the Ecdsa key the same way it's done in substrate (the ecdsa key is 33 bytes, one byte too long)
+			DidVerificationKey::Ecdsa(pub_key) => sp_io::hashing::blake2_256(pub_key.as_ref()),
+			DidVerificationKey::Account(acc_id) => *acc_id.as_ref(),
+		};
+
+		bytes.into()
+	}
+}
+
+impl<AccountId: AsRef<[u8]>> AsRef<[u8]> for DidVerificationKey<AccountId> {
```
ntn-x2, line 92: Why do we need this implementation here?

pallets/did/src/lib.rs
```diff
@@ -1140,9 +1152,139 @@ pub mod pallet {
...
+		/// identifier or if a DID with the same identifier existed and was
+		/// deleted.
+		///
+		/// The origin for this account must be funded and provide the required
+		/// deposit and fee.
+		///
+		/// Emits `DidCreated`.
+		#[pallet::call_index(16)]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::create_from_account())]
+		pub fn create_from_account(
```
ntn-x2, line 1242: I would say that maybe this is the only function that should be trait bounded to cases where `AccountId` can be coalesced into a `DidIdentifier`, right? The bound should not apply to the generic pallet itself.

ntn-x2 (commented):
Love it! I just don't understand why the widespread use of `AsRef<[u8]>` for DID keys?

pallets/did/src/lib.rs
```diff
@@ -1140,9 +1152,139 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			authentication_key: DidVerificationKey<AccountIdOf<T>>,
+		) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
+			let did_identifier: DidIdentifierOf<T> = sender.clone().into();
+			let auth_key_id: AccountIdOf<T> = authentication_key.into_account();
+			ensure!(auth_key_id == sender, Error::<T>::BadDidOrigin);
```
ntn-x2, line 1249: Why this line here? So if you are using one account, you can't use a key different than the one that matches the origin's account? I thought this additional extrinsic was only supposed to remove the signature requirement.

pallets/did/src/did_details.rs
weichweich: Yes that was necessary for some old code but not needed anymore. e9ddbd4d5cfbdf22f1fd89ffdeccabfa50088bf4

pallets/did/src/lib.rs
```diff
@@ -506,7 +509,10 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T> {
+	impl<T: Config> Pallet<T>
+	where
+		T::AccountId: AsRef<[u8; 32]> + From<[u8; 32]>,
```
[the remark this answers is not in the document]
weichweich: I checked the occurrences and they are indeed required everywhere where they are right now

pallets/did/src/did_details.rs
```diff
@@ -70,21 +72,47 @@ impl DidVerificationKey {
 			_ => Err(errors::SignatureError::InvalidFormat),
 		}
 	}
+
+	pub fn into_account(&self) -> AccountId
```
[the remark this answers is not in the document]
weichweich: Good idea: [d04c0e1](https://github.com/KILTprotocol/kilt-node/pull/551/commits/d04c0e10853abd605e31fe98473de8b9d02ca9d7)

pallets/did/src/lib.rs
weichweich: The bounds are required for the `into_account` function. from there they will propagate up to where the generic is actually specified (or set, i.e. in the runtime).
We could add the where clause to both `create_from_account` and `dispatch_as` but since these will be made into enums and are heavily changed by macros, that doesn't work. So we need to keep the where-clause where it is (at least I didn't find a quick way to move it away from where it is). I also think it's fine to have it there, since it's more obvious that this pallet requires special things. If it's for the individual call, it's harder to spot.

pallets/did/src/lib.rs
weichweich: The origin has to be the authentication key. This is basically the check that the owner of the authentication key agrees to creating the DID.
This basically replaces the second DID signature that is required for the old `create` call.
If we would not have this check, I could create DIDs for any public key I want.

weichweich: Ah ok. Now I understand. That would be no different than just changing the authentication key immediately after creating the DID.

weichweich: [a8090e5](https://github.com/KILTprotocol/kilt-node/pull/551/commits/a8090e5243bc73e6eb55f9d2e1d3fef92e1985e9)

pallets/did/src/lib.rs
```diff
@@ -1140,9 +1152,139 @@ pub mod pallet {
...
+				&did_entry.deposit.owner,
+				T::Fee::get(),
+				Precision::Exact,
+				Preservation::Preserve,
+				Fortitude::Polite,
+			)?;
+
+			T::FeeCollector::on_unbalanced(imbalance);
+
+			Self::deposit_event(Event::DidCreated(sender, did_identifier));
```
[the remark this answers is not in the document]
weichweich: Good idea. I also bundled this up with the checks. They where separate since we didn't had transactional extrinsic. Now that they are, we can return errors when ever we want.
[11c4fa4](https://github.com/KILTprotocol/kilt-node/pull/551/commits/11c4fa498c4c457970dda03e512b8968e5f688f0)

pallets/did/src/did_details.rs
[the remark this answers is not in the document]
weichweich: It should not. I also updated the `update_deposit` function. We now have a `try_update_did` function that makes sure the deposit is updated every time the did is changed. [602a0f7](https://github.com/KILTprotocol/kilt-node/pull/551/commits/602a0f7d7e54c05ff2b5fb28cc03c151fadbbeed)

pallets/did/src/lib.rs
```diff
@@ -1178,6 +1320,28 @@ pub mod pallet {
 			Ok(())
 		}
 
+		pub fn verify_account_authorization(
```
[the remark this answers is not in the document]
weichweich: That is true: [cfa486d](https://github.com/KILTprotocol/kilt-node/pull/551/commits/cfa486db9ae4c4b42461cd6ae61cdcec788821aa)

pallets/did/src/lib.rs
```diff
@@ -213,7 +213,10 @@ pub mod pallet {
 			+ DeriveDidCallAuthorizationVerificationKeyRelationship;
 
 		/// Type for a DID subject identifier.
-		type DidIdentifier: Parameter + DidVerifiableIdentifier + MaxEncodedLen;
+		type DidIdentifier: Parameter
+			+ DidVerifiableIdentifier<AccountIdOf<Self>>
+			+ MaxEncodedLen
+			+ From<AccountIdOf<Self>>;
```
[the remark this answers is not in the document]
weichweich: This cannot be removed since it's required by `create_from_account`. There we derive the did_identifier from the calling account and set an unrelated key as the authentication key.

DidIdentifier and AccountId can still be completely different types. There just have to be a way to derive a DidIdentifier from an AccountId.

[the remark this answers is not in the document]
ntn-x2: This goes back to the discussion we had last time. Perhaps the `create_from_account` extrinsic should only exist for instances of the pallet where the DID identifier can indeed be derived from the account ID. Otherwise, this is a strong assumption we are introducing here. I mean, we can always relax this requirement if we need to in the future.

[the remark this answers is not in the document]
weichweich: Yes and as I stated the last time, we can't add where clauses to calls. The macro derives an enum and some functions from this call and the where clause doesn't translate well there. There is no way to add the requirement to all the generated code. So it has to be on the whole pallet.

I could try to setup some conditional compilation and a feature flag or something. But I don't think this is worth the trouble.

```
  error[E0277]: the trait bound `<T as pallet::Config>::DidIdentifier: From<<T as frame_system::Config>::AccountId>` is not satisfied
      --> /pallets/did/src/lib.rs:510:12
       |
  510  |       #[pallet::call]
       |  _______________^
  511  | |     impl<T: Config> Pallet<T>
  512  | |     where
  513  | |         T::AccountId: AsRef<[u8; 32]> + From<[u8; 32]>,
  ...    |
  1211 | |         #[pallet::weight(<T as pallet::Config>::WeightInfo::create_from_account())]
  1212 | |         pub fn create_from_account(
       | |__________________________________^ the trait `From<<T as frame_system::Config>::AccountId>` is not implemented for `<T as pallet::Config>::DidIdentifier`
       |
  note: required by a bound in `pallet::Pallet::<T>::create_from_account`
      --> /pallets/did/src/lib.rs:1217:22
       |
  1212 |         pub fn create_from_account(
       |                ------------------- required by a bound in this associated function
  ...
  1217 |             T::DidIdentifier: From<AccountIdOf<T>>,
       |                               ^^^^^^^^^^^^^^^^^^^^ required by this bound in `Pallet::<T>::create_from_account`
  help: consider further restricting the associated type
       |
  510  |     #[pallet::call where <T as pallet::Config>::DidIdentifier: From<<T as frame_system::Config>::AccountId>]
       |                    ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

  For more information about this error, try `rustc --explain E0277`.
  error: could not compile `did` (lib) due to previous error
```

pallets/did/src/lib.rs
```diff
@@ -1178,6 +1320,28 @@ pub mod pallet {
 			Ok(())
 		}
 
+		pub fn verify_account_authorization(
+			did_identifier: &DidIdentifierOf<T>,
+			account: &AccountIdOf<T>,
```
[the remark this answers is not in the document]
weichweich: Origin would be a `RootOrigin`, `SignedOrigin` etc. Something more wrapped and not a "simple" Id. So I rather name this `submitter_account`. [9b139f6](https://github.com/KILTprotocol/kilt-node/pull/551/commits/9b139f68f50eca7acd6420f57f028e30683cc75f)

pallets/did/src/did_details.rs
weichweich: The whole function isn't that great. I thing it should rather be called `new_with_...` and also it's not really that different to `new` anyways so i merged these two. [9b139f6](https://github.com/KILTprotocol/kilt-node/pull/551/commits/9b139f68f50eca7acd6420f57f028e30683cc75f)

The submitter here should be called `deposit_owner` IMO since that's what the accountId is used for in the end.

pallets/did/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: Yes, we will think about it in the future if and when the time comes.

pallets/did/src/did_details.rs
ntn-x2: Like it!

ntn-x2 (approved):

Ad96el (approved):

PR #553, opened 2023-08-23
feat: add support for relay chain

ntn-x2:
Builds on the new state proofs by adding support for a relay chain to be a consumer with the new `ChildParachainDipStateProof` and `DipChildProviderStateProofVerifier` types. Some traits and types have also been renamed.

The new feature has been tested with a local Rococo 0.9.43 deployment, and can be reproduced by download and compiling our Polkadot fork at the [demo-sub0-23 tag](https://github.com/KILTprotocol/polkadot/tree/demo-sub0-23).

This PR also adds the `postit` pallet that was written for the Decoded demo, and uses that instead of the lookup pallet as an example pallet integrating DIP.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6773,6 +6773,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-postit"
+version = "1.11.0-dev"
+dependencies = [
+ "frame-support",
+ "frame-system",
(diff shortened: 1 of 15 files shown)


ntn-x2:
Merging it in the state proof PR, and will be reviewed altogether as part of that PR.

PR #554, opened 2023-08-24
refactor: rename standalone to kestrel

weichweich:
## fixes https://github.com/KILTprotocol/ticket/issues/2794

* i decided to remove the mashnet naming entirely. Mashnet is the old thing from the very beginning.
* the mashnet-node was renamed to standalone-node since it's not specific to kestrel (there could be a second runtime executed in with this node, like we do it for the parachain-node)
* change the impl-name to `kilt-peregrine`
## Checklist:

    * Link the PR/Ticket here


ntn-x2 (commented):
I see quite few occurrences of mashnet with a quick search function.

README.md
```diff
@@ -1,4 +1,4 @@
-# KILT-node &middot; [![tests](https://gitlab.com/kiltprotocol/mashnet-node/badges/develop/pipeline.svg)](https://gitlab.com/kiltprotocol/mashnet-node/-/commits/develop)
+# KILT-node &middot; [![tests](https://gitlab.com/kiltprotocol/kilt-node/badges/develop/pipeline.svg)](https://gitlab.com/kiltprotocol/kilt-node/-/commits/develop)
```
ntn-x2, line 1: 🥶🥶🥶

ntn-x2 (commented):
`.gitlab-ci.yml` still has a `DOCKER_HUB_STANDALONE: "kiltprotocol/mashnet-node"` which should be renamed.

weichweich:
> `.gitlab-ci.yml` still has a `DOCKER_HUB_STANDALONE: "kiltprotocol/mashnet-node"` which should be renamed.

This is the image on dockerhub. I would keep this name. Not sure what depends on this naming.

runtimes/peregrine/src/lib.rs
```diff
@@ -90,7 +90,7 @@ impl_opaque_keys! {
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("mashnet-node"),
-	impl_name: create_runtime_str!("mashnet-node"),
+	impl_name: create_runtime_str!("kilt-peregrine"),
```
ntn-x2, line 93: Not sure I like this... EIther we don't change it at all, or we change both. What's the reason for changing only the implementation?

weichweich: The idea was that there is at least one difference in the naming. spec name might be the same but at least the impl name is different.

ntn-x2: I think spec name should also be different. That's the assumption the type definitions do as well. I think impl name is probably a different implementation of the Peregrine runtime. But the spec name indicates the runtime. If Peregrine is not mashnet, then the spec name should not be mashnet either.

weichweich: Yes the only reason I'm not changing the spec name is, that I'm not 100% sure about the consequences

weichweich: Ok let's do this as part of another PR once we researched all the consequences of renaming the spec.

[8adcaa4](https://github.com/KILTprotocol/kilt-node/pull/554/commits/8adcaa454c4ba8fe4dc2de08634695d67d4c4861)

ntn-x2 (approved):
🐦‍⬛🐦‍⬛🐦‍⬛

weichweich:
As a cherry on top, here is the `subwasm diff` report:

<details>
  <summary>Develop branch Peregrine Diff</summary>

```
!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
No change detected
SUMMARY:
- Compatible.......................: true
- Require transaction_version bump.: false

!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
```

</details>

<details>
  <summary>Develop branch Spiritnet Diff</summary>

```
!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
No change detected
SUMMARY:
- Compatible.......................: true
- Require transaction_version bump.: false

!!! THE SUBWASM REDUCED DIFFER IS EXPERIMENTAL, DOUBLE CHECK THE RESULTS !!!
```

</details>

PR #558, opened 2023-08-31
refactor: split tests and improve key generation

weichweich:
* Using the seed instead of a bool makes it more flexible to use and clear that what the parameter does. A bool is not as self explaining as a seed.
* the new file structure makes it easier to navigate the tests. it also categorizes the test.


pallets/did/src/tests/create.rs
```diff
@@ -0,0 +1,757 @@
+use frame_support::{
```
Ad96el, line 1: The licence is missing

pallets/did/src/mock.rs
```diff
@@ -277,84 +277,44 @@ pub fn get_did_identifier_from_ecdsa_key(public_key: ecdsa::Public) -> DidIdenti
 	MultiSigner::from(public_key).into_account()
 }
 
-pub fn get_ed25519_authentication_key(default: bool) -> ed25519::Pair {
-	if default {
-		ed25519::Pair::from_seed(&DEFAULT_AUTH_SEED)
-	} else {
-		ed25519::Pair::from_seed(&ALTERNATIVE_AUTH_SEED)
-	}
+pub fn get_ed25519_authentication_key(seed: &[u8; 32]) -> ed25519::Pair {
```
Ad96el, line 280: If we already start to refactor our mocks and tests, I would rather see these functions into the `mock_util`. The mock should only contain mocked pallets and the `ExtBuilder`.

Ad96el (commented):
LGTM, two minor comments. I think we should also do that for the other pallets. Files containing over 1000+ lines of code should be avoided.

pallets/did/src/mock.rs
Ad96el: Alright. Will create some tickets.

Ad96el (approved):

PR #564, opened 2023-09-21
ci: add try-runtime builds

weichweich:
This PR adds Wasm builds with the `try-runtime` feature enabled. This is in preparation for more automation in the release process.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -121,7 +121,7 @@ build-wasm-try-runtime:
   script:
     - cargo build --release -p $RUNTIME-runtime --features try-runtime
     - mkdir ./out
-    - mv target/release/wbuild/$RUNTIME-runtime/*.wasm /out/
+    - mv target/release/wbuild/$RUNTIME-runtime/*.wasm ./out/
   artifacts:
     paths:
       - out/*.wasm
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -118,6 +118,10 @@ build-wasm-try-runtime:
       - RUNTIME: "spiritnet"
   image: paritytech/ci-unified:bullseye-1.70.0
   stage: build
+  only:
+    - develop
+    - master
+    - /^[0-9]+(?:\.[0-9]+){2}(?:-(rc)*([0-9])+)?$/
   script:
     - cargo build --release -p $RUNTIME-runtime --features try-runtime
     - mkdir ./out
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -110,3 +110,24 @@ build-wasm-spiritnet:
...
+      - RUNTIME: "peregrine"
+      - RUNTIME: "spiritnet"
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: build
+  only:
+    - develop
+    - master
+    - /^[0-9]+(?:\.[0-9]+){2}(?:-(rc)*([0-9])+)?$/
+  script:
+    - cargo build --release -p $RUNTIME-runtime --features try-runtime
```
ntn-x2, line 126: `--locked`? Also, I see srtool has a `default-features` option which we could use to specify the `try-runtime` feature. I think it would not hurt if we would use srtool also to generate the WASM for runtime upgrade testing. What do you think?

weichweich: Yeah I thought about using the SR Tool and also make the 4 options (spiritnet, peregrine, spiritnet+try-runtime, peregrine+try-runtime) a matrix. But I didn't want to mix up the production build with the try-runtime builds. I feared that it would make it easier to mix them up.

Also why would you use the SR-tool for this?

ntn-x2: > Yeah I thought about using the SR Tool and also make the 4 options (spiritnet, peregrine, spiritnet+try-runtime, peregrine+try-runtime) a matrix. But I didn't want to mix up the production build with the try-runtime builds. I feared that it would make it easier to mix them up.

Agree. They can be two separate matrices.

> Also why would you use the SR-tool for this?

Since we enable SRTool-built runtimes, using SRTool also for try-runtime builds would make sure the code is consistent. We test the exact same code that will be deployed, without the `try-runtime` feature.

weichweich: Not sure how much we win. AFAIK we use the same compiler in the SR-Tool and the CI. The feature already messes up the wasm, so that it's not the same as the production version. Not sure how much the SR-Tool will help there. But I guess it's really doesn't hurt. 🤷

ntn-x2: Yes, probably not a big win overall, but if the effort is not too big, I would go down this road. Also from a mental model point of view, we are now making sure the SRTool is used to build all WASM artifacts, regardless of their features. If SRTool does not make it possible to build the try-runtime WASM, then of course the CI is close enough, I guess.

PR #572, opened 2023-10-20
feat: add meaningful errors

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2552, by both removing all occurrences of `.expect()` and `.unwrap()` and replacing them with `Internal` errors, and by introducing error enums. The logic is that for both the provider and the consumer, a generic error is bubbled up from the runtime, and because of the constraints imposed by Substrate, the error reason is encoded as a `u16`, where each "class" of error is namespaced with a 255 value increment. A snippet of this is shown below:

```rust
fn from(
    value: DipSiblingProviderStateProofVerifierError<
        ParachainHeadMerkleProofVerificationError,
        IdentityCommitmentMerkleProofVerificationError,
        DipProofVerificationError,
        DidSignatureVerificationError
    >,
) -> Self {
    match value {
        DipSiblingProviderStateProofVerifierError::ParachainHeadMerkleProofVerificationError(error) => {
            error.into()
        }
        DipSiblingProviderStateProofVerifierError::IdentityCommitmentMerkleProofVerificationError(error) => {
            u8::MAX as u16 + error.into() as u16
        }
        DipSiblingProviderStateProofVerifierError::DipProofVerificationError(error) => {
            u8::MAX as u16 * 2 + error.into() as u16
        }
        DipSiblingProviderStateProofVerifierError::DidSignatureVerificationError(error) => {
            u8::MAX as u16 * 3 + error.into() as u16
        }
    }
}
```

The specific reason for the error can then be retrieved and decoded directly from the Polkadot Apps interface, as the example below shows, where the first byte `01` indicates the enum variant index, the second byte is the varint encoding of the SCALE-encoded u16, and the third and forth bytes `e000` represent the `u16` error reason. This can be decoded as decimal, where each value has a one-to-one link to a specific error from the runtime.

The pallet consumer expects any other implementation of the DIP consumer protocol to expose the proof verification error reason as a `u16`, but of course developers are free to not give any meaning to it and treat everything as a generic `InvalidProof` error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4530,6 +4530,7 @@ dependencies = [
  "frame-system",
  "hash-db",
  "hex-literal 0.3.4",
+ "log",
  "pallet-dip-consumer",
  "pallet-dip-provider",
  "pallet-relay-store",
--- a/crates/kilt-dip-support/Cargo.toml
+++ b/crates/kilt-dip-support/Cargo.toml
@@ -13,6 +13,7 @@ version.workspace = true
 [dependencies]
 # External dependencies
 hash-db.workspace = true
+log.workspace = true
 
 # Internal dependencies
 did.workspace = true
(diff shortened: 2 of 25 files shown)


ntn-x2:
Merging without reviews. It will be reviewed later on from the `aa/dip` branch directly.

PR #574, opened 2023-11-02
feat: add deposits for DIP provider pallet

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2983.

Add a generic pallet to store deposits, as well as a hooking mechanism to both the existing provider pallet (to allow runtimes that take deposits to take deposits, or to do any other operation), and to the new deposit pallet, which allows other pallets to clean up storage if a storage is reclaimed back by its owner (e.g., an identity commitment is removed).

Tests will follow in a separate PR, when time allows for them 😁 The whole DIP branch won't be merged on `develop` before tests are written, so it's not a risk. The logic has been tested by spinning up a local network and trying to commit an identity (which takes a deposit), remove the commitment (which frees up the deposit), re-commit the identity, reclaim the deposit (which removes the commitment).

- [x] Implement logic for deposit removal hooks
- [x] Test
- [x] Review


dip-template/runtimes/dip-provider/src/dip.rs
```diff
@@ -27,22 +27,59 @@ use runtime_common::dip::{
...
-#[derive(Encode, Decode, TypeInfo)]
-pub struct RuntimeApiDipProofRequest {
-	pub(crate) identifier: DidIdentifier,
-	pub(crate) version: IdentityCommitmentVersion,
-	pub(crate) keys: Vec<KeyIdOf<Runtime>>,
-	pub(crate) accounts: Vec<LinkableAccountId>,
-	pub(crate) should_include_web3_name: bool,
+use crate::{AccountId, Balances, DidIdentifier, Hash, Runtime, RuntimeEvent};
+
+pub mod runtime_api {
```
ntn-x2, line 32: Please ignore this, not relevant for what I'd like to get reviewed.

dip-template/runtimes/dip-provider/src/dip.rs
```diff
@@ -27,22 +27,59 @@ use runtime_common::dip::{
...
+
+	pub struct Namespace;
+
+	impl Get<BoundedVec<u8, ConstU32<MAX_NAMESPACE_LENGTH>>> for Namespace {
+		fn get() -> BoundedVec<u8, ConstU32<MAX_NAMESPACE_LENGTH>> {
+			(*b"DipProvider").to_vec().try_into().expect("Should never fail.")
+		}
+	}
+
+	pub const DEPOSIT_AMOUNT: Balance = 100 * UNIT;
```
ntn-x2, line 70: This will be a constant specified somewhere in the runtime config.

trusch: If we want to reuse this pallet, we should put some more thought in how to configure it for different actions that require deposits.

dip-template/runtimes/dip-provider/Cargo.toml
Ad96el, line 1: Maybe not part of this PR, but does it not make more sense to have the dip-template in his own repo?

dip-template/runtimes/dip-provider/src/lib.rs
```diff
@@ -568,14 +569,14 @@ impl_runtime_apis! {
...
+		fn generate_proof(request: runtime_api::DipProofRequest) -> Result<CompleteMerkleProof<Hash, DidMerkleProofOf<Runtime>>, runtime_api::DipProofError> {
 			let linked_did_info = match <Runtime as pallet_dip_provider::Config>::IdentityProvider::retrieve(&request.identifier) {
 				Ok(Some(linked_did_info)) => Ok(linked_did_info),
-				Ok(None) => Err(RuntimeApiDipProofError::IdentityNotFound),
-				Err(e) => Err(RuntimeApiDipProofError::IdentityProviderError(e))
+				Ok(None) => Err(runtime_api::DipProofError::IdentityNotFound),
+				Err(e) => Err(runtime_api::DipProofError::IdentityProviderError(e))
 			}?;
-			DidMerkleRootGenerator::<Runtime>::generate_proof(&linked_did_info, request.version, request.keys.iter(), request.should_include_web3_name, request.accounts.iter()).map_err(RuntimeApiDipProofError::MerkleProofError)
+			DidMerkleRootGenerator::<Runtime>::generate_proof(&linked_did_info, request.version, request.keys.iter(), request.should_include_web3_name, request.accounts.iter()).map_err(runtime_api::DipProofError::MerkleProofError)
```
Ad96el, line 579: These lines look very complicated. Can we maybe split it? Especially if it is used as template

Ad96el (commented):
Very nice! I like the architectural design and the default implementation of the traits. Some comments are a bit nitpicking - sorry for that. I assume the pallet-deposit-storage will also be deployed on Spiritnet, right? If so, the storage version has to be initialized in the runtimes. Additionally, I would appreciate more documentation. Perhaps this can be addressed in another PR?

pallets/pallet-deposit-storage/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+		},
+	};
+	use frame_system::pallet_prelude::*;
+	use parity_scale_codec::FullCodec;
+	use scale_info::TypeInfo;
+	use sp_runtime::DispatchError;
+	use sp_std::fmt::Debug;
+
+	/// The current storage version.
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```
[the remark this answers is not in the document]
ntn-x2: It's storage version 0, so migrations would not fail if they check that, right? The default would still be 0. I would migrate it once it is bumped to 1, no?

pallets/pallet-deposit-storage/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+	pub trait Config: frame_system::Config {
+		#[pallet::constant]
+		type MaxKeyLength: Get<u32>;
+
+		type CheckOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+		type Currency: Mutate<Self::AccountId, Reason = Self::RuntimeHoldReason>;
+		type DepositHooks: DepositStorageHooks<Self>;
+		type Namespace: Parameter;
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type RuntimeHoldReason: From<HoldReason> + Clone + PartialEq + Debug + FullCodec + MaxEncodedLen + TypeInfo;
```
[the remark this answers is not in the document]
ntn-x2: Yeah I haven't worked on the docs yet. Sorry if this is hard to read, I will work on a docs PR later on.

pallets/pallet-dip-provider/src/lib.rs
```diff
@@ -31,23 +31,17 @@ pub mod pallet {
 	use frame_support::{pallet_prelude::*, traits::EnsureOrigin};
 	use frame_system::pallet_prelude::*;
 	use parity_scale_codec::FullCodec;
-	use sp_io::MultiRemovalResults;
 	use sp_std::fmt::Debug;
 
-	use crate::traits::{IdentityCommitmentGenerator, IdentityProvider, SubmitterInfo};
+	use crate::traits::{IdentityCommitmentGenerator, IdentityProvider, ProviderHooks, SubmitterInfo};
```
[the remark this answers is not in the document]
ntn-x2: You mean that there is a `Config` type which is also called `ProviderHooks`?

dip-template/runtimes/dip-provider/Cargo.toml
ntn-x2: Yes, absolutely. Not part of this PR, but it will be its own repo, eventually.

dip-template/runtimes/dip-provider/src/lib.rs
ntn-x2: Like this? https://github.com/KILTprotocol/kilt-node/pull/574/commits/24cb5ac5c19b80461a0d86ba6515b8c521aa255a

pallets/pallet-deposit-storage/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Are you sure that the on-chain storage version is set to 0 by default for a new pallet? As far as I know, the on-chain version needs to be initialized. This is what we did in this [PR](https://github.com/KILTprotocol/kilt-node/commit/8e36fb26011887f726eaef21036cf56e02450c4a#diff-0fb85acd71505115281e23edf6e01a99b84b52eb8ba07629ba7518534a71d3d7R54) for the multisig pallet.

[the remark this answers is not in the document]
ntn-x2: No I'm not saying it is set to 0, but when being read, if not present, the default value of `0` is returned, which matches the pallet storage version, hence the check `stored_version == pallet_version` will not fail. Of course, once the pallet version is updated to `1`, then it will have to be written on storage. The issue with Substrate pallet, is that it was not written on storage when the pallet version was greater than `0`, and THAT is a problem.

pallets/pallet-dip-provider/src/lib.rs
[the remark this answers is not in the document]
Ad96el: No, sorry - I'm not referring to the Config. I shouldn't have marked it here.

In this [line](https://github.com/KILTprotocol/kilt-node/pull/574/files#diff-6dc5259ea5e53a98cd0b1ecd6ce9819c2efbb44cb8957070cb0e69b75fbdfa65R65), you're implementing the trait `ProviderHooks` for `FixedDepositCollectorViaDepositsPallet`. When I read the line, my initial assumption was that `ProviderHooks` are from the pallet-deposit-storage. It would be more clear if the trait somehow indicated that the hooks are from the DIP pallet.

pallets/pallet-deposit-storage/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Ah, I see. There is a default implementation. Thanks for your explanation :racehorse:

dip-template/runtimes/dip-provider/src/lib.rs
Ad96el: Yes, I think it is already better. Maybe we can have a type `IdentityProviderOf<Runtime>` instead of `<Runtime as pallet_dip_provider::Config>::IdentityProvider`. But this is not really urgent or important.

pallets/pallet-dip-provider/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: Mhh what makes you think that `ProviderHooks` is part of the deposit pallet? Because the type implementing it is declared here? Also, the trait is imported from `pallet_dip_provider` with `use pallet_dip_provider::traits::ProviderHooks`. But I anyway renamed the import to avoid confusion, what do you think?  https://github.com/KILTprotocol/kilt-node/pull/574/commits/f73778daaf1dbf38db5d0a6a948698c124001146

dip-template/runtimes/dip-provider/src/lib.rs
ntn-x2: Yup: https://github.com/KILTprotocol/kilt-node/pull/574/commits/f73778daaf1dbf38db5d0a6a948698c124001146.

pallets/pallet-dip-provider/src/lib.rs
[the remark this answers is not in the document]
Ad96el: I believe that 'ProviderHooks' is a broad term that could be associated with many pallets. I think that with the renaming, it becomes more declarative. I needed to scroll up to see where the trait comes from - will be not necessary anymore.

Ad96el (approved):
:rocket:

dip-template/runtimes/dip-provider/src/dip.rs
ntn-x2: As discussed on Slack, the initial scope of this pallet is to support storing/removing deposits. More complex logic COULD be added in the future, if we see fit. For now I'm marking this conversation as resolved.

pallets/pallet-dip-provider/src/lib.rs
```diff
@@ -31,23 +31,18 @@ pub mod pallet {
...
 
+	pub type IdentityProviderOf<T> = <T as Config>::IdentityProvider;
 	pub type IdentityOf<T> = <<T as Config>::IdentityProvider as IdentityProvider<<T as Config>::Identifier>>::Success;
 	pub type IdentityCommitmentVersion = u16;
 
 	pub const LATEST_COMMITMENT_VERSION: IdentityCommitmentVersion = 0;
+	pub const MAX_COMMITMENTS_PER_IDENTITY: u16 = LATEST_COMMITMENT_VERSION + 1;
```
Ad96el, line 43: What is the purpose of this constant? It appears that it may not be used.

ntn-x2: Ops, good catch! I removed it. It was meant to put an upper bound on the number of commitments per user, but since it's a double map with (DID + commitment version), there is already an upper bound given by the number of versions supported.

ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/574/commits/1841f0fa64e49381d4e18d3124950624a8343d65

pallets/pallet-dip-provider/src/lib.rs
```diff
@@ -31,23 +31,17 @@ pub mod pallet {
...
 	use parity_scale_codec::FullCodec;
-	use sp_io::MultiRemovalResults;
 	use sp_std::fmt::Debug;
 
-	use crate::traits::{IdentityCommitmentGenerator, IdentityProvider, SubmitterInfo};
+	use crate::traits::{IdentityCommitmentGenerator, IdentityProvider, ProviderHooks, SubmitterInfo};
 
+	pub type IdentityProviderOf<T> = <T as Config>::IdentityProvider;
 	pub type IdentityOf<T> = <<T as Config>::IdentityProvider as IdentityProvider<<T as Config>::Identifier>>::Success;
 	pub type IdentityCommitmentVersion = u16;
```
Ad96el, line 40: For my learning: Why not create a type in the config that implements the default trait? Wouldn't it be cleaner? I often struggle to determine what should be included in the configuration and what should not. :confused:

ntn-x2: You mean using the latest Substrate features that allow to specify defaults for a pallet's `Config` trait? Or what do you mean?

pallets/pallet-deposit-storage/src/deposit.rs
```diff
@@ -0,0 +1,169 @@
...
+			FixedDepositCollectorViaDepositsPalletError::DepositAlreadyTaken => 0,
+			FixedDepositCollectorViaDepositsPalletError::DepositNotFound => 1,
+			FixedDepositCollectorViaDepositsPalletError::FailedToHold => 2,
+			FixedDepositCollectorViaDepositsPalletError::FailedToRelease => 3,
+			FixedDepositCollectorViaDepositsPalletError::Internal => u16::MAX,
+		}
+	}
+}
+
+impl<Runtime, DepositsNamespace, FixedDepositAmount> DipProviderHooks<Runtime>
```
trusch, line 65: Somehow it feels weird to see this tight coupling of the deposits and the dip pallet here. Shouldn't this logic also be passed in while configuring the pallet for a specific runtime like with the removal?

trusch (approved):
lgtm in general, just one question/remark

pallets/pallet-deposit-storage/src/deposit.rs
ntn-x2: Do you think the name of the type is misleading? This is already a concrete implementation of a trait of the DIP provider pallet, and it is supposed to work only when you are storing identity commitments and want to take deposits for them. I don't understand what you refer with the logic being passed? Can you elaborate on that?

trusch: I think the question is more from an architectural point of view: Why does something that is obviously specific to the DIP pallet (like this implementation of the DipProviderHooks) living within the deposit-storage pallet? Consequently this would mean that the deposit pallet code base would need to grow with every new "consumer" pallet that uses the deposit pallet. I think no action item should result from this comment, I'm only interested in your reasoning for structuring it this way.

ntn-x2: The problem is the Rust orphan rule. You can only implement a trait for a type if either of them is native to the crate. In this case, it would mean either implementing the trait in the pallet storage or in the pallet provider crate, as I see no point in having it in any other place. Would you rather have the type in the provider crate?

pallets/pallet-dip-provider/src/lib.rs
ntn-x2: I took a shot in https://github.com/KILTprotocol/kilt-node/pull/577. Resolving this, let me know there what you think!

PR #579, opened 2023-11-09
chore: prepare new dev version

Ad96el:
## Checklist:

    * Link the PR/Ticket here


Cargo.toml
```diff
@@ -6,7 +6,7 @@ homepage = "https://kilt.io/"
 license-file = "LICENSE"
 readme = "README.md"
 repository = "https://github.com/KILTprotocol/kilt-node"
-version = "1.12.0-dev"
+version = "1.13.0-dev"
```
ntn-x2, line 9: Also the spec version need to be bumped to 11300 so that type augmentation works properly.

Ad96el: Yes. you are right: [1af49a7](https://github.com/KILTprotocol/kilt-node/pull/579/commits/1af49a7b5569bab0799fb24fee9a4757df2533c4)

Ad96el:
I have also removed the migrations for the parachainStaking and the bump in storage versions. A minimal migration struct is still present, just in case we want to add a new pallet to the runtime and need to initialize the storage version.

Cargo.toml
ntn-x2: Can you do it also for the DIP runtime templates?

Ad96el: [0ebdac0](https://github.com/KILTprotocol/kilt-node/pull/579/commits/0ebdac0b8f5335b20f89cf201405ef1ad890c2ef)

ntn-x2: What's with all these changes? 😅

Ad96el: rebased: [e0c4a3d](https://github.com/KILTprotocol/kilt-node/pull/579/commits/e0c4a3d6747b54641ab25c43afbb6b248ebfd2f5)

ntn-x2 (approved):

PR #583, opened 2023-11-20
chore: fix docs and add check step in CI

ntn-x2:
Running `cargo doc` produces a bunch of warnings. Although we don't do much with them yet, it is nice that we make sure they are correct. Our test step already includes any snippets (we don't have any), but we don't check that the docs themselves are correct.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,11 +2,12 @@ stages:
   - test
   - build
 
-clippy:
+clippy-and-docs:
   image: paritytech/ci-unified:bullseye-1.70.0
   stage: test
   script:
     - cargo clippy --all-features --all-targets --locked -- -D warnings
+    - cargo doc --all-features --no-deps --locked
 
 fmt:
   image: paritytech/ci-unified:bullseye-1.70.0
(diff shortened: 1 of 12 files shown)


.gitlab-ci.yml
```diff
@@ -2,11 +2,12 @@ stages:
   - test
   - build
 
-clippy:
+clippy-and-docs:
   image: paritytech/ci-unified:bullseye-1.70.0
   stage: test
   script:
     - cargo clippy --all-features --all-targets --locked -- -D warnings
+    - cargo doc --all-features --no-deps --locked
```
weichweich, line 10: Were the errors you fixed here all found automatically? I don't think so right? I tried to put something in `<>` and it didn't complain.

weichweich (approved):

.gitlab-ci.yml
ntn-x2: Yes, I did not do anything more than running the command above. I had to decide how to address the warning, since it was not done automatically. I think the linter looks at what's inside the `<>`. What did you try it with? If you try to undo any of the changes to the docs, you should see that specific warning popping up again.

PR #587, opened 2023-11-23
feat: update templates and Dockerfile to support Zombienet

ntn-x2:
Zombienet uses the empty chainspec for a parachain, and that was not supported by the node templates. Also, the Dockerfile assumes the `kilt-parachain` is being run, but this is different now, and could also be different in the future, so the binary target should be parametrized: anyway, it is not necessary to compile the whole workspace if only a specific binary is required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,20 +5,21 @@ FROM paritytech/ci-unified:bullseye-1.70.0 as builder
 
 WORKDIR /build
 
-ARG FEATURES=default
+ARG BINARY=kilt-parachain
 
 COPY . .
 
-RUN cargo build --locked --release --features $FEATURES
+RUN cargo build --locked --release -p $BINARY
 
 # ===== SECOND STAGE ======
 
 FROM docker.io/library/ubuntu:20.04
-LABEL description="This is the 2nd stage: a very small image where we copy the kilt-parachain binary."
 
-ARG NODE_TYPE=kilt-parachain
+ARG BINARY=kilt-parachain
 
-COPY --from=builder /build/target/release/$NODE_TYPE /usr/local/bin/node-executable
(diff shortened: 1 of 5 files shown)


Ad96el (commented):
LGTM. Nice changes. I didn't even know you could load the chain spec from a JSON file. I also appreciate the Dockerfile changes, making it possible to compile a specific binary seems reasonable. I don't see any issues there.

Could we perhaps avoid renaming the ChainSpec trait? I can imagine this being something to handle when updating Polkadot dependencies. I don't really see any benefits.

Dockerfile
```diff
@@ -6,19 +6,21 @@ FROM paritytech/ci-unified:bullseye-1.70.0 as builder
 WORKDIR /build
 
 ARG FEATURES=default
+ARG BINARY=kilt-parachain
 
 COPY . .
 
-RUN cargo build --locked --release --features $FEATURES
+RUN cargo build --locked --release --features $FEATURES -p $BINARY
```
weichweich, line 13: FYI the reason to compile all binaries in the first stage was so that we only need to run the first stage once and only compile dependencies once. this would lead to a slightly longer first stage, but you only need to run it once.

ntn-x2: Where do we make use of that information? I only see a second stage being used, and we don't, for instance, use the same first stage to also build the mashnet-node. Am I wrong?

weichweich: We (at least try to) do that here: https://github.com/KILTprotocol/kilt-node/blob/develop/.maintain/build-image.sh
I would need to verify that this works. The last pipeline took ~80min

weichweich: Also this PR breaks this script.

ntn-x2: The first stage takes around 70 min anyway. But good to know. If that is the case, I will revert this commit before merging the DIP branch in `develop`.

PR #588, opened 2023-12-01
chore: balance test

Ad96el:
Checks the compatibility between reserves vs holds and freezes vs locks. 
 

## Checklist:

- [ ]  I have verified that the code works
    * Link the PR/Ticket here


ntn-x2 (commented):
Can we also test that trying to reserve when there's not enough coins fails? Same thing for holding, locking and freezing. Here we are testing the getters, but for the sake of completeness we should check whether the actual action of reserving tokens (which is what we rely on) actually works.

PR #592, opened 2023-12-11
chore: more improvements

ntn-x2:
Addresses most of the outstanding comments in https://github.com/KILTprotocol/kilt-node/pull/494. Which ones have been addressed are specified in that PR.


pallets/pallet-dip-provider/src/lib.rs
```diff
@@ -158,30 +158,27 @@ pub mod pallet {
...
-					)
-					.map_err(|e| Error::<T>::Hook(e.into()))?;
-					Self::deposit_event(Event::<T>::VersionedIdentityDeleted {
-						identifier: identifier.clone(),
-						version: commitment_version,
-					});
+			match Self::delete_identity_commitment_storage_entry(&identifier, commitment_version) {
+				// If an old commitment is found, call the relevant hook to deal with it.
+				Ok(commitment) => {
+					T::ProviderHooks::on_commitment_removed(&identifier, &dispatcher, &commitment, commitment_version)
```
weichweich, line 164: I think this should be inside the `delete_identity_commitment_storage_entry` function

weichweich (commented):
Minor remark.

Looks good otherwise! 🚀

crates/kilt-dip-primitives/src/export/mod.rs
```diff
@@ -16,21 +16,20 @@
...
-pub use child::{
-	DipChildProviderStateProofVerifierError, KiltVersionedChildProviderVerifier, VersionedChildParachainDipStateProof,
+pub use parachain::{
+	DipParachainStateProofVerifierError, KiltVersionedParachainVerifier, VersionedDipParachainStateProof,
 };
-pub use sibling::{
-	DipSiblingProviderStateProofVerifierError, KiltVersionedSiblingProviderVerifier,
-	VersionedSiblingParachainDipStateProof,
+pub use relaychain::{
+	DipRelaychainStateProofVerifierError, KiltVersionedRelaychainVerifier, VersionedRelaychainStateProof,
```
Ad96el, line 30: is this not a bit redundant? The mods are public or not?

pallets/pallet-dip-provider/src/lib.rs
ntn-x2: The function is called from the storage deposit hook: https://github.com/KILTprotocol/kilt-node/blob/c90b64b577407f661dd2fe810dfa537d73d3009f/dip-template/runtimes/dip-provider/src/dip.rs#L133. The only way to add it in there (I tried before already) is to remove the `submitter` parameter from the hook. What do you think?

ntn-x2: The deposit hook does not have information about the `submitter`. I thought about using the deposit owner as the submitter, but that felt wrong.

Ad96el (approved):
LGTM. Nice renamings/restructure. One minor comment.

crates/kilt-dip-primitives/src/export/mod.rs
ntn-x2: Perhaps. The goal is to allow a consumer to import directly from the root, and we export the most important types there.

pallets/pallet-dip-provider/src/lib.rs
weichweich: Mh you could also make the dispatcher optional. Having the deposit owner there might also work since they are the one who reclaimed the deposit right?

ntn-x2: Yes, that is an assumption we make in our code, and hopefully it holds all the time. I would go that way. Making the dispatcher optional now means that the deposit has been reclaimed, but not necessarily from someone. I will rename `submitter` to `owner` 👍

crates/kilt-dip-primitives/src/export/mod.rs
Ad96el: I see. I do not have a strong opinion on that. Discard the comment if you like.

Maybe the developer experience is better having shorter names. Instead of `DipParachainStateProofVerifierError` it could be renamed to `StateProofVerifierError`, which is imported from `parachain`? Just an idea.

ntn-x2: There are conflicts if the same type is re-exported from two different modules. I would move on for now, but won't close this issue. It will probably be part of the next batch of improvements.

pallets/pallet-dip-provider/src/lib.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/592/commits/ad3317a75146bd620e550dc5ffebf89dc0e3dd3d

weichweich (approved):

PR #594, opened 2023-12-12
feat: add DIP to Peregrine

ntn-x2:
Based on top of https://github.com/KILTprotocol/kilt-node/pull/494.

Mirrors the behaviour in the provider template, with the following differences:

1. The deposit is calculated taking into account the actual length of an identity commitment.
2. The key that is stored in the pallet deposit storage is an enum, which allows more keys to be added in the future


weichweich (approved):

pallets/pallet-dip-consumer/Cargo.toml
```diff
@@ -53,3 +53,5 @@ std = [
   "frame-benchmarking?/std",
   "sp-runtime?/std",
 ]
+
+try-runtime = []
```
Ad96el, line 57: `frame-system` and other pallets are missing. Can be done in another PR I guess. Just afraid that the try-runtime tests are not working right now. Did you verify that?

Ad96el (approved):
Two minor comments: The new pallets are not listed in the benchmarks. I can address this later in the release process. I don't want to delay the release. I can verify if the try-runtime tests are still functioning. Most likely, they are, but I have a slight uncertainty of 1%.

Otherwise, looks good

pallets/pallet-dip-consumer/Cargo.toml
ntn-x2: The pallet does not have any support for `try-runtime` yet. It was required for clippy not to complain. I guess once we add support for `try-runtime` (e.g., once we have a migration), then we will import the required dependencies?

ntn-x2: I mean for the runtime to compile, but in essence having the feature enabled has no effect on those pallets (yet).

PR #595, opened 2023-12-12
chore: add DIP templates Docker images

ntn-x2:
Partially fixes https://github.com/KILTprotocol/ticket/issues/3051. Needed for easier DIP-SDK integration tests. Binaries will have to be compiled and exported in a different PR.

### DIFF
--- a/.maintain/build-image.sh
+++ b/.maintain/build-image.sh
@@ -27,14 +27,14 @@ docker build \
 PROVIDER_BIN_NAME="dip-provider-node-template"
 docker build \
 	--cache-from $AWS_REGISTRY/kilt-parachain/collator:builder \
-	--cache-from $AWS_REGISTRY/kilt-parachain/collator:$target_tag \
+	--cache-from $AWS_REGISTRY/dip-provider-node-template:$target_tag \
 	--build-arg NODE_TYPE=$PROVIDER_BIN_NAME \
 	-t local/$PROVIDER_BIN_NAME:$target_tag \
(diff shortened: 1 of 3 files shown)


.maintain/build-image.sh
```diff
@@ -22,3 +22,19 @@ docker build \
     --build-arg NODE_TYPE=standalone-node \
     -t local/standalone-node:$target_tag \
     .
+
+# build DIP provider and consumer templates
+PROVIDER_BIN_NAME="dip-provider-node-template"
+docker build \
+	--cache-from $AWS_REGISTRY/kilt-parachain/collator:builder \
```
ntn-x2, line 29: Based on https://github.com/KILTprotocol/kilt-node/pull/587#discussion_r1407337975, I used the same base image for caching. Is that correct @weichweich?

.maintain/build-image.sh
```diff
@@ -22,3 +22,19 @@ docker build \
...
+	--cache-from $AWS_REGISTRY/dip-provider-node-template:$target_tag \
```
ntn-x2, line 30: Is this also correct? I have to take cache from the latest image of the same type pushed, and not from the base `kilt-parachain` image, right? The naming is a bit misleading, but I hope I got it right.

.maintain/build-image.sh
weichweich: LGTM

weichweich (approved):

PR #596, opened 2023-12-14
chore: maintain scripts

Ad96el:
adds missing pallets for maintain scripts


weichweich (approved):

scripts/run_benches_for_runtime.sh
```diff
@@ -34,6 +35,8 @@ pallets=(
 	pallet-xcm
 	parachain-staking
 	public-credentials
+	pallet-deposit-storage
```
ntn-x2, line 38: I think these should only be added for Peregrine, not Spiritnet, for now.

scripts/run_benches_for_runtime.sh
```diff
@@ -9,6 +9,7 @@ chain=$([ "$1" == "spiritnet" ] && echo "spiritnet-dev" || echo "dev")
 standard_args="--release --locked --features=runtime-benchmarks --bin=kilt-parachain"
 
 pallets=(
+	pallet-migration
```
ntn-x2, line 12: Same for this, no?

Ad96el: `pallet migration` is in Spiritnet and Peregrine present. It will be released in 1.12

scripts/run_benches_for_runtime.sh
Ad96el: The runtime can be set as an argument to this script. For spiritnet the command would fail, since the dip pallets are not in the runtime yet. 

I would add them since they will be part of 1.13, and I am afraid to forget it.

ntn-x2: It's just that few lines below more pallets are added if the runtime is peregrine. So we should stay consistent. Either we remove those lines, or we add the only-Peregrine pallets there.

Ad96el: Ah sorry. Its late: [717121b](https://github.com/KILTprotocol/kilt-node/pull/596/commits/717121bddfefcc2870449371aa7eb73c0917f07c)

PR #601, opened 2024-01-12
docs: Document safe use of Twox64Concat

weichweich:

<details>

```
```

</details>

<details>

```
```

</details>

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-deposit-storage/src/lib.rs
+++ b/pallets/pallet-deposit-storage/src/lib.rs
@@ -143,10 +143,11 @@ pub mod pallet {
 	/// Storage of all deposits. Its first key is a namespace, and the second
 	/// one the deposit key. Its value includes the details associated to a
 	/// deposit instance.
+	// SAFETY: The Namespace is not chosen by the user but by the pallet and therefore doesn't allow for arbitrary data.
 	#[pallet::storage]
 	#[pallet::getter(fn deposits)]
 	pub(crate) type Deposits<T> =
-		StorageDoubleMap<_, Twox64Concat, <T as Config>::Namespace, Twox64Concat, DepositKeyOf<T>, DepositEntryOf<T>>;
+		StorageDoubleMap<_, Twox64Concat, <T as Config>::Namespace, Blake2_128Concat, DepositKeyOf<T>, DepositEntryOf<T>>;
 
 	#[pallet::pallet]
 	#[pallet::storage_version(STORAGE_VERSION)]
--- a/pallets/pallet-dip-consumer/src/lib.rs
+++ b/pallets/pallet-dip-consumer/src/lib.rs
@@ -41,7 +41,6 @@ pub mod pallet {
 		dispatch::{Dispatchable, GetDispatchInfo, PostDispatchInfo},
 		pallet_prelude::*,
 		traits::{Contains, EnsureOriginWithArg},
-		Twox64Concat,
 	};
 	use frame_system::pallet_prelude::*;
 	use parity_scale_codec::{FullCodec, MaxEncodedLen};
@@ -108,7 +107,7 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn identity_proofs)]
 	pub(crate) type IdentityEntries<T> =
-		StorageMap<_, Twox64Concat, <T as Config>::Identifier, <T as Config>::LocalIdentityInfo>;
+		StorageMap<_, Blake2_128Concat, <T as Config>::Identifier, <T as Config>::LocalIdentityInfo>;
 
 	#[pallet::pallet]
 	#[pallet::storage_version(STORAGE_VERSION)]
--- a/dip-template/pallets/pallet-postit/src/lib.rs
+++ b/dip-template/pallets/pallet-postit/src/lib.rs
@@ -57,10 +57,12 @@ pub mod pallet {
 		type Username: Encode + Decode + TypeInfo + MaxEncodedLen + Clone + PartialEq + Debug + Default;
 	}
 
+	// Safety: The hash is calculated on chain. The hashing algorithm provided by frame_system::Config must be cryptographically secure.
 	#[pallet::storage]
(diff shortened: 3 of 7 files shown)


pallets/pallet-deposit-storage/src/lib.rs
```diff
@@ -143,10 +143,19 @@ pub mod pallet {
...
+	// doesn't allow for arbitrary data.
 	#[pallet::storage]
 	#[pallet::getter(fn deposits)]
-	pub(crate) type Deposits<T> =
-		StorageDoubleMap<_, Twox64Concat, <T as Config>::Namespace, Twox64Concat, DepositKeyOf<T>, DepositEntryOf<T>>;
+	pub(crate) type Deposits<T> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		<T as Config>::Namespace,
+		Blake2_128Concat,
```
weichweich, line 155: note this change. I think the DepositKey might be chosable by the user

pallets/pallet-dip-provider/src/lib.rs
```diff
@@ -85,6 +85,11 @@ pub mod pallet {
 	/// double map. Its first key is the `Identifier` of subjects, while the
 	/// second key is the commitment version. The values are identity
 	/// commitments.
+	// SAFETY of Twox64Concat:
```
ntn-x2, line 88: We can still change this, and simply disregard whatever will be written on Peregrine until then, what do you think? I think is's a more sane approach for the pallet to ensure safety here.

pallets/pallet-dip-consumer/src/lib.rs
```diff
@@ -105,6 +104,9 @@ pub mod pallet {
 	/// The pallet contains a single storage element, the `IdentityEntries` map.
 	/// It maps from a subject `Identifier` to an instance of
 	/// `LocalIdentityInfo`.
+	// SAFETY of Twox64Concat:
+	// The Identifier cannot be chosen freely it is therefore not efficient to
+	// craft specific hashes.
```
ntn-x2, line 109: Same comment as for the provider pallet. Do you think it's a better design if the pallet ensures safety by not letting the runtime mess up? Technically, the `IdentityEntries` is populated using the information in the origin. In the case of DIDs this is probably not possible, but in the case of other IDs this could still be possible by providing the right origin.

Alghouth, we have not made the same considerations for the runtime's generic `AccountId` tho...

ntn-x2, line 109: Same comment as for the provider pallet. Do you think it's a better design if the pallet ensures safety by not letting the runtime mess up? Technically, the `IdentityEntries` is populated using the information in the origin. In the case of DIDs this is probably not possible, but in the case of other IDs this could still be possible by providing the right origin.

Alghouth, we have not made the same considerations for the runtime's generic `AccountId` tho...

pallets/pallet-deposit-storage/src/lib.rs
ntn-x2: This follows the same reasoning as my other comments (not sure if above or below). The concrete type of a deposit key is defined by the runtime, but it could take user-specified values, so I am all in favor for this change.

pallets/did/src/lib.rs
```diff
@@ -334,6 +334,10 @@ pub mod pallet {
 	/// Service endpoints associated with DIDs.
 	///
 	/// It maps from (DID identifier, service ID) to the service details.
+	// SAFETY of Twox64Concat:
+	// The DID Identifier must be registered on chain first. To register a DID
+	// Identifier, you must provide a valid signature for the identifier or
+	// control the origin that corresponds to the identifier.
```
ntn-x2, line 340: How does this play out with cross-chain origins? I know probably migration would be hard here, but it might still be worth it.

ntn-x2 (commented):
The public credentials pallet also has a  `pub type Credentials<T> =
		StorageDoubleMap<_, Twox64Concat, SubjectIdOf<T>, Blake2_128Concat, CredentialIdOf<T>, CredentialEntryOf<T>>;` which, more than anything else, allows for a user-chosen `SubjectId`, which is typically an asset DID which, although it has a specific syntax, can still be crafted by the user.

pallets/did/src/lib.rs
weichweich: The DID that creates the endpoint still needs to be a valid accountId. If you would be able to choose an arbitrary accountId you could impersonate other accounts. So I don't think that this will ever be possible.

ntn-x2: No, but the resulting account ID is given by our location converter. There might be a way for a chain to craft origins which are then converted into something else on our chain. That would be the same as crafting origins on our chain.

weichweich: Creating specific AccountIds on our chain should never be possible? Even if the origin is coming from another chain.

ntn-x2: Not saying that. But to have an account on our chain you must have a valid private key to generate the signature. On a different chain, "trying" with different accounts might be cheaper, e.g., if, by absurd, you simply create a chain to spam all possible account IDs. This is probably very theoretical, but still possible. All those origins would be converted into some sort of account ID on our chain, but we don't really know what's the "proof of ownership" for that account ID on the source chain. Maybe this is somehow prevented by having XCM fees in place, which is also a mitigation in our case.

ntn-x2:
This will break cross-chain DIP proofs since the storage keys (and the related proofs) have changed, hence we might want to change at least the DIP stuff to make it resistant.

ntn-x2:
DIP stuff was fixed in https://github.com/KILTprotocol/kilt-node/commit/f4a94f0edb452a53a0ff42b8ddd1710e61f6a248.

ntn-x2:
I'm putting this on hold until we get to a polkadot-sdk version that supports multi-block-migrations, after which we can revise this PR and perform the necessary migrations.

PR #603, opened 2024-01-17
feat: open up XCM configurations

ntn-x2:
# Fixes https://github.com/KILTprotocol/ticket/issues/3082.

The following PR enables reserve transfers to other chains. Important DID pallets are whitelisted to authorize these extrinsics for execution within the transact instruction. Validation is required to ensure the correctness of the whitelisting, as the whitelisted call should not trigger another dispatch call.

The PR includes tests for the following scenarios:

- Reserve transfers to another parachain.
- Denial of reserve transfers to the relay chain.
- Denial of teleportation.
- Denial of sudo calls from other parachains/relay chains.
- Creation of a DID from another chain.


runtimes/peregrine/src/xcm_config.rs
```diff
@@ -98,8 +108,16 @@ pub type XcmBarrier = DenyThenTry<
...
+			call,
+			RuntimeCall::Did { .. }
+				| RuntimeCall::Ctype { .. }
+				| RuntimeCall::DidLookup { .. }
+				| RuntimeCall::Web3Names { .. }
+				| RuntimeCall::PublicCredentials { .. }
+				| RuntimeCall::Attestation { .. }
+		)
 	}
 }
```
Ad96el, line 122: I whitelisted all relevant DID pallets. According to the comment, I guess I should only allow the `create_from_account` extrinsic in the did pallet.

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -163,17 +181,22 @@ parameter_types! {
...
 
 impl pallet_xcm::Config for Runtime {
-	type MaxRemoteLockConsumers = ConstU32<10>;
-	type RemoteLockConsumerIdentifier = [u8; 8];
```
Ad96el, line 167: Is there a reason for this change? The spiritnet config has to be updated as well.

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,114 @@
...
+		let balance_after_xcm_call: u128 =
+			<<Peregrine as Parachain>::Balances as Inspect<AccountId>>::balance(&asset_hub_sovereign_account).into();
+
+		assert!(balance_after_xcm_call < init_balance);
+	});
+
+	// No event on the relaychain (message is meant for AssetHub)
+	Rococo::execute_with(|| {
+		assert_eq!(RococoSystem::events().len(), 0);
+	});
```
Ad96el, line 113: Ideally I would like to check, if the did is really created. But this is not as straight forward as anticipated. Will be checked in the e2e tests with chopsticks. The `reporterror` is blocked by the barrier of assethub.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: I think it's enough to ban anything recursive like `proxy` calls and `utility` batches.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: Yeah I did this when I was looking at how other parachains were doing it. So it should be changed with Spiritnet as well.

integration-tests/emulated/src/tests/peregrine/did.rs
ntn-x2: Can't you check on the provider chain state directly? You can capture the result of an `execute_with`.

Ad96el:
The compile error originates `from integration-tests-common`.

One of the crates where the error is thrown is `asset-hub-westend-runtime`.

The Cargo inverse tree provides the following logs:

```
asset-hub-westend-runtime v0.9.420 (https://github.com/paritytech/cumulus?branch=polkadot-v1.0.0#0d17cf6b)
└── integration-tests-common v1.0.0 (https://github.com/paritytech/cumulus?branch=polkadot-v1.0.0#0d17cf6b)
    └── xcm-integration-tests v1.12.0-dev (/home/adel/kilt-node/integration-tests/emulated)
    [dev-dependencies]
    └── xcm-integration-tests v1.12.0-dev (/home/adel/kilt-node/integration-tests/emulated)

```
It appears to be a bug in Substrate, which is already resolved in version [1.1.0](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/cumulus/parachains/integration-tests/emulated/common/Cargo.toml) (please take a look on the runtime-benchmarks feature).

For now, the best solution might be to remove the integration-tests-common dependency. It seems unnecessary to have this dependency specifically for tests.

runtimes/peregrine/src/xcm_config.rs
Ad96el: Alright will do. 

I was thinking of moving the xcm_config.rs into the runtime-common part. Spiritnet and Peregrine should be as close as possible together. I guess we want to have the same xcm config for both runtimes.

ntn-x2:
I mean we have a lot of constants that we use from that crate. We can also simply add broken dependency by hand, for me it's the `collectives-polkadot-runtime`. So add it to the root Cargo file, then add it as a dev dependency to the testing crate. That's one option. The other option is to manually re-declare all the constants we need from the broken pallet. I would prefer the first.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: Yes, please make sure everything else is left unchanged.

runtimes/peregrine/src/xcm_config.rs
Ad96el: Alright okay. I think it is also fine. Do you want to allow more? Otherwise, I would resolve this conversation.

Ad96el:
Thanks for pointing that out. I wasn't thinking of adding the broken dependencies in the cargo file. With the commit [30c5afb](https://github.com/KILTprotocol/kilt-node/pull/603/commits/30c5afb2209b150566ee6dd8f09f0f0235762be0), it should compile now.

integration-tests/emulated/src/tests/peregrine/did.rs
Ad96el: I couldn't find a way to check the state directly, but I was able to produce the Did create event: [bcc4a6a](https://github.com/KILTprotocol/kilt-node/pull/603/commits/bcc4a6a7c1c87ddf75dc47d24c52c26864f80829)

runtimes/spiritnet/src/xcm_config.rs
```diff
@@ -16,46 +16,48 @@
...
 /// `Transact`. There is an `OriginKind` which can bias the kind of local
 /// `Origin` it will become.
 pub type XcmOriginToTransactDispatchOrigin = (
-	// We don't include `SovereignSignedViaLocation<LocationToAccountId, RuntimeOrigin>` since we don't want to allow
-	// other chains to manage accounts on our network.
-
+	// Sovereign account converter; this attempts to derive an `AccountId` from the origin location
+	// using `LocationToAccountId` and then turn that into the usual `Signed` origin. Useful for
+	// foreign chains who want to have a local sovereign account on this chain which they control.
+	SovereignSignedViaLocation<LocationToAccountId<RelayNetworkId>, RuntimeOrigin>,
```
Ad96el, line 60: Albi was actively denying the sovereign account converter. I've noticed that it was allowed on other projects. I'll contact him to inquire about the reasoning behind his decision.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: It is a measure to limit PoV sizes, so really anything that does not have unbounded execution. As far as I know everything but nested calls are bounded, so it should be fine and future-proof.

The comment from the source code for v1.0.0 is:

```
/// Calls that are allowed through this filter must:
/// 1. Have a fixed weight;
/// 2. Cannot lead to another call being made;
/// 3. Have a defined proof size weight, e.g. no unbounded vecs in call parameters.
```

I am actually not sure why point 3 is there. AFAIK PoV is not affected by input length, but only by accessed storage size (see [here](https://substrate.stackexchange.com/questions/9264/what-is-proof-size)). Anyway, it makes sense to check if we take `Vec`s anywhere (we shouldn't), and exclude those calls as well.

Also, I read somewhere, next versions of Polkadot remove the need to have this filter configured, not sure which version that was tho.

integration-tests/emulated/src/tests/peregrine/did.rs
ntn-x2: That's good. I think it makes sense to check what the value of the event is. I.e., what's the expected DID identifier and what's the expected submitter? They will probably be the same in this case.

integration-tests/emulated/src/tests/peregrine/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[test]
+fn test_teleport_asset_from_regular_peregrine_account_to_asset_hub() {
+	MockNetworkRococo::reset();
+
+	let alice_account_id = get_account_id_from_seed::<sr25519::Public>(ALICE);
+	let bob_account_id = get_account_id_from_seed::<sr25519::Public>(BOB);
+
+	Peregrine::execute_with(|| {
+		assert_err!(
```
ntn-x2, line 26: This is not an `assert_noop!`?

integration-tests/emulated/src/tests/peregrine/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+						network: None,
+						id: bob_account_id.into()
+					})
+					.into()
+				),
+				Box::new((Here, 1000 * ED).into()),
+				0,
+				WeightLimit::Unlimited,
+			),
+			DispatchError::Module(ModuleError {
```
ntn-x2, line 41: This is the error of the xcm pallet. You can simply use that here, like `pallet_xcm::Error::<PeregrineRuntime>::Filtered`, no?

integration-tests/emulated/src/tests/peregrine/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+				error: [2, 0, 0, 0],
+				message: Some("Filtered")
+			})
+		);
+	});
+	// No event on the relaychain Message is for AssetHub
+	Rococo::execute_with(|| {
+		assert_eq!(Rococo::events().len(), 0);
+	});
+	// Fails on AssetHub since peregrine is not a trusted registrar.
```
ntn-x2, line 52: This is not really true. The message is not even sent to AssetHub.

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
```diff
@@ -0,0 +1,80 @@
...
+
+		assert_expected_events!(
+			AssetHubRococo,
+			vec![
+				RuntimeEvent::PolkadotXcm(pallet_xcm::Event::Sent { .. }) => {},
+			]
+		);
+	});
+
+	Peregrine::execute_with(|| {
```
ntn-x2, line 63: But this is not what we want, right? We want sudo messages from the relaychain to be executed on the parachain no matter what. I mean the relaychain has complete power over a parachain's runtime anyway.

ntn-x2: Or what was your reasoning behind this?

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
```diff
@@ -0,0 +1,80 @@
...
+	let call: DoubleEncoded<()> = <Spiritnet as Parachain>::RuntimeCall::System(frame_system::Call::set_code { code })
+		.encode()
+		.into();
+	let sudo_origin = <AssetHubPolkadot as Parachain>::RuntimeOrigin::root();
+	let parachain_destination: VersionedMultiLocation =
+		ParentThen(Junctions::X1(Junction::Parachain(spiritnet::PARA_ID))).into();
+
+	let weight_limit = WeightLimit::Unlimited;
+	let require_weight_at_most = Weight::from_parts(1000000000, 200000);
+	let origin_kind = OriginKind::Superuser;
```
ntn-x2, line 30: Can we add the same also for the other `OriginKind` variants? It should be fairly easy to add.

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+						network: None,
+						id: bob_account_id.into()
+					})
+					.into()
+				),
+				Box::new((Here, 1000 * ED).into()),
+				0,
+				WeightLimit::Unlimited,
+			),
+			DispatchError::Module(ModuleError {
```
ntn-x2, line 41: Same here for the error type.

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[test]
+fn test_teleport_asset_from_regular_spiritnet_account_to_asset_hub() {
+	MockNetworkPolkadot::reset();
+
+	let alice_account_id = get_account_id_from_seed::<sr25519::Public>(ALICE);
+	let bob_account_id = get_account_id_from_seed::<sr25519::Public>(BOB);
+
+	Spiritnet::execute_with(|| {
+		assert_err!(
```
ntn-x2, line 26: This should also be `assert_noop` probably.

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
```diff
@@ -0,0 +1,56 @@
...
+				error: [2, 0, 0, 0],
+				message: Some("Filtered")
+			})
+		);
+	});
+	// No event on the relaychain Message is for AssetHub
+	Polkadot::execute_with(|| {
+		assert_eq!(Polkadot::events().len(), 0);
+	});
+	// Fails on AssetHub since spiritnet is not a trusted registrar.
```
ntn-x2, line 52: False.

integration-tests/emulated/src/tests/spiritnet/sudo.rs
```diff
@@ -0,0 +1,163 @@
...
+	let call: DoubleEncoded<()> = <Spiritnet as Parachain>::RuntimeCall::System(frame_system::Call::set_code { code })
+		.encode()
+		.into();
+	let sudo_origin = <AssetHubPolkadot as Parachain>::RuntimeOrigin::root();
+	let parachain_destination: VersionedMultiLocation =
+		ParentThen(Junctions::X1(Junction::Parachain(spiritnet::PARA_ID))).into();
+
+	let weight_limit = WeightLimit::Unlimited;
+	let require_weight_at_most = Weight::from_parts(1600000000000, 200000);
+	let origin_kind = OriginKind::Superuser;
```
ntn-x2, line 106: Also here, all other `OriginKind`s as well.

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -74,17 +77,26 @@ pub type XcmOriginToTransactDispatchOrigin = (
...
-	DenyReserveTransferToRelayChain,
-	WithComputedOrigin<
+pub type XcmBarrier = TrailingSetTopicAsId<
+	DenyThenTry<
+		DenyReserveTransferToRelayChain,
 		(
-			// We allow everything from the relay chain if it was sent by the relay chain legislative (i.e., democracy
-			// vote). Since the relaychain doesn't own KILTs and missing fees shouldn't prevent calls from the
-			// relaychain legislative, we allow unpaid execution.
-			AllowTopLevelPaidExecutionFrom<ParentLegislative>,
```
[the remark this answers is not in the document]
ntn-x2: I did not see other projects doing it tho, not sure why.

ntn-x2 (commented):
Some preliminary comments, but it looks very good so far.

integration-tests/chopsticks/package.json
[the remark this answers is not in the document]
Ad96el: yes makes sense: [7799488](https://github.com/KILTprotocol/kilt-node/pull/603/commits/7799488888b31095fed389331442ba560125a1b2)

integration-tests/emulated/src/mock/para_chains.rs
```diff
@@ -0,0 +1,207 @@
...
+		genesis = spiritnet::genesis(),
+		on_init = (),
+		runtime = {
+			Runtime: spiritnet_runtime::Runtime,
+			RuntimeOrigin: spiritnet_runtime::RuntimeOrigin,
+			RuntimeCall: spiritnet_runtime::RuntimeCall,
+			RuntimeEvent: spiritnet_runtime::RuntimeEvent,
+			XcmpMessageHandler: spiritnet_runtime::XcmpQueue,
+			DmpMessageHandler: spiritnet_runtime::DmpQueue,
+			LocationToAccountId: LocationToAccountId<spiritnet_runtime::xcm_config::RelayNetworkId>,
```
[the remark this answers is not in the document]
Ad96el: Good point: [cf81073](https://github.com/KILTprotocol/kilt-node/pull/603/commits/cf810738e9aae7e28e2910c24fb0acb932100261)

integration-tests/emulated/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg(test)]
+pub mod mock;
```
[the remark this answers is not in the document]
Ad96el: At the beginning, I was assuming that maybe other projects could use the mock of spiritnet or peregrine for their emulated tests. 

[a34345a](https://github.com/KILTprotocol/kilt-node/pull/603/commits/a34345af3e02ca194a57af39c0b8132c6b7629f0)

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,106 @@
...
+#[test]
+fn test_did_creation_from_asset_hub() {
+	MockNetworkRococo::reset();
+
+	// create the sovereign account of AssetHub
+	let asset_hub_sovereign_account =
+		Peregrine::sovereign_account_id_of(Peregrine::sibling_location_of(AssetHubRococo::para_id()));
+
+	let call: DoubleEncoded<()> = <Peregrine as Parachain>::RuntimeCall::Did(did::Call::create_from_account {
+		authentication_key: did::did_details::DidVerificationKey::Account(asset_hub_sovereign_account.clone()),
```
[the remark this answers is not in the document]
Ad96el: Removed in: [f69bbd7](https://github.com/KILTprotocol/kilt-node/pull/603/commits/f69bbd7b05268767b0bf1dc87950287ebb06e3c2)

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,106 @@
...
+			fees: (Here, withdraw_balance).into(),
+			weight_limit: WeightLimit::Unlimited,
+		},
+		Transact {
+			origin_kind,
+			require_weight_at_most,
+			call,
+		},
+		// refund back the withdraw assets.
+		RefundSurplus,
```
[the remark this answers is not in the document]
Ad96el: Ah yeah sorry was for testing: [b7598f2](https://github.com/KILTprotocol/kilt-node/pull/603/commits/b7598f24df905c19948289e385f016311dded2e7)

integration-tests/emulated/src/mock/para_chains.rs
```diff
@@ -0,0 +1,207 @@
...
+		genesis = peregrine::genesis(),
+		on_init = (),
+		runtime = {
+			Runtime: peregrine_runtime::Runtime,
+			RuntimeOrigin: peregrine_runtime::RuntimeOrigin,
+			RuntimeCall: peregrine_runtime::RuntimeCall,
+			RuntimeEvent: peregrine_runtime::RuntimeEvent,
+			XcmpMessageHandler: peregrine_runtime::XcmpQueue,
+			DmpMessageHandler: peregrine_runtime::DmpQueue,
+			LocationToAccountId: LocationToAccountId<peregrine_runtime::xcm_config::RelayNetworkId>,
```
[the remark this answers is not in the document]
Ad96el: [cf81073](https://github.com/KILTprotocol/kilt-node/pull/603/commits/cf810738e9aae7e28e2910c24fb0acb932100261)

integration-tests/zombienet/.env-example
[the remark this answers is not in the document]
Ad96el: [c450875](https://github.com/KILTprotocol/kilt-node/pull/603/commits/c450875dd3595558754128147165c3d88951c6f2) :(

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,106 @@
...
+		);
+	});
+
+	Peregrine::execute_with(|| {
+		type PeregrineRuntimeEvent = <Peregrine as Parachain>::RuntimeEvent;
+		assert_expected_events!(
+			Peregrine,
+			vec![
+				PeregrineRuntimeEvent::XcmpQueue(cumulus_pallet_xcmp_queue::Event::Success { .. }) => {},
+				PeregrineRuntimeEvent::Did(did::Event::DidCreated(_, _)) => {},
```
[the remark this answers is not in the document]
Ad96el: Yes good point: [30e2dd4](https://github.com/KILTprotocol/kilt-node/pull/603/commits/30e2dd4de4ef862a19f1dbe8cc4cb0c0427e7e84)

runtimes/peregrine/src/xcm_config.rs
Ad96el: Yeah, but if we don't allow `dispatch_as`  the did pallets are not really useful. Other parachains can not use their DID to do any operation. 
 
 I think we have to implement our own origin converter which creates the double origin for specific runtime calls.

integration-tests/emulated/src/tests/peregrine/did.rs
Ad96el: [30e2dd4](https://github.com/KILTprotocol/kilt-node/pull/603/commits/30e2dd4de4ef862a19f1dbe8cc4cb0c0427e7e84)

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,106 @@
...
+	.into();
+
+	let sudo_origin = <AssetHubRococo as Parachain>::RuntimeOrigin::root();
+
+	let parachain_destination: VersionedMultiLocation =
+		ParentThen(Junctions::X1(Junction::Parachain(peregrine::PARA_ID))).into();
+
+	// the Weight parts are copied from logs.
+	let require_weight_at_most = Weight::from_parts(10_000_600_000_000, 200_000_000_000);
+	let origin_kind = OriginKind::SovereignAccount;
```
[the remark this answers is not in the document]
Ad96el: For `OriginKind::Native` the test would not fail. As you have said, we convert them in our config. The message will be submitted. Which error do you expect?

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -0,0 +1,106 @@
...
+				PeregrineRuntimeEvent::Did(did::Event::DidCreated(_, _)) => {},
+			]
+		);
+
+		// we also expect that the sovereignAccount of AssetHub has some coins now
+		let balance_after_xcm_call: u128 =
+			<<Peregrine as Parachain>::Balances as Inspect<AccountId>>::balance(&asset_hub_sovereign_account);
+
+		// since a did is created some of the free balance should now be on hold. Therefore the balance should be less.
+		assert!(balance_after_xcm_call < init_balance);
```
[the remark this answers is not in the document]
Ad96el: Was 2 lazy :see_no_evil: : [83abcdf](https://github.com/KILTprotocol/kilt-node/pull/603/commits/83abcdf2d33d56ddaf0bd96cdabe754ba51128f4)

integration-tests/emulated/src/tests/peregrine/sudo.rs
```diff
@@ -0,0 +1,167 @@
...
+			]
+		);
+	});
+
+	Peregrine::execute_with(|| {
+		type PeregrineRuntimeEvent = <Peregrine as Parachain>::RuntimeEvent;
+
+		let events = Peregrine::events();
+
+		println!("{:?}", events);
```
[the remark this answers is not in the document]
Ad96el: [7f1016c](https://github.com/KILTprotocol/kilt-node/pull/603/commits/7f1016cec4f75e682cccf3c5a78d16f5948bdd1a)

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
Ad96el: yeah would be better: [4d6b9af](https://github.com/KILTprotocol/kilt-node/pull/603/commits/4d6b9af30851c78404e5d711ffaae561188fef33)

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
Ad96el: [4d6b9af](https://github.com/KILTprotocol/kilt-node/pull/603/commits/4d6b9af30851c78404e5d711ffaae561188fef33)

integration-tests/emulated/src/tests/peregrine/teleportation.rs
Ad96el: yeah good suggestion: [4d6b9af](https://github.com/KILTprotocol/kilt-node/pull/603/commits/4d6b9af30851c78404e5d711ffaae561188fef33)

integration-tests/emulated/src/tests/peregrine/teleportation.rs
Ad96el: [7f1016c](https://github.com/KILTprotocol/kilt-node/pull/603/commits/7f1016cec4f75e682cccf3c5a78d16f5948bdd1a)

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
Ad96el: I do not get your point. The test case checks if unpaid transaction from another parachain are blocked by the barrier. Messages from the relay chain are always unpaid and will be executed.

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
Ad96el: The unpaid execution should not care for the `OriginKind`. I simplified the tests: [4b5526b](https://github.com/KILTprotocol/kilt-node/pull/603/commits/4b5526b10cc8ee1aff1f4b2eba6417c5b6fd226b). 

I hope it is clearer now.

integration-tests/emulated/src/tests/spiritnet/sudo.rs
Ad96el: This would be a bit pointless, no? If I select another `OriginKind`, it's obvious that the origin on the target chain isn't transformed to root, which is what I intended to test here.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: Or rather I would only check that it calls one of `ctype`, `attestation`, `delegation`, `web3name`, and `did` management operations, to start with.

runtimes/spiritnet/src/xcm_config.rs
ntn-x2: Any new info on this?

integration-tests/emulated/src/mock/network.rs
```diff
@@ -0,0 +1,32 @@
...
+};
+use xcm_executor::traits::ConvertLocation;
+
+use crate::mock::{
+	para_chains::{AssetHubPolkadot, AssetHubRococo, Peregrine, Spiritnet},
+	relay_chains::{Polkadot, Rococo},
+};
+
+decl_test_networks! {
+	pub struct MockNetworkPolkadot {
```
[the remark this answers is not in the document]
ntn-x2: We have different crates for Spiritnet and Peregrine, so it makes sense we have separate crates also to test those, even just for compile times, if we want to run tests only for a specific deployment. Also in my opinion it makes more sense to mock a full deployment and not put many things together.

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -52,16 +52,21 @@ parameter_types! {
 		Parachain(ParachainInfo::parachain_id().into()).into();
 }
 
+/// This type specifies how a `MultiLocation` can be converted into an `AccountId` within the
+/// Peregrine network, which is crucial for determining ownership of accounts for asset transactions
+/// and for dispatching XCM `Transact` operations.
+pub type PeregrineLocationToAccountId = LocationToAccountId<RelayNetworkId>;
```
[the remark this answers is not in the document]
ntn-x2: Same for the Spiritnet one.

integration-tests/emulated/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: When testing the DIP components I realised that cross-crate sharing of mocks is not really a good idea. Each crate should be self-contained, so it does not make sense to have a `pub` mock for this. Good we make it private again 👍

integration-tests/emulated/src/tests/peregrine/did.rs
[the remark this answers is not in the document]
ntn-x2: The message will be submitted, but its execution will still fail, right? The pallet's call expects a `Signed` origin, while the `OriginKind::Native` gives a `Parachain` origin. Or does it actually go all the way through?

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -93,12 +93,13 @@ fn test_did_creation_from_asset_hub() {
...
-			<<Peregrine as Parachain>::Balances as Inspect<AccountId>>::balance(&asset_hub_sovereign_account);
+		let balance_on_hold = <<Spiritnet as Parachain>::Balances as Inspect<AccountId>>::balance_on_hold(
+			&peregrine_runtime::RuntimeHoldReason::from(did::HoldReason::Deposit),
+			&asset_hub_sovereign_account,
+		);
 
-		// since a did is created some of the free balance should now be on hold. Therefore the balance should be less.
-		assert!(balance_after_xcm_call < init_balance);
+		// since a did is created, 2 of the free balance should now be on hold
+		assert_eq!(balance_on_hold, UNIT * 2);
```
ntn-x2, line 102: We can technically use the constant for the DID deposit to match against here.

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
ntn-x2: Your comment tricked me: `//Send XCM message from relay chain`. But yeah, this is testing the message from AssetHub.

ntn-x2: Although we should add operations from the relaychain as well.

integration-tests/emulated/src/tests/spiritnet/sudo.rs
ntn-x2: It is obvious to who? It's internal logic of the destination parachain to transform a (location, origin_kind) tuple into a local origin, and there's technically nothing stopping us from (mistakenly) transform anything unwanted to the root origin. It would be a very bad mistake, but it can happen. Hence I think a test covering that would go a long way.

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
ntn-x2: Resolving this in favor of https://github.com/KILTprotocol/kilt-node/pull/603#discussion_r1524499307.

integration-tests/emulated/src/tests/peregrine/did.rs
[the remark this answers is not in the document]
Ad96el: No, it does not pass. I do not receive the did create event by changing the `OriginKind`. I will add them :+1:

integration-tests/emulated/src/tests/spiritnet/sudo.rs
ntn-x2: In pratice we want to tests that under no circumstances a call to a pallet that requires a `sudo` origin can be dispatched over XCM.

ntn-x2: In pratice we want to tests that under no circumstances a call to a pallet that requires a `sudo` origin can be dispatched over XCM.

ntn-x2: In pratice we want to test that under no circumstances a call to a pallet that requires a `sudo` origin can be dispatched over XCM.

Ad96el: I see okay I will add the other test cases.

Ad96el: I added tests cases for all possible origin kinds: [5f86edf](https://github.com/KILTprotocol/kilt-node/pull/603/commits/5f86edfa7932e4cd58174f351447b35c2c61c29a)

integration-tests/emulated/src/mock/network.rs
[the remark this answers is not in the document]
Ad96el: Yes okay let's tackle this in a different PR

integration-tests/emulated/src/mock/para_chains.rs
```diff
@@ -0,0 +1,207 @@
...
+use integration_tests_common::constants::{accounts, asset_hub_polkadot, polkadot::ED};
+use runtime_common::{xcm_config::LocationToAccountId, AuthorityId};
+use sp_core::sr25519;
+use sp_runtime::{BuildStorage, Storage};
+use xcm_emulator::{decl_test_parachains, BridgeMessageHandler, Parachain, TestExt};
+
+use crate::utils::{get_account_id_from_seed, get_from_seed};
+
+const SAFE_XCM_VERSION: u32 = xcm::prelude::XCM_VERSION;
+pub mod spiritnet {
```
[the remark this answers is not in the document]
Ad96el: Okay, let's do it in a different PR

runtimes/peregrine/src/xcm_config.rs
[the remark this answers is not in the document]
Ad96el: renamed in: [57d564f](https://github.com/KILTprotocol/kilt-node/pull/603/commits/57d564f50578c0b2d332572f77e3d98f5625ac2b)

integration-tests/emulated/src/tests/spiritnet/sudo.rs
Ad96el: I added here all pssible origin kinds: [5f86edf](https://github.com/KILTprotocol/kilt-node/pull/603/commits/5f86edfa7932e4cd58174f351447b35c2c61c29a)

integration-tests/emulated/src/tests/peregrine/did.rs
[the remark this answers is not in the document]
Ad96el: added here the additional test cases: [f042871](https://github.com/KILTprotocol/kilt-node/pull/603/commits/f042871b5938544cdec4efea5779f6709b758046)

integration-tests/emulated/src/tests/peregrine/did.rs
Ad96el: Sorry, I mixed up 2 commits. But yes thats a good hint: [d30d226](https://github.com/KILTprotocol/kilt-node/pull/603/commits/d30d22658eb2f0ba7c464b8d050d2222914a8da2)

runtimes/spiritnet/src/xcm_config.rs
Ad96el: yes, it is fine to extend the converter. The comment was from William. I will mark this conversation as resolved.

integration-tests/emulated/src/mock/para_chains.rs
```diff
@@ -0,0 +1,207 @@
...
+			XcmpMessageHandler: spiritnet_runtime::XcmpQueue,
+			DmpMessageHandler: spiritnet_runtime::DmpQueue,
+			LocationToAccountId: LocationToAccountId<spiritnet_runtime::xcm_config::RelayNetworkId>,
+			System: spiritnet_runtime::System,
+			Balances: spiritnet_runtime::Balances,
+			ParachainSystem: spiritnet_runtime::ParachainSystem,
+			ParachainInfo: spiritnet_runtime::ParachainInfo,
+		},
+		pallets_extra = {
+			Did: spiritnet_runtime::Did,
```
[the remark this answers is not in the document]
Ad96el: I think with the `pallets_extra` you can extend more important pallets. For instance, if would not include the DID pallet here, I would not get any Did related events.

integration-tests/emulated/src/tests/peregrine/reserve_transfers.rs
```diff
@@ -0,0 +1,125 @@
...
+			<<Peregrine as Parachain>::Balances as Inspect<AccountId>>::balance(&asset_hub_sovereign_account);
+
+		assert_eq!(balance_after_transfer, balance_to_transfer);
+	});
+	// No event on the relaychain (message is meant for AssetHub.
+	Rococo::execute_with(|| {
+		assert_eq!(Rococo::events().len(), 0);
+	});
+	// Fails on AssetHub since peregrine is not a trusted registrar
+	AssetHubRococo::execute_with(|| {
```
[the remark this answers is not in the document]
Ad96el: Yes, this is expected. In reserved transfers, the coin has to live somewhere trusted. In the case of hydradx, they allowed spiritnet as a trusted registrar for the coins. 

Assethub is not allowing reserved transfers at all. That's why It will fail here.

integration-tests/emulated/src/tests/peregrine/teleportation.rs
Ad96el: the comments are updated here: [ab54319](https://github.com/KILTprotocol/kilt-node/pull/603/commits/ab543193f419feab2881cf04a1015b860e2fcb58)

integration-tests/emulated/src/tests/spiritnet/teleportation.rs
Ad96el: updated in: [ab54319](https://github.com/KILTprotocol/kilt-node/pull/603/commits/ab543193f419feab2881cf04a1015b860e2fcb58)

integration-tests/emulated/src/mock/network.rs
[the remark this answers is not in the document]
ntn-x2: Please make a ticket for it.

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
ntn-x2: Ok, and what goes into `pallets` compared to `pallets_extra`?

integration-tests/emulated/src/tests/peregrine/did.rs
```diff
@@ -56,19 +45,36 @@ fn test_did_creation_from_asset_hub() {
 			require_weight_at_most,
 			call,
 		},
-	]));
+	]))
+}
+
+fn get_destination() -> VersionedMultiLocation {
```
ntn-x2: Rather `get_peregrine_as_sibling_location`, I'd say.

ntn-x2 (commented):
I feel quite more confident now this works as expected. I'll review once again top-to-bottom when all outstanding comments will be addressed 👍

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
Ad96el: There is no `pallets` in the macro call. It is probably renamed to `pallets_extra`

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -163,17 +183,22 @@ parameter_types! {
...
+	// prepended to the message.
 	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmRouter = XcmRouter;
+	// We allow execution of XCM programs because it is required by the
+	// `reserve_transfer_assets` operation.
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
-	// Disable dispatchable execution on the XCM pallet.
+	// Disable calls to pallet_xcm::execute(), by still allowing the other
+	// extrinsics to be called.
 	// NOTE: For local testing this needs to be `Everything`.
```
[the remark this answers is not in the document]
Ad96el: Removed in: [3b06313](https://github.com/KILTprotocol/kilt-node/pull/603/commits/3b063135b2e8a0bad834f56450352f3a8f8a5dff)

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -163,17 +183,22 @@ parameter_types! {
 }
 
 impl pallet_xcm::Config for Runtime {
-	type MaxRemoteLockConsumers = ConstU32<10>;
-	type RemoteLockConsumerIdentifier = [u8; 8];
+	type MaxRemoteLockConsumers = ConstU32<0>;
+	type RemoteLockConsumerIdentifier = ();
 	type RuntimeEvent = RuntimeEvent;
+	// Allows anyone to send XCM messages. For regular origins, a `DescendOrigin` is
+	// prepended to the message.
```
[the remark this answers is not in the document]
Ad96el: Updated in: [3b06313](https://github.com/KILTprotocol/kilt-node/pull/603/commits/3b063135b2e8a0bad834f56450352f3a8f8a5dff)

runtimes/peregrine/src/xcm_config.rs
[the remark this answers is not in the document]
Ad96el: Updated in: [3b06313](https://github.com/KILTprotocol/kilt-node/pull/603/commits/3b063135b2e8a0bad834f56450352f3a8f8a5dff)

integration-tests/emulated/src/mock/network.rs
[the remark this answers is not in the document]
Ad96el: Ticket created: https://github.com/KILTprotocol/ticket/issues/3226

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
Ad96el: Ticket created: https://github.com/KILTprotocol/ticket/issues/3226

runtimes/peregrine/src/xcm_config.rs
Ad96el: the `SafecallFilter` is updated in[2387069](https://github.com/KILTprotocol/kilt-node/pull/603/commits/2387069acf5cfc76035ba8535f0e69e896dc04fc)

integration-tests/emulated/src/tests/peregrine/did.rs
Ad96el: It is updated, but I can not find the commit sorry :smile:

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
```diff
@@ -0,0 +1,142 @@
...
+		assert_eq!(Rococo::events().len(), 0);
+	});
+}
+
+// TODO: Check why test is passing. Unpaid execution should work now.
+#[test]
+fn test_unpaid_execution_from_rococo_to_peregrine() {
+	MockNetworkRococo::reset();
+
+	let sudo_origin = <Rococo as RelayChain>::RuntimeOrigin::root();
```
Ad96el, line 90: There appears to be an issue here. Since unpaid execution is now allowed, I need to establish a governance origin.

When I executed the test case with trace logs, I received this log message: `xcm::barriers: AllowUnpaidExecutionFrom origin: MultiLocation { parents: 1, interior: Here }, ...`

In the `WithComputedOrigin` step, the origin is the same:  `origin: xcm::barriers: WithComputedOrigin origin: MultiLocation { parents: 1, interior: Here }, ...`

@ntn-x2 Do you maybe know, how I can create a proper origin?

Ad96el: Here are the logs: 

image

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
ntn-x2: Sorry, I clearly meant the `runtime` part.

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -115,16 +115,45 @@ pub type XcmBarrier = TrailingSetTopicAsId<
...
+							| did::Call::change_deposit_owner { .. }
+							| did::Call::reclaim_deposit { .. }
+						)
+				)
+		}
+		
+
+		match c {
+			RuntimeCall::Did (c) => match c {
+				did::Call::dispatch_as { call, did_identifier: _ } => is_call_allowed(call),
```
ntn-x2, line 150:
```suggestion
				did::Call::dispatch_as { call, .. } => is_call_allowed(call),
```

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -115,16 +115,45 @@ pub type XcmBarrier = TrailingSetTopicAsId<
...
+				did::Call::submit_did_call { did_call, signature: _ } => is_call_allowed(&did_call.call),
```
ntn-x2, line 151:
```suggestion
				did::Call::submit_did_call { did_call, .. } => is_call_allowed(&did_call.call),
```

runtimes/peregrine/src/xcm_config.rs
```diff
@@ -115,16 +115,45 @@ pub type XcmBarrier = TrailingSetTopicAsId<
...
+				_ => true
```
ntn-x2, line 152: This should rather be `_ => is_call_allowed(c)` as well, right? So we make sure we match against the expected allowed DID calls.

runtimes/spiritnet/src/xcm_config.rs
```diff
@@ -111,16 +111,45 @@ pub type XcmBarrier = TrailingSetTopicAsId<
...
+						)
+				)
+		}
+		
+
+		match c {
+			RuntimeCall::Did (c) => match c {
+				did::Call::dispatch_as { call, did_identifier: _ } => is_call_allowed(call),
+				did::Call::submit_did_call { did_call, signature: _ } => is_call_allowed(&did_call.call),
+				_ => true
```
ntn-x2, line 151: Same here.

runtimes/peregrine/src/xcm_config.rs
ntn-x2: Or maybe `_ => is_call_allowed(c.into())` if `is_call_allowed` expects a `RuntimeCall`, when `c` is a `Did::call`.

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
ntn-x2: Using the root location, gives parachain location, of course. This is equivalent to dispatching a call from the root track. To have the `ParentLegislative` origin, we would have to have a referendum on the relaychain. I think that's too much. Maybe we can go with the relay root origin directly, for now. I will check that. We need to investigate more how opengov origins get converted into `MultiLocation`s.

ntn-x2: Yeah the problem is the use of `.encode()`, which generates a vector that is prefixed by the length of the vector itself, which of course makes for an invalid asset DID. Currently we don't have a way to convert an asset DID to a raw vector, as far as I could see, so the test case needs to be constructed by hand, as I did here: https://github.com/KILTprotocol/kilt-node/pull/603/commits/b2cb68a105cf83da6b0e30ec31dbc6c5c597e91a.

integration-tests/emulated/src/tests/spiritnet/did_pallets/public_credentials.rs
```diff
@@ -0,0 +1,185 @@
...
+};
+
+fn get_xcm_message_add_public_credential(
+	origin_kind: OriginKind,
+	withdraw_balance: Balance,
+	ctype_hash: H256,
+) -> VersionedXcm<()> {
+	let asset_hub_sovereign_account = get_asset_hub_sovereign_account();
+
+	let subject_id = AssetIdentifier::ether_currency();
```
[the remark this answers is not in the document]
ntn-x2: Yeah the problem is the use of .encode(), which generates a vector that is prefixed by the length of the vector itself, which of course makes for an invalid asset DID. Currently we don't have a way to convert an asset DID to a raw vector, as far as I could see, so the test case needs to be constructed by hand, as I did here: https://github.com/KILTprotocol/kilt-node/commit/b2cb68a105cf83da6b0e30ec31dbc6c5c597e91a.

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
ntn-x2: Using the root location, gives parachain location, of course. This is equivalent to dispatching a call from the root track. To have the `ParentLegislative` origin, we would have to have a referendum on the relaychain. Also, we would be testing with stuff that does not exist anymore, since all relaychains have migrated to opengov by now. Maybe we can go with the relay root origin directly, for now. I will check that. We need to investigate more how opengov origins get converted into `MultiLocation`s, and adjust our configuration accordingly.

ntn-x2: I changed the barrier here, but tests need to be updated: https://github.com/KILTprotocol/kilt-node/pull/603/commits/d9daae88e82c9693215ee1dffd6712c5a263e705.

runtimes/peregrine/src/xcm_config.rs
Ad96el: I've already matched the call against the DID pallet and verified that it is neither dispatch_as nor submit_did_call. All other calls within the DID pallet are permitted. Still updated in: [7305135](https://github.com/KILTprotocol/kilt-node/pull/603/commits/73051351de08551eb1173cad2d82dadd52fb2998)

runtimes/peregrine/src/xcm_config.rs
Ad96el: Updated in: [7305135](https://github.com/KILTprotocol/kilt-node/pull/603/commits/73051351de08551eb1173cad2d82dadd52fb2998)

runtimes/peregrine/src/xcm_config.rs
Ad96el: Updated in: [7305135](https://github.com/KILTprotocol/kilt-node/pull/603/commits/73051351de08551eb1173cad2d82dadd52fb2998)

runtimes/spiritnet/src/xcm_config.rs
Ad96el: Updated in: [7305135](https://github.com/KILTprotocol/kilt-node/pull/603/commits/73051351de08551eb1173cad2d82dadd52fb2998)

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
```diff
@@ -80,3 +81,55 @@ fn test_unpaid_execution_to_spiritnet() {
...
+fn test_unpaid_execution_from_polkadot_to_spiritnet() {
+	MockNetworkPolkadot::reset();
+
+	let sudo_origin = <Polkadot as RelayChain>::RuntimeOrigin::root();
+	let parachain_destination: VersionedMultiLocation = Junctions::X1(Junction::Parachain(spiritnet::PARA_ID)).into();
+
+	let weight_limit = WeightLimit::Unlimited;
+	let check_origin = None;
+
+	let xcm = VersionedXcm::from(Xcm(vec![UnpaidExecution {
```
ntn-x2, line 95: Can we add another instruction after this one, to make sure the operation goes through? I guess a system remark is fine.

integration-tests/emulated/src/mock/para_chains.rs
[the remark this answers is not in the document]
Ad96el: I assume the runtime parts defines pallets, which are mandatory in order to emulate xcm messaging between the chains. A runime which is missing one of the pallets/structs can not be emulated.

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
Ad96el: yeah I could not use system remark or any other extrinsic. We only whitelisted did related extrinsic. So I adjusted the testcase in such a way, that the sovereign account of polkadot wants to create a did: [5699a44](https://github.com/KILTprotocol/kilt-node/pull/603/commits/5699a44b4c96a8b62e63b792a5a76f9cddf33529)

Ad96el:
I think a lot of aspects can still be improved in this PR. Because of time constraints, we should move them back for now. To address these issues, I created this ticket: https://github.com/KILTprotocol/ticket/issues/3241

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
```diff
@@ -124,8 +154,18 @@ fn test_unpaid_execution_from_rococo_to_peregrine() {
...
 					..
 				}) => {},
+				PeregrineRuntimeEvent::Did(did::Event::DidCreated(account, did_identifier)) => {
+					account: account == &polkadot_sovereign_account,
+					did_identifier:  did_identifier == &polkadot_sovereign_account,
+				},
 			]
 		);
+
+		// Since the user have not paid any tx fees, we expect that the free balance is the 2*ED
```
ntn-x2, line 164:
```suggestion
		// Since the user have not paid any tx fees, we expect that the free balance is the ED
```

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
```diff
@@ -124,8 +155,18 @@ fn test_unpaid_execution_from_polkadot_to_spiritnet() {
...
 					..
 				}) => {},
+				SpiritnetRuntimeEvent::Did(did::Event::DidCreated(account, did_identifier)) => {
+					account: account == &polkadot_sovereign_account,
+					did_identifier:  did_identifier == &polkadot_sovereign_account,
+				},
 			]
 		);
+
+		// Since the user have not paid any tx fees, we expect that the free balance is the 2*ED
```
ntn-x2, line 165:
```suggestion
		// Since the user have not paid any tx fees, we expect that the free balance is the ED
```

ntn-x2 (commented):
I can only comment, not approve, since it's my own PR 😄 But LGTM ✅✅✅

integration-tests/emulated/src/tests/peregrine/unpaid_execution.rs
Ad96el: [44863e4](https://github.com/KILTprotocol/kilt-node/pull/603/commits/44863e4136c3a1016f14a547908fe0146acaa586)

integration-tests/emulated/src/tests/spiritnet/unpaid_execution.rs
Ad96el: [44863e4](https://github.com/KILTprotocol/kilt-node/pull/603/commits/44863e4136c3a1016f14a547908fe0146acaa586)

Ad96el (dismissed):

Ad96el (approved):

PR #607, opened 2024-02-09
chore: remove unused deps

ntn-x2:
WIP.

## Checklist

Make sure the following combinations work as expected, as they are the most useful ones:

- [ ] `--all-targets`: no default features, this is the final WASM that is deployed
- [ ] `--all-targets --all-features`: all features enabled, which can reduce our compile times by eliminating unnecessary dependencies


ntn-x2:
Problem I'm having is with a dev dependency that is only used when the benchmarks are being tested. This is currently not possible to remove, hence I was thinking to either add a new feature and make the dependency a regular optional dependency instead of a dev one, or then just try what other tools like cargo-udeps and cargo-machete yield. This is low priority but still somehow important to reduce the size of our WASM, binaries and bring down compilation times.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ members = [
   "runtimes/*",
   "support",
 ]
+resolver = "2"
```
weichweich, line 23: [This shouldn't be necessary](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html) since it's implied by `edition = "2021"`

ntn-x2: Good point, will remove that. Thanks for the heads up

Dudleyneedham:
@Antonio what is the status of this?

ntn-x2:
Will pick it up again at some point, probably next week.

ntn-x2:
Closing for now. Will make a proper ticket to track this.

PR #608, opened 2024-02-12
feat: increase service URL size limit

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3131.

The ticket mentions that 500 would be a good new limit for the inbound request we had. Reading elsewhere online, such as [here](https://stackoverflow.com/a/417184) it seems that 2000 characters is a good upper bound limit for URLs. Since every change would require a runtime upgrade, I would rather make the value future proof and set it to 2000 directly.


Ad96el (approved):

Dudleyneedham (approved):
LGTM

weichweich (approved):

PR #610, opened 2024-02-15
docs: Bring docs changes into DIP read me files

ChrisChinchilla:
## fixes KILTProtocol/ticket#2976

Docs: Tweaks some grammar and readability to the DIP pallet readmes so they can be included in docs.

### DIFF
--- a/pallets/pallet-dip-consumer/README.md
+++ b/pallets/pallet-dip-consumer/README.md
@@ -1,60 +1,81 @@
 # Decentralized Identity Provider (DIP) provider consumer pallet
 
-This pallet is a core component of the Decentralized Identity Provider protocol.
+This pallet is a component of the DIP protocol.
 It enables entities with an identity on a connected Substrate-based chain (provider) to use those identities on the chain this pallet is deployed (consumers) without requiring those entities to set up a new identity locally.
-A consumer chain is *connected* to a provider if there is a way for the consumer chain to verify state proofs about parts of the state of the provider chain.
+
+A consumer chain is _connected_ to a provider if there is a way for the consumer chain to verify proofs about parts of the state of the provider chain.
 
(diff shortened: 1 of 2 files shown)


ntn-x2 (commented):
Looks good! These changes would be applied to `develop`, which is ok, but we branched off for the upcoming release and `develop` will soon have changes that will diverge from the release content. How is that taken into account in the docs? Does it pull the READMEs from a specific branch/tag? In that case, we will need to cherry pick these changes onto the release branch as well.

ntn-x2:
Also please update the PR title to follow conventional commits (the CI fails due to that): https://www.conventionalcommits.org/en/v1.0.0/.
This is probably a `docs:` PR.

ChrisChinchilla:
@ntn-x2 At the moment, the plugin, which I have added here https://github.com/KILTprotocol/docs/pull/291 just pulls a raw file, but I think I can change the URL to be branches of tags instead. However, we need to setup versioning in the docs too to support this, which is tracked in https://github.com/KILTprotocol/ticket/issues/3149 so for now it will have to just be "develop", but that can change… soon ™️ . Short term not ideal, but I think we can track it manually until then.

ntn-x2:
Mhhh then I'd rather have the docs point to the release branch than to develop. I'm saying this because there'll soon be few PRs which make the docs outdated.

ChrisChinchilla:
@ntn-x2 Release 1.12.1?

ChrisChinchilla:
@ntn-x2 OK, done! I'll now fix up the other end.

ntn-x2:
@Ad96el @weichweich do you think we can merge this commit on the release branch and then cherry-pick it on `develop`? I am not sure if the changed README changes the metadata of the WASM, since it's used as the crate top-level documentation as well.

ntn-x2:
@ChrisChinchilla can you check if anything needs to be updated, and then we can merge this?

ChrisChinchilla:
@ntn-x2 Oh! i thought it already was! Nothing has changed that affects this PR, no.

PR #613, opened 2024-03-06
fix: change DIP components storage hashers

ntn-x2:
Changing the storage hasher before we deploy on production, otherwise we would then need a migration strategy. I think 
we can disregard any migrations on our testnets, as the effort is not justified.

I also lowered the limits of the DIP provider template, which were higher than Peregrine, so that it is faster to generate the benchmark worst case.

Related to https://github.com/KILTprotocol/kilt-node/pull/601, based on top of https://github.com/KILTprotocol/kilt-node/pull/612.


Ad96el (approved):
LGTM. I also think a migration for your test net is not necessary, but I would probably delete the 2 or 3 commitments via sudo rights. Doesn't feel right to have 2 broken commitments in the storage.

ntn-x2:
On hold until https://github.com/KILTprotocol/kilt-node/pull/612 is reviewed.

PR #614, opened 2024-03-21
chore: integration tests chopsticks

Ad96el:
## fixes [KILTProtocol/ticket#3098](https://github.com/KILTprotocol/ticket/issues/3098) and fixes https://github.com/KILTprotocol/ticket/issues/3239 and fixes https://github.com/KILTprotocol/ticket/issues/3239.

This PR introduces e2e tests based on chopsticks. Currently, only the `limitedReserAssetsTransfer` between KILT and HydraDx are tested.

Due to the XCM configuration change, the XCM pallet is benchmarked. Otherwise, the extrinsic would exceed the block limits.


Ad96el:
The network starts to boot now. While sending the reserved xcm message, the `DestinationUnsupported` error is returned. This happens while executing the `TransferReserveAsset` instruction in the xcm executor.

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "eslint-plugin-promise": "^6.0.0",
+    "prettier": "^3.2.5",
+    "ts-node": "^10.9.2",
+    "tsx": "^4.7.1",
+    "vitest": "^1.4.0"
+  },
+  "scripts": {
+    "lint": "eslint .",
+    "lint:fix": "eslint --fix .",
+    "clean": "rm -rf ./db && cargo build",
```
ntn-x2, line 38: Why do we `cargo build`? If this is necessary, I think it's enough if we build the required runtime, with `cargo build -p spiritnet-runtime`.

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "test": "LOG_LEVEL=error vitest --testTimeout=5000",
+    "dev": "tsx ./src/index.ts"
```
ntn-x2, line 40: I don't like the naming here. The script name could be something else, like `start-network`, or something like that. Second, the file is executed directly, hence the name `index.ts` is not really appropriate, as that's not a package, but a script. The file name should probably reflect the script name.

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "name": "e2e",
+  "version": "0.0.1",
+  "description": "chopsticks integration tests",
+  "main": "src/index.ts",
+  "type": "module",
+  "repository": "git@github.com:KILTprotocol/kilt-node.git",
+  "author": "[\"KILT <info@kilt.io>\"]",
+  "license": "MIT",
+  "dependencies": {
+    "@acala-network/chopsticks": "0.9.12",
```
ntn-x2, line 11: Aren't these all `devDependencies`?

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "e2e",
+  "version": "0.0.1",
+  "description": "chopsticks integration tests",
+  "main": "src/index.ts",
```
ntn-x2, line 5: We don't need this. Also, I would add a `"private"` property to make sure we don't publish this package.

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "e2e",
```
ntn-x2, line 2: This also deserves a better name, please 😃

integration-tests/chopsticks/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es2016",
+    "lib": ["es2019"],
+    "module": "NodeNext",
+    "rootDir": "src",
+    "outDir": "./bin",
```
ntn-x2, line 7: We don't need this, do we?

integration-tests/chopsticks/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import { connectParachains, connectVertical } from '@acala-network/chopsticks'
+
+import * as SpiritnetNetwork from './network/spiritnet.js'
+import * as PolkadotNetwork from './network/polkadot.js'
+import * as HydraDxNetwork from './network/hydroDx.js'
+import { keysCharlie, keysBob } from './helper.js'
+
+/// Helper function to validate the storage of the chains. The chains will not produce any blocks on demand.
+/// TODO: fix that
+async function spinUpNetwork() {
```
ntn-x2, line 10: So what's the goal of this function?

integration-tests/chopsticks/src/helper.ts
```diff
@@ -0,0 +1,8 @@
+import { Keyring } from '@polkadot/keyring'
+
+const keyring = new Keyring({ type: 'ed25519', ss58Format: 2 })
+keyring.setSS58Format(38)
```
ntn-x2, line 4: Why not using it directly as the parameter in the previous line?

integration-tests/chopsticks/src/tests/xcm.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+beforeAll(async () => {
+	xcmLogger.level = 'info'
+	spiritnetContext = await SpiritnetNetwork.getContext()
+	hydradxContext = await HydraDxNetwork.getContext()
+	polkadotContext = await PolkadotNetwork.getContext()
+
+	await polkadotContext.dev.setStorage(PolkadotNetwork.defaultStorage(keysCharlie.address))
+	await spiritnetContext.dev.setStorage(SpiritnetNetwork.defaultStorage(keysBob.address))
+	await hydradxContext.dev.setStorage(HydraDxNetwork.defaultStorage(keysBob.address))
```
ntn-x2, line 23: Did you mean `keysAlice` in any of these? It's not used anywhere.

integration-tests/chopsticks/src/tests/xcm.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	// Setup network
+	await connectVertical(polkadotContext.chain, spiritnetContext.chain)
+	await connectVertical(polkadotContext.chain, hydradxContext.chain)
+	await connectParachains([spiritnetContext.chain, hydradxContext.chain])
+
+	// fixes api runtime disconnect warning
+	await new Promise((r) => setTimeout(r, 500))
```
ntn-x2, line 31: Doesn't simply `await setTimeout(500)` work?

integration-tests/chopsticks/src/tests/xcm.test.ts
```diff
@@ -0,0 +1,72 @@
...
+	// Perform runtime upgrade
+	await Promise.all([polkadotContext.dev.newBlock(), spiritnetContext.dev.newBlock(), hydradxContext.dev.newBlock()])
```
ntn-x2, line 33: Why do we need to produce a block also for HydraDX? Is that required by the `connectParachains` and `connectVertical`?

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
...
+		Account: [[[addr], { providers: 1, data: { free: 1000 * 1e12 } }]],
+	},
+	polkadotXcm: {
+		safeXcmVersion: 3,
+	},
+})
+
+export const paraId = 2086
+export const hydraDxDestinationV2 = {
+	V2: {
```
ntn-x2, line 26: We can't use `V3`?

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
...
+		interior: {
+			X1: {
+				Parachain: HydraDxConfig.paraId,
+			},
+		},
+	},
+}
+export const hydraDxBeneficiaryV2 = {
+	V2: {
+		parents: 1,
```
ntn-x2, line 37: This does not match what we were told by HydraDX guys. `parents` should be `0`.

ntn-x2: And that makes sense since this is evaluated already on HydraDX, why would there be a parent in here?

integration-tests/chopsticks/src/tests/xcm.test.ts
```diff
@@ -0,0 +1,72 @@
...
+		)
+		.signAsync(keysBob)
+
+	const events = await sendTransaction(signedTx)
+
+	// fixes api runtime disconnect warning
+	await new Promise((r) => setTimeout(r, 50))
+
+	await spiritnetContext.chain.newBlock()
+	checkEvents(events, 'xcmpQueue').toMatchSnapshot('sender events xcm queue pallet')
```
ntn-x2, line 62: Where can we check that tokens were taken from the sender's account?

integration-tests/chopsticks/src/network/polkadot.ts
```diff
@@ -0,0 +1,26 @@
+import { setupContext, SetupOption } from '@acala-network/chopsticks-testing'
+import type { Config } from './types.js'
+
+export const options: SetupOption = {
+	endpoint: ['wss://rpc.polkadot.io', 'wss://polkadot-rpc.dwellir.com', 'wss://rpc.ibp.network/polkadot'],
+	db: './db/polkadot.db.sqlite',
+	port: 8000,
```
ntn-x2, line 7: Env variable.

integration-tests/chopsticks/src/network/polkadot.ts
```diff
@@ -0,0 +1,26 @@
+import { setupContext, SetupOption } from '@acala-network/chopsticks-testing'
+import type { Config } from './types.js'
+
+export const options: SetupOption = {
+	endpoint: ['wss://rpc.polkadot.io', 'wss://polkadot-rpc.dwellir.com', 'wss://rpc.ibp.network/polkadot'],
```
ntn-x2, line 5: This as well.

integration-tests/chopsticks/src/network/hydroDx.ts
```diff
@@ -0,0 +1,54 @@
+import { setupContext, SetupOption } from '@acala-network/chopsticks-testing'
+import type { Config } from './types.js'
+import { u8aToHex } from '@polkadot/util'
+import { decodeAddress } from '@polkadot/util-crypto'
+import * as SpiritnetConfig from './spiritnet.js'
+
+export const options: SetupOption = {
+	endpoint: ['wss://hydradx-rpc.dwellir.com', 'wss://rpc.hydradx.cloud'],
+	db: './db/hydradx.db.sqlite',
+	port: 8001,
```
ntn-x2, line 10: Same here.

integration-tests/chopsticks/src/network/hydroDx.ts
```diff
@@ -0,0 +1,54 @@
+import { setupContext, SetupOption } from '@acala-network/chopsticks-testing'
+import type { Config } from './types.js'
+import { u8aToHex } from '@polkadot/util'
+import { decodeAddress } from '@polkadot/util-crypto'
+import * as SpiritnetConfig from './spiritnet.js'
+
+export const options: SetupOption = {
+	endpoint: ['wss://hydradx-rpc.dwellir.com', 'wss://rpc.hydradx.cloud'],
```
ntn-x2, line 8: And here.

integration-tests/chopsticks/src/network/hydroDx.ts
```diff
@@ -0,0 +1,54 @@
...
+		acceptedCurrencies: [[[kiltTokenId], 100_000]],
+	},
+
+	System: {
+		Account: [[[addr], { providers: 1, data: { free: 100 * 10e12 } }]],
+	},
+})
+
+export const paraId = 2034
+export const sovereignAccount = u8aToHex(decodeAddress('5Eg2fntQqFi3EvFWAf71G66Ecjjah26bmFzoANAeHFgj9Lia'))
```
ntn-x2, line 50: On Spiritnet?

ntn-x2 (commented):
Left some remarks and some questions I'd like to get answered to know better what exactly is going on.

integration-tests/chopsticks/src/index.ts
Ad96el: It was quite handsome to see if updating the storage of the chains did work as expected. I could connect to polkadotjs and investigate the chain states.

integration-tests/chopsticks/src/tests/xcm.test.ts
Ad96el: `setTimeout` does not return a promise. Awaiting would be meaningless.

integration-tests/chopsticks/src/index.ts
ntn-x2: Yes, I get that. So you start the network before running the tests?

.nvmrc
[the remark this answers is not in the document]
Ad96el: moved in: [bdb5f2d](https://github.com/KILTprotocol/kilt-node/pull/614/commits/bdb5f2da6b37752e786a316fc133f5520648a1d9)

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
+import { setupContext, SetupOption } from '@acala-network/chopsticks-testing'
+import type { Config } from './types.js'
+import * as HydraDxConfig from './hydroDx.js'
```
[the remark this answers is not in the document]
Ad96el: ranamed in: [91a1d8d](https://github.com/KILTprotocol/kilt-node/pull/614/commits/91a1d8d236b2e93238e75acd6e0ca39bb9d92887)

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const options: SetupOption = {
+	endpoint: 'wss://kilt-rpc.dwellir.com',
```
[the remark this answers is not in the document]
Ad96el: I gave the possibility to overwrite the default values in: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
...
+	db: './db/spiritnet.db.sqlite',
+	port: 8002,
```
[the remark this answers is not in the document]
Ad96el: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,60 @@
...
+	wasmOverride: '../../target/debug/wbuild/spiritnet-runtime/spiritnet_runtime.wasm',
+	allowUnresolvedImports: true,
```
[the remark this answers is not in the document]
Ad96el: added comment in: [2a95581](https://github.com/KILTprotocol/kilt-node/pull/614/commits/2a9558114f546cffa8078fa49cc473a97e42cd79)

integration-tests/chopsticks/package.json
Ad96el: Good point: [48626b5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/48626b53ebcdb7e1a8c66170da8c678e479e55b3)

integration-tests/chopsticks/package.json
Ad96el: renamed in: [48626b5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/48626b53ebcdb7e1a8c66170da8c678e479e55b3)

integration-tests/chopsticks/package.json
Ad96el: True. Updated in: [48626b5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/48626b53ebcdb7e1a8c66170da8c678e479e55b3)

integration-tests/chopsticks/package.json
Ad96el: The private field is new for me :+1: [48626b5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/48626b53ebcdb7e1a8c66170da8c678e479e55b3)

integration-tests/chopsticks/package.json
Ad96el: I am terrible in naming :smile:. Is e2e-tests better? : [48626b5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/48626b53ebcdb7e1a8c66170da8c678e479e55b3)

integration-tests/chopsticks/tsconfig.json
Ad96el: Updated ts config in: [35f7268](https://github.com/KILTprotocol/kilt-node/pull/614/commits/35f7268a42ccdf11ce2809a8d2f3fccefc1bdb7b)

integration-tests/chopsticks/src/index.ts
Ad96el: No, this is not necessary. The problem with the tests was, that the network shuts down once the test cases are over. I could avoid this by using the `.pause` function of the chain network configs, but then I needed to increase the time out of the tests: which was quite annoying.

integration-tests/chopsticks/src/helper.ts
Ad96el: True: [58b6caa](https://github.com/KILTprotocol/kilt-node/pull/614/commits/58b6caa6a3cc3253a8573b878263b23f3c8c7c93)

integration-tests/chopsticks/src/tests/xcm.test.ts
Ad96el: This was intendant. I wanted to have one account without any balance.

integration-tests/chopsticks/src/tests/xcm.test.ts
Ad96el: Yes, a block needs to be produced for the updated XCM connections. I changed the comment in: [58b6caa](https://github.com/KILTprotocol/kilt-node/pull/614/commits/58b6caa6a3cc3253a8573b878263b23f3c8c7c93)

integration-tests/chopsticks/src/network/hydroDx.ts
Ad96el: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/hydroDx.ts
Ad96el: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/polkadot.ts
Ad96el: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/polkadot.ts
Ad96el: [46f5119](https://github.com/KILTprotocol/kilt-node/pull/614/commits/46f511924ffda667b943584517f5d87b9a5f3bcb)

integration-tests/chopsticks/src/network/spiritnet.ts
Ad96el: I added another test case for v3: https://github.com/KILTprotocol/kilt-node/pull/614/files#diff-ce4305e382897732899a9335dad6b6d4350a23b2d2be29fc5c4ca6f864806a48R11-R62

integration-tests/chopsticks/src/tests/xcm.test.ts
Ad96el: I added additional tests for the balance here: https://github.com/KILTprotocol/kilt-node/pull/614/commits/3f16f58266f0f363c4692fed1170b785f3c66b0a#diff-ce4305e382897732899a9335dad6b6d4350a23b2d2be29fc5c4ca6f864806a48R102-R107 

same for other test cases

integration-tests/chopsticks/src/network/spiritnet.ts
Ad96el: Yes, that was a bug. It is resolved, but I can not find the commit.

integration-tests/chopsticks/src/network/hydroDx.ts
Ad96el: Yes, that should be sovereign account of Spiritnet in hydradx. I removed this value this it is not needed. The funds are directly transferred to the omnipool account.

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReserveTransferSpiritnetHydraDx.test.ts.snap
```diff
@@ -0,0 +1,147 @@
...
+  },
+]
+`;
+
+exports[`Limited Reserve V2 Transfers from Spiritnet Account Bob -> HydraDx > sender events Balances 1`] = `
+[
+  {
+    "data": {
+      "amount": "(rounded 170000000000)",
+      "who": "4seWojfEHrk5YKPahdErazQ3CWEHZYi6NV4gKz5AaejWbRPJ",
```
Ad96el, line 39: This is the address of Alice

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReserveTransferSpiritnetHydraDx.test.ts.snap
```diff
@@ -0,0 +1,147 @@
...
+]
+`;
+
+exports[`Limited Reserve V3 Transfers from Spiritnet Account Bob -> HydraDx > receiver events currencies 1`] = `
+[
+  {
+    "data": {
+      "amount": 100000000000000000,
+      "currencyId": 60,
+      "who": "7L53bUTBbfuj14UpdCNPwmgzzHSsrsTWBHX5pys32mVWM3C1",
```
Ad96el, line 82: Omnipool account

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReseveTransferHydraDxSpiritnet.test.ts.snap
```diff
@@ -0,0 +1,139 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Limited Reserve Transfers from HydraDx Account Bob -> Spiritnet > receiver events Balances 1`] = `
+[
+  {
+    "data": {
+      "amount": 100000000000000000,
+      "who": "4qXPdpioJ6D8cgdeYXaukV2Y2oAQUHaX1VnGhdbSRqJn2CBt",
```
Ad96el, line 8: hydraDx sovereign account in spiritnet

integration-tests/chopsticks/src/network/spiritnet.ts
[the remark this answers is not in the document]
ntn-x2: Yeah I won't be too annoying here, but I think we should not have defaults at all. Will resolve this but make a comment in the refactoring ticket.

integration-tests/chopsticks/src/network/spiritnet.ts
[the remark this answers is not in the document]
ntn-x2: It explains what it does, but not why it is required. After reading the comment, I still don't understand if and when I need to use it.

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDx.test.ts
```diff
@@ -0,0 +1,120 @@
...
+	// Create some new blocks to have consistent snapshots
+	await new Promise((r) => setTimeout(r, 50))
+	await spiritnetContext.dev.newBlock()
+	await hydradxContext.dev.newBlock()
+
+	// check initial balance
+	const balanceSovereignAccountHydraDxBeforeTx = await getFreeBalanceSpiritnet(
+		SpiritnetConfig.hydraDxSovereignAccount
+	)
+	expect(balanceSovereignAccountHydraDxBeforeTx).eq(BigInt(0))
```
ntn-x2, line 27: Can you add the Omnipool balance test, to match the post-operation checks?

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDx.test.ts
```diff
@@ -0,0 +1,120 @@
...
+	// Create some new blocks to have consistent snapshots
+	await new Promise((r) => setTimeout(r, 50))
+	await spiritnetContext.dev.newBlock()
+	await hydradxContext.dev.newBlock()
+
+	// pre submit extrinsic checks
+	const balanceSovereignAccountHydraDxBeforeTx = await getFreeBalanceSpiritnet(
+		SpiritnetConfig.hydraDxSovereignAccount
+	)
+	expect(balanceSovereignAccountHydraDxBeforeTx).eq(BigInt(0))
```
ntn-x2, line 82: Same here.

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReserveTransferSpiritnetHydraDx.test.ts.snap
ntn-x2: Shouldn't it be Bob's?

integration-tests/chopsticks/src/tests/xcm/limitedReseveTransferHydraDxSpiritnet.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	// Give the sovereign account of HydraDx some kilt coins.
+	await spiritnetContext.dev.setStorage(
+		SpiritnetConfig.defaultStorage(u8aToHex(decodeAddress(SpiritnetConfig.hydraDxSovereignAccount)))
+	)
+	await hydradxContext.dev.setStorage(HydraDxConfig.defaultStorage(keysBob.address))
+
+	// check initial balance of alice
+	const aliceBalanceBeforeTx = await getFreeBalanceSpiritnet(keysAlice.address)
+	expect(aliceBalanceBeforeTx).eq(BigInt(0))
```
ntn-x2, line 27: Also here, can we test the same things we test after the operation is executed?

ntn-x2 (commented):
Already much better!

integration-tests/chopsticks/src/tests/xcm.test.ts
Ad96el: yes looks much cleaner: [4f63eef](https://github.com/KILTprotocol/kilt-node/pull/614/commits/4f63eef43f9dc5410aa0090c0370bf81ea4229f4)

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDx.test.ts
Ad96el: This is already checked. Please have a look on these lines: https://github.com/KILTprotocol/kilt-node/pull/614/files#diff-ce4305e382897732899a9335dad6b6d4350a23b2d2be29fc5c4ca6f864806a48R61-R62

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDx.test.ts
Ad96el: This is also checked. Please have a look on these lines: https://github.com/KILTprotocol/kilt-node/pull/614/files#diff-ce4305e382897732899a9335dad6b6d4350a23b2d2be29fc5c4ca6f864806a48R117-R118

integration-tests/chopsticks/src/tests/xcm/limitedReseveTransferHydraDxSpiritnet.test.ts
Ad96el: For Alice, this is a bit tricky.

1 KILT is moved from the sovereign account to the holding register. The fees for the transaction are paid with that, which finally results in Alice receiving slightly less than the transmitted funds.

Here is the error:

```
AssertionError: expected 99999999999971175n to equal 100000000000000000n

- Expected
+ Received

- 100000000000000000n
+ 99999999999971175n
```

So, do you think rounding is fine? The fees can be slightly different each time, right?

Currently this is checked by checking the Endowed event.

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReserveTransferSpiritnetHydraDx.test.ts.snap
Ad96el: No, please take a look here: https://github.com/KILTprotocol/kilt-node/pull/614/files#diff-ce4305e382897732899a9335dad6b6d4350a23b2d2be29fc5c4ca6f864806a48R31-R39

Alice is transmitting the extrinsic. The balance has to be withdrawn from her right?

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDx.test.ts
Ad96el: [ea2fa38](https://github.com/KILTprotocol/kilt-node/pull/614/commits/ea2fa38e25cb44e82b4a6ff5607b6b34032302b7)

integration-tests/chopsticks/src/tests/xcm/__snapshots__/limitedReserveTransferSpiritnetHydraDx.test.ts.snap
ntn-x2: The snapshot says "Limited Reserve V2 Transfers from Spiritnet Account Bob -> HydraDx > sender events Balances 1".

Ad96el: True sorry: [71a810b](https://github.com/KILTprotocol/kilt-node/pull/614/commits/71a810bfdfdbdaf7b1a2022f283f59036e790a16)

integration-tests/chopsticks/src/tests/xcm/limitedReseveTransferHydraDxSpiritnet.test.ts
ntn-x2: I don't think the fees can be different, so it's fine to compare it to the value here and explaining why that's the case, with a comment.

integration-tests/chopsticks/src/network/hydraDx.ts
```diff
@@ -0,0 +1,71 @@
...
+		],
+	},
+	multiTransactionPayment: {
+		acceptedCurrencies: [[[kiltTokenId], 100_000]],
+	},
+
+	System: {
+		Account: [[[addr], { providers: 1, data: { free: initialBalanceHDX } }]],
+	},
+})
```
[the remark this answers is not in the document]
ntn-x2: We have a ticket for it, will close this conversation.

integration-tests/chopsticks/src/tests/xcm/limitedReseveTransferHydraDxSpiritnet.test.ts
Ad96el: [95256c5](https://github.com/KILTprotocol/kilt-node/pull/614/commits/95256c55d9d16d4051d54939d027a7bdecf2120f)

integration-tests/chopsticks/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	await polkadotContext.dev.setStorage(PolkadotNetwork.defaultStorage(keysCharlie.address))
+	await spiritnetContext.dev.setStorage(SpiritnetNetwork.defaultStorage(keysBob.address))
+	await hydradxContext.dev.setStorage(HydraDxNetwork.defaultStorage(keysBob.address))
+
+	// Setup network
+	await connectVertical(polkadotContext.chain, spiritnetContext.chain)
+	await connectVertical(polkadotContext.chain, hydradxContext.chain)
+	await connectParachains([spiritnetContext.chain, hydradxContext.chain])
+}
```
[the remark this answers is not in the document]
ntn-x2: Should we make it as part of a different PR?

integration-tests/chopsticks/src/network/spiritnet.ts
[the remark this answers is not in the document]
ntn-x2: One more comment for this pls?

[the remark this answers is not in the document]
Ad96el: The flag was needed, otherwise I receive an error, if I try to build the runtime from the WASM. 
[e2595f0](https://github.com/KILTprotocol/kilt-node/pull/614/commits/e2595f0b85e2da8a7a281cc3d7cd7c13aed07c28)

integration-tests/chopsticks/src/index.ts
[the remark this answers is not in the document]
Ad96el: Yeah, let's tackle that in a different PR. It is also not really urgent. It is removed in: [6031e34](https://github.com/KILTprotocol/kilt-node/pull/614/commits/6031e34446827c112bec650cc5953a1504d17937)

integration-tests/chopsticks/src/index.ts
Ad96el: No, respinning the network overwrites the database. It was just used to observe how the storage appears after using `.dev.setStorage`.

integration-tests/chopsticks/src/index.ts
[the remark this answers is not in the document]
ntn-x2: Nice. I still think it's very good if we could spin up a network that uses the same config we use for tests, without having to duplicate config files as I did in https://github.com/KILTprotocol/kilt-node/pull/619.

ntn-x2 (approved):
Please add a CI step so that all of this is actually used 😃

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
...
+integration-tests:
+  timeout: 5 minutes
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: test
+  script:
+    - curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+    - apt-get install -y nodejs
+    - cargo build -p spiritnet-runtime
+    - cd ./integration-tests/chopsticks
+    - yarn
```
ntn-x2, line 42:
```suggestion
    - yarn --immutable
```

integration-tests/chopsticks/src/network/hydraDx.ts
```diff
@@ -0,0 +1,69 @@
...
+export const kiltTokenId = 60
+
+export function setGovernance(addr: string[]) {
+	return {
+		TechnicalCommittee: { Members: addr },
+		Council: { Members: addr },
+	}
+}
+
+export function assignNativeTokensToAccount(addr: string[], balance: bigint = initialBalanceHDX) {
```
ntn-x2, line 21: `
```suggestion
export function assignNativeTokensToAccounts(addr: string[], balance: bigint = initialBalanceHDX) {
```

integration-tests/chopsticks/src/network/hydraDx.ts
```diff
@@ -0,0 +1,69 @@
...
+	return {
+		System: {
+			Account: addr.map((address) => [[address], { providers: 1, data: { free: balance } }]),
+		},
+	}
+}
+
+export function assignKiltTokensToAccount(addr: string[], balance: bigint = initialBalanceKILT) {
```
ntn-x2, line 29:
```suggestion
export function assignKiltTokensToAccounts(addr: string[], balance: bigint = initialBalanceKILT) {
```

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -0,0 +1,50 @@
...
+	endpoint: process.env.SPIRITNET_WS || 'wss://kilt-rpc.dwellir.com',
+	db: './db/spiritnet.db.sqlite',
+	port: toNumber(process.env.SPIRITNET_PORT) || 9002,
+	wasmOverride: '../../target/debug/wbuild/spiritnet-runtime/spiritnet_runtime.wasm',
+	// Whether to allow WASM unresolved imports when using a WASM to build the parachain. This Flag is needed otherwise, the runtime can not be built from the WASM. Chopsticks throws an error when it encounters an unresolved import.
+	allowUnresolvedImports: true,
+}
+
+/// Assigns the native tokens to an account
+export function assignNativeTokensToAccount(addr: string[], balance: bigint = initialBalanceKILT) {
```
ntn-x2, line 16:
```suggestion
export function assignNativeTokensToAccounts(addr: string[], balance: bigint = initialBalanceKILT) {
```

integration-tests/chopsticks/src/network/hydraDx.ts
ntn-x2: Can you add comments for these functions as well?

integration-tests/chopsticks/src/network/polkadot.ts
```diff
@@ -0,0 +1,38 @@
...
+	endpoint: process.env.POLKADOT_WS || [
+		'wss://rpc.polkadot.io',
+		'wss://polkadot-rpc.dwellir.com',
+		'wss://rpc.ibp.network/polkadot',
+	],
+	db: './db/polkadot.db.sqlite',
+	port: toNumber(process.env.POLKADOT_PORT) || 9000,
+}
+
+export function setAddrNativeTokens(addr: string[], balance: bigint = initialBalanceDOT) {
```
ntn-x2, line 15: Can you name it the same as elsewhere?

integration-tests/chopsticks/src/network/utils.ts
```diff
@@ -0,0 +1,53 @@
+export function getSiblingDestination(paraId: number) {
```
ntn-x2, line 1: Nitpick, but I would call this `getSiblingLocation`, or `getSiblingMultiLocation`.

ntn-x2: Same for the others.

integration-tests/chopsticks/src/network/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function getAccountDestinationV3(addr: string) {
+	return {
+		V3: {
+			parents: 0,
+			interior: {
+				X1: {
+					AccountId32: {
+						id: addr,
```
ntn-x2, line 40: You forgot the `network` here?

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import { getAccountDestinationV2, getNativeAssetIdLocation, getSiblingDestination } from '../../network/utils.js'
+import { checkBalance, createBlock, hexAddress, setStorage } from '../utils.js'
+
+const KILT_ASSET_V2 = { V2: [getNativeAssetIdLocation(KILT)] }
+
+test('Limited Reserve V2 Transfers from Spiritnet Account Alice -> HydraDx', async ({ expect }) => {
+	const { checkEvents, checkSystemEvents } = withExpect(expect)
+
+	// Set storage
+	await setStorage(spiritnetContext, SpiritnetConfig.assignNativeTokensToAccount([keysAlice.address]))
```
ntn-x2, line 17: Should these be part of a `beforeEach` or `beforeAll` hook?

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
```diff
@@ -0,0 +1,52 @@
...
+		.signAsync(keysAlice)
+
+	const events = await sendTransaction(signedTx)
+
+	// Order matters here, we need to create a block on the sender first
+	await createBlock(spiritnetContext)
+	await createBlock(hydradxContext)
+
+	// Check events sender
+	checkEvents(events, 'xcmpQueue').toMatchSnapshot('sender events xcm queue pallet')
```
ntn-x2, line 39: I think it would make more sense to check these conditions right after we produce a block on the sender. Then we produce a block on the receiver and check for the receiver conditions.

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
```diff
@@ -0,0 +1,52 @@
...
+	checkEvents(events, { section: 'balances', method: 'Withdraw' }).toMatchSnapshot('sender events Balances')
+
+	// Check events receiver
+	checkSystemEvents(hydradxContext, { section: 'currencies', method: 'Deposited' }).toMatchSnapshot(
+		'receiver events currencies'
+	)
+	checkSystemEvents(hydradxContext, 'xcmpQueue').toMatchSnapshot('receiver events xcmpQueue')
+
+	// check balance
+	await checkBalance(getFreeBalanceSpiritnet, SpiritnetConfig.hydraDxSovereignAccount, expect, KILT)
```
ntn-x2, line 50: Also the balance checks I would all group them together around the sender action. So instead of diving the tests by what we are testing, diving them by sender and receiver, since that's what we likely want to check. What do you think?

integration-tests/chopsticks/src/tests/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+	// Setup network
+	await connectVertical(polkadotContext.chain, spiritnetContext.chain)
+	await connectVertical(polkadotContext.chain, hydradxContext.chain)
+	await connectParachains([spiritnetContext.chain, hydradxContext.chain])
+
+	const newBlockConfig = { count: 2 }
+	// fixes api runtime disconnect warning
+	await setTimeout(50)
+	// Perform runtime upgrade and establish xcm connections.
```
ntn-x2, line 28: This will need to go after we go live, right? Same for some of the HydraDX storage overrides. Do you think we should point the tests to a specific block on all the three chains, to reduce even more externalities?

integration-tests/chopsticks/src/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export function toNumber(value: string | undefined): number | undefined {
+	if (value === undefined) {
+		return undefined
+	}
+
+	return Number(value)
+}
+
+export const KILT = BigInt(10 * 10e15)
+export const DOT = BigInt(10 * 10e10)
```
ntn-x2, line 18: Are you sure these are correct? This to me sounds like 100 DOTs, not one DOT. Did you want to refer to a single DOT?

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
...
+    - yarn 
+    - yarn test
+
+lint-typescript:
+  timeout: 5 minutes
+  image: node:20
+  stage: test
+  script:
+    - cd ./integration-tests/chopsticks
+    - yarn
```
ntn-x2, line 51:
```suggestion
    - yarn --immutable
```

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
...
+  stage: test
+  script:
+    - curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+    - apt-get install -y nodejs
+    - cargo build -p spiritnet-runtime
+    - cd ./integration-tests/chopsticks
+    - yarn 
+    - yarn test
+
+lint-typescript:
```
ntn-x2, line 45: I don't think we need a whole separate job to check for lints. We can add it to the previous step.

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
...
+  timeout: 5 minutes
+  image: node:20
+  stage: test
+  script:
+    - cd ./integration-tests/chopsticks
+    - yarn 
+    - yarn lint
```
ntn-x2, line 52: What about prettier? We don't check for that? Or is that also checked as part of `yarn lint`?

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
   script:
     - cargo test --all --all-features --all-targets --locked
 
+integration-tests:
+  timeout: 5 minutes
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: test
+  script:
+    - curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
```
ntn-x2, line 38: We have a `.nvmrc` file, can't we simply install [nvm](https://github.com/nvm-sh/nvm) and run `nvm install` from the `chopsticks` dir? Then we don't need to install a version of Node which could get out-of-sync with our `.nvmrc` file.

.gitlab-ci.yml
```diff
@@ -30,6 +30,27 @@ test-features:
   script:
     - cargo test --all --all-features --all-targets --locked
 
+integration-tests:
```
ntn-x2, line 33: I think we should aim to change the workdir to be directly `chopsticks`, and do everything from inside there, including building the runtime, which could be another yarn script.

ntn-x2: Since this could potentially also be its own repo, so we should treat is as completely isolated from the rest of the repo.

integration-tests/chopsticks/src/network/utils.ts
Ad96el: In XCM version 3, network is optional. Any reason to provide it here?

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
Ad96el: I am not sure. My intention was to keep the initial state as minimalistic as possible to ensure it's as generic as possible in every test case. So this was an active design decision. Changing that would also mean, to update the test case for sending funds from hydra -> spiritnet, since I assume there, Alice has no funds.

integration-tests/chopsticks/src/network/utils.ts
ntn-x2: Ok then it's fine, I was not aware of that.

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
ntn-x2: Yeah makes sense.

integration-tests/chopsticks/src/network/hydraDx.ts
Ad96el: updated in: [41c2859](https://github.com/KILTprotocol/kilt-node/pull/614/commits/41c2859fbd6ce083c9dc7fc8ea53a50bc36fb37a)

integration-tests/chopsticks/src/network/hydraDx.ts
Ad96el: Updated in: [bec955e](https://github.com/KILTprotocol/kilt-node/pull/614/commits/bec955eb1a2f911de2ef4479826913cacd5e5638)

integration-tests/chopsticks/src/network/spiritnet.ts
Ad96el: updated in: [bec955e](https://github.com/KILTprotocol/kilt-node/pull/614/commits/bec955eb1a2f911de2ef4479826913cacd5e5638)

integration-tests/chopsticks/src/network/polkadot.ts
Ad96el: updated in: [bec955e](https://github.com/KILTprotocol/kilt-node/pull/614/commits/bec955eb1a2f911de2ef4479826913cacd5e5638)

integration-tests/chopsticks/src/network/utils.ts
Ad96el: update in: [41c2859](https://github.com/KILTprotocol/kilt-node/pull/614/commits/41c2859fbd6ce083c9dc7fc8ea53a50bc36fb37a)

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
Ad96el: I do not have a strong opinion on that. If you think this makes stuff clearer and more intuitive: [e6e5f44](https://github.com/KILTprotocol/kilt-node/pull/614/commits/e6e5f4478ddc5254d35b846f22c393f2f22b56d9)

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransferSpiritnetHydraDxV2.test.ts
Ad96el: I do not have a strong opinion on that. If you think, it makes the flow clearer, I am in: [e6e5f44](https://github.com/KILTprotocol/kilt-node/pull/614/commits/e6e5f4478ddc5254d35b846f22c393f2f22b56d9)

integration-tests/chopsticks/src/tests/index.ts
Ad96el: That's a good point. I am not sure. I think having the latest chain state has more value than having consistent tests. So I would propose to update these tests, once the referendum from HydraDx is through. We would also notice here, if they have changed something on their side. What do you think?

integration-tests/chopsticks/src/utils.ts
Ad96el: Good catch: [0ebcde8](https://github.com/KILTprotocol/kilt-node/pull/614/commits/0ebcde8b490b1a24a07d4825f09acf7035cdf810)

.gitlab-ci.yml
Ad96el: I am not too much in favor to move the directory into the workdir. First, I do not like, if the root directory is super full. Second, the “project” aims to provide some tests. I like the strict separation of logic and testing.

integration-tests/chopsticks/src/tests/index.ts
ntn-x2: Yeah but I feel this approach is not maintainable. What if we have tests for 10 different chains, 6 of which must be updated after our runtime upgrade, and 2 after they have upgraded their runtime? I am not sure there is an easy solution here. At the very least, I would probably separate e2e tests by runtime couples, something like (spiritnet-1.12.1 <-> hydra-1) another for (spiritnet-1.13.0 <-> hydra-1), and another one for (spiritnet-1.13.0 <-> hydra-2). Then we know exactly when each test becomes deprecated and can be removed. Let's make a ticket for investigating how to best solve this issue, and maybe paste the suggestion I just gave. We can have a proper discussion once we will tackle the issue properly.

.gitlab-ci.yml
Ad96el: I am not in favor of mixing these two tests into one. Checking for quality, such as formatting and linting, should, in my opinion, be a single step. The second step should only focus on the logic itself, I would say.

ntn-x2: Yeah but these are integration tests, we are not publishing a library. I think it's overkill to take CI time to test for something that it's "nice to have" but not strictly required. What would be the downside of having the lints checked before the tests are run? If the tests pass but there's a lint issue, the next push will trigger the whole thing again, and we have just "wasted" time to re-run tests that already passed.

.gitlab-ci.yml
Ad96el: It should be tested in yarn lint. I would verify it again.

.gitlab-ci.yml
Ad96el: Thats a good point: [63a0d0c](https://github.com/KILTprotocol/kilt-node/pull/614/commits/63a0d0cd7ca9f852e8802b58234b9877b6bf164e)

.gitlab-ci.yml
Ad96el: Yeah okay. The tests is actually super fast (takes less then 1 min). But I added the check a step ahead: [63a0d0c](https://github.com/KILTprotocol/kilt-node/pull/614/commits/63a0d0cd7ca9f852e8802b58234b9877b6bf164e)

.gitlab-ci.yml
ntn-x2: That is not what I am suggesting. I was suggesting setting the `WORKDIR` to be `integration-tests/chopsticks` instead of `cd`ing into it. But I could not find a keyword that GitLab would support, so I guess `cd`ing into it is fine. I was not suggesting to copy or move anything 😄

Ad96el: Ah okay sorry. Yeah I could also not find anything. That is why I `cd` into the directory.

.gitlab-ci.yml
Ad96el: [6e25909](https://github.com/KILTprotocol/kilt-node/pull/614/commits/6e25909d5b148bbe1dbb6f4031b69ef73e078f8f)

.gitlab-ci.yml
Ad96el: [6e25909](https://github.com/KILTprotocol/kilt-node/pull/614/commits/6e25909d5b148bbe1dbb6f4031b69ef73e078f8f)

integration-tests/chopsticks/src/tests/index.ts
ntn-x2: > Currently, I am afraid that we might spend time solving a problem that may not arise in the next 1, 2, or even 4 years.
I don't actually think that's true. If anything, now that we have a "more open" XCM configuration, we might get more inbound requests to support other use cases, and we definitely need tests for that. Anyway, as long as we have a ticket, this conversation can be resolved. But please, make another ticket to remove any stuff that is not relevant after our first runtime upgrade goes through, after HDX first proposal goes through, our second proposal goes through, and HDX second proposal goes through 😁

.gitlab-ci.yml
Ad96el: It wasn't checked. Seems like something in the config was wrong. I update the lint command to also check the fmt: [5905280](https://github.com/KILTprotocol/kilt-node/pull/614/commits/59052803d6f9d62b13d657d32f9f3b84c99dd271)

integration-tests/chopsticks/src/tests/index.ts
Ad96el: I hope I am wrong. This would be great for KILT :100:. And I totally agree, we should test each interaction with another chain. I created a ticket to update the storage mocks, once the HydraDx referendum is through: [#3344](https://github.com/KILTprotocol/ticket/issues/3344)

ntn-x2 (dismissed):

.gitlab-ci.yml
```diff
@@ -10,7 +10,7 @@ clippy-and-docs:
     - cargo clippy --all-features --all-targets --locked -- -D warnings
     - RUSTDOCFLAGS='-D warnings' cargo doc --all-features --no-deps --locked
 
-fmt:
+fmt-rust:
```
ntn-x2, line 13: Why this?

integration-tests/chopsticks/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "typescript": "*",
+    "vitest": "^1.4.0",
+    "eslint-plugin-jsx-a11y": "^6.8.0"
+  },
+  "scripts": {
+    "lint": "eslint src && prettier --check src",
+    "lint:fix": "eslint --fix src && prettier --write src",
+    "clean": "rm -rf ./db && cargo build -p spiritnet-runtime",
+    "test": "LOG_LEVEL=error vitest",
+    "test:CI": "vitest --bail 0  --no-file-parallelism"
```
ntn-x2, line 37: Does `--bail 0` mean the CI will continue regardless of any errors?

ntn-x2: Why is `--no-file-parallelism` required?

Ad96el: If `--no-file-parallelism` isn't specified, vitest executes all three test files concurrently. This includes spawning three services for each test case. In the logs, I noticed that some services aren't responding, likely due to insufficient computational resources. Enabling this flag ensures that tests run sequentially, preventing individual service failures during initialization.

I retried the tests 3 times while other tests were running in the background. All three succeeded.

ntn-x2: I think to me this calls for suboptimal test setup then. Did you figure out if it's something we do or something Chopsticks does? I did not find the same flag on their e2e tests, so I suspect it could be some concurrency issue. Anyway, not blocking this merge, but it can perhaps be tackled as part of https://github.com/KILTprotocol/ticket/issues/3345.

ntn-x2 (approved):
👍👍👍

PR #623, opened 2024-04-10
fix: try-runtime support in CI

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3300.

Currently srtool does not really support features (https://github.com/chevdor/srtool-cli/issues/32), so we need a workaround until features are supported in srtool. Since this is what we do anyway now, I decided to go with using the same build infrastructure we use for tests.


ggera (approved):
LGTM

PR #624, opened 2024-04-10
test: add test-try-runtime step to CI

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3036.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,6 +30,26 @@ test-features:
   script:
     - cargo test --all --all-features --all-targets --locked
 
+# TODO: The try-runtime-cli executable could be built as part of the Docker image directly, saving some time.
+test-try-runtime:
+  parallel:
+    matrix:
+      - RUNTIME: "peregrine"
+        ENDPOINT: "wss://peregrine.kilt.io:443"
+      - RUNTIME: "spiritnet"
+        ENDPOINT: "wss://spiritnet.kilt.io:443"
+  timeout: 2 hours
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: test
+  variables:
+    TRY_RUNTIME_CLI_VERSION_TAG: v0.6.1
+  script:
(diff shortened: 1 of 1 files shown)


ntn-x2:
@Ad96el @ggera we caught a nice bug because of this PR. I will rebased on top of `develop` once https://github.com/KILTprotocol/kilt-node/pull/625 is merged.

.gitlab-ci.yml
```diff
@@ -30,6 +30,29 @@ test-features:
...
+  timeout: 2 hours
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: test
+  variables:
+    TRY_RUNTIME_CLI_VERSION_TAG: v0.6.1
+  script:
+    - curl -sL https://github.com/paritytech/try-runtime-cli/releases/download/${TRY_RUNTIME_CLI_VERSION_TAG}/try-runtime-x86_64-unknown-linux-musl -o try-runtime
+    - chmod +x ./try-runtime
+    - ./try-runtime --version
+    - cargo build --locked --release -p ${RUNTIME}-runtime --features try-runtime
```
ggera, line 50: isn't this line double work of  `build-wasm-try-runtime` job? Can this be optimized by either falgging it  ` stage: test` then using artifacts or updating the existing job to do the checcccks as well?

ntn-x2: To be honest, we don't need to build the `try-runtime` WASM every time we push a commit to a PR, but rather only when the PR is merged, so we can't put that and running tests under the same step. Yeah I acknowledge it's "redundant", but I would like to overhaul the whole CI stuff at some point. For now I think it's important to run these tests often, as we've seen that some testing logic was incorrect.

ntn-x2: Also, in this PR, `build-wasm-try-runtime` uses the `srtool` still, which does not support specifying features, so that step's broken. I could fix it once we merge this PR and https://github.com/KILTprotocol/kilt-node/pull/623, perhaps.

ggera (dismissed):

ggera (approved):
LGTM

PR #630, opened 2024-04-23
feat: add logging to DIP components

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3128.

I introduced different new logging targets. For the deposit storage and relay store pallet, which are not technically part of the DIP protocol, the target is a standalone one. For the other things, meaning DIP pallets and DIP provider/consumer components, the log targets have the following structure: `"dip::<provider>|<consumer>::<type_name>"`. So filtering provider logs would require `RUST_LOG=dip::provider=info`, for instance. Most of the logs are `info` level, with some `error` where we return `Internal` errors or errors coalesced into a `u8` value, and a couple of `trace` for the less important stuff to show.


crates/kilt-dip-primitives/src/merkle_proofs/v0/provider_state/mod.rs
```diff
@@ -176,6 +176,7 @@ impl<
 		ProviderHeader: Decode + HeaderT<Hash = OutputOf<RelayHasher>, Number = KiltBlockNumber>,
 	{
 		let provider_head_storage_key = calculate_parachain_head_storage_key(provider_para_id);
+		log::trace!(target: "dip::consumer::ParachainDipDidProofV0", "Calculated storage key for para ID {:#?} = {:#?}", provider_para_id, provider_head_storage_key);
```
Ad96el, line 179: Should the log be consistent? The first two started with Failed. Should this maybe also start with: Failed to calculate ... ?

crates/kilt-dip-primitives/src/merkle_proofs/v0/provider_state/mod.rs
```diff
@@ -466,10 +468,11 @@ impl<
 			self.dip_proof.blinded.as_slice(),
 			proof_leaves_key_value_pairs.as_slice(),
 		)
+		// Can't log since the result returned by `verify_trie_proof` implements `Debug` only with `std`.
 		.map_err(|_| Error::InvalidDidMerkleProof)?;
 
 		let revealed_leaves = BoundedVec::try_from(self.dip_proof.revealed).map_err(|_| {
-			log::error!("Should not fail to construct BoundedVec since bounds were checked before.");
+			log::error!(target: "dip::consumer::DipDidProofWithVerifiedSubjectCommitmentV0", "Failed to construct BoundedVec<u8, {MAX_REVEALED_LEAVES_COUNT}>.");
```
Ad96el, line 475: Do you think it could help to log, the actual size of the `dp_proof.revealed` and the max size?

Ad96el, line 475: Do you think it could help to log, the actual size of the dip_proof.revealed length and the max size? Are there more possibilities, how this can fail?

crates/kilt-dip-primitives/src/merkle_proofs/v0/provider_state/mod.rs
Ad96el, line 179: Would be nice if all logs have a consistent fmt. Some start with Failed to ..., Some have the failed in the middle.

Ad96el (dismissed):
Nice changes. 

Two picky comments. Would be charming if some logs are more consistent. Some start with Failed and some others have the failed in the middle of the sentence, but nothing which should block this PR to be merged. 

I assume the integration-tests are currently failing because we are in an intermediate state. I tried to fix them and I get the error Transport("NoChannel"). 

These errors are probably resolved, once we accept the channel request from HydraDx. Nevertheless, some changes are required. The KILT token registration is no longer necessary. I have fixed them already and will create a PR, once the referendum is through.

crates/kilt-dip-primitives/src/merkle_proofs/v0/provider_state/mod.rs
ntn-x2: I mean, for those we should hopefully have access to the tx payload itself, which reveals what the input was. I tried to log only intermediate computations that are not directly visible by looking at a transaction input or output, mostly.

ntn-x2: In this case, `revealed` is provided by the user in the proof, so we know beforehand what its length is.

crates/kilt-dip-primitives/src/merkle_proofs/v0/provider_state/mod.rs
ntn-x2: This is not an error message, is a trace about the calculated storage key. I made the error messages consistent in https://github.com/KILTprotocol/kilt-node/pull/630/commits/f99015d526e888041d255c3fb61b109a55af3bc4. This requires a re-approval from you.

Ad96el (dismissed):

PR #639, opened 2024-05-16
ci: cancel old piplines

Ad96el:
## fixes KILTProtocol/ticket#3406

This PR introduces automatic cancellation of pipelines when a new commit is pushed to the branch.

Furthermore, if any test fails, the entire pipeline will be canceled. If this is not wished, I can revert the changes.

Lastly, if a commit has the title "ci: Updates Snapshots", the tests will be skipped.

Example branch, where the pipeline is skipped because of the commit message: https://gitlab.com/kiltprotocol/kilt-node/-/pipelines/1293783978


.gitlab-ci.yml
```diff
@@ -1,8 +1,22 @@
+before_script:
+  - >
+    if echo "$CI_COMMIT_TITLE" | grep -q "ci: Update snapshots"; then
```
ntn-x2, line 3: Actually I think we can do something better: can we define a generic prefix, for instance `ci-skip-rust:` and `ci-skip-e2e:` which can be followed by a generic string indicating the reason, that would let us skip either the Rust or the e2e tests? So we can push different commit "semantics" that would all have the same effect, i.e., skipping some tests.

.gitlab-ci.yml
```diff
@@ -1,8 +1,22 @@
...
+      echo "Skipping tests due to commit title"
+      exit 0
+    fi
+
+workflow:
+  auto_cancel:
+    on_new_commit: interruptible
+    on_job_failure: all
```
ntn-x2, line 11: I'm not sure on this one. Technically it would be good that we catch more than one error, so that we can push a new commit that fixes them all, rather than finding out one by one.

.gitlab-ci.yml
ntn-x2: Anyway if it's not easily doable, don't spend too much time on this. We will overhaul the CI structure anyway at some point.

.gitlab-ci.yml
Ad96el: Yeah, I was also unsure about this flag. The reason I decided to introduce it is that normally the tests should be executed locally. The CI step should only validate the final state of a PR.

I don't have a strong opinion on this. If you prefer, I can revert it.

ntn-x2: I think in case the PR changes hands for some reasons, it's good to have a full picture of what the current state is, rather than relying on local-only context.

.gitlab-ci.yml
ntn-x2: If this does what's intended, I think it's good 👍 Question: where can we document this, for us and for others?

.gitlab-ci.yml
ntn-x2: Also, seeing how the integration tests are failing often, letting all other tests running could be good to then say "let's merge anyway since everything else works"

.gitlab-ci.yml
Ad96el: It should work. I have to example commits where you can investigate the output: 

for rust: [ba85c34](https://github.com/KILTprotocol/kilt-node/pull/639/commits/ba85c34538b4f401584bebd772e4593e8980eb6d) pipeline: https://gitlab.com/kiltprotocol/kilt-node/-/pipelines/1295088348 

for integration tests: [f5449e7](https://github.com/KILTprotocol/kilt-node/pull/639/commits/f5449e7ea432c3cb0a5791e3c55dbb117a73d54d) pipeline: https://gitlab.com/kiltprotocol/kilt-node/-/pipelines/1295116631

Ad96el: I'm unsure where to place the documentation. I'm considering introducing a `WORKFLOW_README.md` in the root folder. We could also add an explanation to the [command-bot](https://github.com/KILTprotocol/command-bot), but I'm not certain if that's the right place.

ntn-x2: Yeah it does not look like we can add it to the README. Maybe we can use a "Contributing" file like the one the polkadot-sdk repo has: https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/CONTRIBUTING.md.

Ad96el: For me, it feels like a new issue. I created a new ticket to track this: [#3416](https://github.com/KILTprotocol/ticket/issues/3416). We should also maybe discuss which aspects should be described.

ntn-x2 (approved):

PR #645, opened 2024-05-20
chore: sort TOML files

ntn-x2:
Built on top of https://github.com/KILTprotocol/kilt-node/pull/644.

I applied the following:

1. Sort keys that were not sorted in the top-level `package` field for all crates
2. Replace all simple occurrences of a crate version (crate = "x.y.z) or workspace inheritance (package.workspace = true) with their object-based counterparts (crate = { version = "x.y.z" } and package = { workspace = true }) to keep our style consistent
3. Made sure the dependencies are listed as build dependencies first, then dev dependencies, then regular dependencies for all crates
4. Add the taplo formatter (thanks @Ad96el)

I could not find any plug-ins or tools that would allow us to force any rules on the TOML files, so we'll have to try and stick with the current way, and hopefully it's now easier once all crates are consistent and can be copy-pasted to be used as templates for new ones.


Ad96el:
Did you use the `even better toml` VS Code extension for the formatting?

.gitlab-ci.yml
```diff
@@ -16,6 +16,13 @@ fmt:
   script:
     - cargo fmt -- --check
 
+fmt-tomls:
+  image: paritytech/ci-unified:bullseye-1.70.0
+  stage: test
+  script:
+    - cargo install taplo-cli --version 0.9.0
+    - taplo fmt --check 
+
```
Ad96el, line 25: If this is not wished, I can revert it

Ad96el (approved):

PR #650, opened 2024-05-31
Ag chopsticks design

Ad96el:

<details>

```
```

</details>

<details>

```
```

</details>

## Checklist:

    * Link the PR/Ticket here


Ad96el:
It might happen that that still errors are thrown by some concurrence issues regarding SQLite. We could either ignore those errors, since they are not part of the testing logic, or run the test not in parallel, which would result into a slower CI (currently 10 to maybe 15 mins).  

Resolving those issues seems challenging since they come directly from chopsticks. 

Ignoring the errors could result in some unforeseen side effects which could give use false positive tests.

.gitlab-ci.yml
```diff
@@ -68,16 +68,20 @@ integration-tests:
   image: paritytech/ci-unified:bullseye-1.70.0
   stage: test
   variables:
-    CI: "true"
+    SPIRITNET_BLOCK_NUMBER: 6390611
```
ntn-x2, line 71: Please move these to the relevant part of our new GH workflow.

.gitlab-ci.yml
```diff
@@ -68,16 +68,20 @@ integration-tests:
...
+    HYDRADX_BLOCK_NUMBER: 5235787
+    POLKADOT_BLOCK_NUMBER: 21010819 
+    SPIRITNET_WASM_OVERRIDE: "../../target/debug/wbuild/spiritnet-runtime/spiritnet_runtime.compact.compressed.wasm"
```
ntn-x2, line 74: No compact WASM is generated anymore, so this should refer to the regular, uncompressed, uncompacted one.

integration-tests/chopsticks/package.json
```diff
@@ -19,22 +19,21 @@
...
+    "vitest": "^1.4.0"
   },
   "scripts": {
     "ts-check": "tsc --noEmit",
     "lint": "eslint src && prettier --check src",
     "lint:fix": "eslint --fix src && prettier --write src",
     "clean": "rm -rf ./db",
-    "test": "LOG_LEVEL=error vitest",
-    "test:CI": "vitest --bail 0  --no-file-parallelism"
+    "test": "LOG_LEVEL=error vitest"
```
ntn-x2, line 37: Definitely lower priority, but we should investigate how to make the tests run faster without them throw errors. Our new GH pipeline currently takes almost 1h, but I have no idea if this refactoring will make tests go faster or slower.

integration-tests/chopsticks/README.md
```diff
@@ -0,0 +1,74 @@
+# e2e Chopsticks tests
+
+This project is a set of end-to-end tests for the KILT protocol against other parachains. 
+Other functionalities such as a creation of DID can be easy inserted.
```
ntn-x2, line 4:
```suggestion
Other functionalities such as a creation of DID can be easily added.
```

integration-tests/chopsticks/.env-example
```diff
@@ -1,6 +1,10 @@
-HYDRADX_WS=
-HYDRADX_PORT=
-POLKADOT_WS=
-POLKADOT_PORT=
-SPIRITNET_WS=
-SPIRITNET_PORT=
+# Those env variables are mandatory for the tests to work.
```
ntn-x2, line 1:
```suggestion
# These env variables are mandatory for the tests to work.
```

integration-tests/chopsticks/src/tests/xcm/types.ts
```diff
@@ -0,0 +1,48 @@
...
+	receiverAccount: KeyringPair
+}
+
+export interface NetworkSetupOption {
+	sender: SetupOption
+	receiver: SetupOption
+	relay: SetupOption
+}
+
+export interface BasisTxContext {
```
ntn-x2, line 23: You mean

```suggestion
export interface BasicTxContext {
```

?

integration-tests/chopsticks/src/tests/xcm/types.ts
```diff
@@ -0,0 +1,48 @@
+import type { ApiPromise } from '@polkadot/api'
+import type { SubmittableExtrinsic } from '@polkadot/api/types'
+import type { KeyringPair } from '@polkadot/keyring/types'
+import type { SetupOption, EventFilter } from '@acala-network/chopsticks-testing'
+
+export interface Storage {
```
ntn-x2, line 6: Ok so this assumes the tests always include sender, relay, and receiver? What if we want to test XCM hopping on two different chains? What if we don't care about the receiver storage? I think a simpler and more generic option would be to simply use a `Record<string, Record<string, Record<string, unknown>>>`, so basically replacing your first-class properties with a key, which you can set and get in each test case. In that case, this whole type could be just a type alias like `export type Storage = Record<string, Record<string, Record<string, unknown>>>`. You can of course refactor each nested type into its own for more clarity, if you wish.

ntn-x2: If you want, you can then `export const storageKey = <string>` for each chain config.

integration-tests/chopsticks/src/tests/xcm/types.ts
```diff
@@ -0,0 +1,48 @@
...
+	receiverStorage: Record<string, Record<string, unknown>>
+	relayStorage: Record<string, Record<string, unknown>>
+}
+
+export interface Accounts {
+	senderAccount: KeyringPair
+	receiverAccount: KeyringPair
+}
+
+export interface NetworkSetupOption {
```
ntn-x2, line 17: Same reasoning applies here as for the `export interface Storage` above.

integration-tests/chopsticks/src/tests/xcm/types.ts
```diff
@@ -0,0 +1,48 @@
...
+		sender: EventFilter[]
+		receiver: EventFilter[]
+	}
+}
+
+export interface BasicConfig {
+	desc: string
+}
+
+interface SovereignAccount {
```
ntn-x2, line 36: I think also here. You should use each chain's identifier to index into the map of accounts.

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransfer/config.ts
```diff
@@ -0,0 +1,293 @@
...
+
+import type { ApiPromise } from '@polkadot/api'
+import type { BasicConfig, BasicXcmTestConfiguration, BasisTxContext } from '../types.js'
+import type { SubmittableExtrinsic } from '@polkadot/api/types'
+
+interface Config extends BasicConfig {
+	precision: bigint
+}
+
+interface Query {
```
ntn-x2, line 15: I think here it's ok to make it specific to only sender and receiver since we know we are dealing with two chains only. So this is ok in this context.

integration-tests/chopsticks/src/tests/xcm/types.ts
ntn-x2: Also, same reasoning about a more generic approach at least for the `pallets` property applies.

integration-tests/chopsticks/src/tests/xcm/limitedReserveTransfer/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+			senderAccount = a
+			receiverAccount = b
+		}, 20_000)
+
+		afterEach(async () => {
+			try {
+				await shutDownNetwork([senderContext, receiverContext, relayContext])
+			} catch (error) {
+				if (!(error instanceof TypeError)) {
+					console.error(error)
```
ntn-x2, line 50: How is this different then just not `catch`ing the error at all? It will be printed to `consoler.err` anyway. Or you don't want to print type errors to console error? If not, why not?

integration-tests/chopsticks/src/helper/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Keyring instance to generate key pairs with 38 ss58Format encoding.
+ */
+const keyring = new Keyring({ type: 'ed25519', ss58Format: 38 })
+
+/**
+ * Key pairs for Alice, Bob, and Charlie used for testing.
+ */
+export const keysAlice = keyring.addFromUri('//alice', undefined, 'ed25519')
```
ntn-x2, line 14: Key type is redundant if the keyring already contains that info, as is the case here.

integration-tests/chopsticks/src/network/hydraDx.ts
```diff
@@ -1,15 +1,22 @@
...
+	({
+		endpoint: process.env.HYDRADX_WS || ['wss://hydradx-rpc.dwellir.com', 'wss://rpc.hydradx.cloud'],
+		db: './db/hydradx.db.sqlite',
+		port: toNumber(process.env.HYDRADX_PORT) || 9001,
+		blockNumber,
+		wasmOverride,
+	}) as SetupOption
 
 /// Sets the [TechnicalCommittee] and [Council] governance to the given accounts
 export function setGovernance(addr: string[]) {
```
ntn-x2, line 22: You could group all storage modifiers into a `storage` object, to make it clear these do not do anything other than returing a specific storage component.

integration-tests/chopsticks/src/network/hydraDx.ts
```diff
@@ -37,12 +44,13 @@ export function assignKiltTokensToAccounts(addr: string[], balance: bigint = ini
 	}
 }
 
+// Sibling Sovereign Account
+export const siblingSovereignAccount = '5Eg2fntQqFi3EvFWAf71G66Ecjjah26bmFzoANAeHFgj9Lia'
```
ntn-x2, line 48: I think it would also be better to group "required" info for each parachain, such as its sibling account and para ID into its own object with its own defined type. Everything else is extra, which can extend the basic object, kinda like you did for the tests setup.

integration-tests/chopsticks/src/network/index.ts
```diff
@@ -0,0 +1,40 @@
...
+	if (value === undefined) {
+		throw new Error(`Environment variable ${name} is not set.`)
+	}
+	return value
+}
+
+/*
+ * Object with all the chain configurations.
+ */
+export const all: ChainConfigs = {
```
ntn-x2, line 20: Is this really needed?

integration-tests/chopsticks/src/network/index.ts
```diff
@@ -0,0 +1,40 @@
+import * as SpiritnetConfig from './spiritnet.js'
+import * as PolkadotConfig from './relay.js'
+import * as HydraDxConfig from './hydraDx.js'
+import { ChainConfigs } from './types.js'
+
+/*
+ * Get an environment variable and throw an error if it is not set.
+ */
+function getEnvVariable(name: string): string {
```
ntn-x2, line 9: Then I would rename this function to:

```suggestion
function getRequiredEnvVariable(name: string): string {
```

integration-tests/chopsticks/src/network/utils.ts
```diff
@@ -1,53 +1,73 @@
...
+ * @param {SetupOption} relayChain - The relay chain option for the network setup.
+ * @param {SetupOption} sender - The sender option for the network setup.
+ * @param {SetupOption} receiver - The receiver option for the network setup.
+ *
+ * @returns {Promise<{relayChainContext: Config, senderChainContext: Config, receiverChainContext: Config}>}
+ * An object containing the contexts of the relay chain, sender, and receiver.
+ */
+export async function setupNetwork(
+	relayChain: SetupOption,
+	sender: SetupOption,
```
ntn-x2, line 50: This as well could involve configuring more than a single receiver chain, so it should probably be generalized.

ntn-x2: If you need to treat the realy chain differently because of a different function name, then I would simply have the relay chain argument and a list of chains to all connect horizontally with each other and vertically with the relaychain. As far as I can see, you don't need to know who the sender is here.

integration-tests/chopsticks/src/network/spiritnet.ts
```diff
@@ -28,11 +37,7 @@ export function setGovernance(addr: string[]) {
 
 /// Spiritnet ParaId
 export const paraId = 2086
+export const KILT = { Concrete: { parents: 0, interior: 'Here' } }
```
ntn-x2, line 40: The naming of this could be improved. What this is, is the `Here` location within the context of the chain itself. So perhaps just call it `HERE`?

ntn-x2 (commented):
I will definitely give it another review once the comments are addressed and the conflicts with `develop` are resolved. In general, I think the new approach is viable, as long as it is not too much focused on our current set of tests but could be expanded with relatively little effort, which is the main goal of this PR, besides improving test reliability. Hence I would suggest to make all the different components more generic. I have not gone into the test logic itself yet, and I'll keep that for my new review!

Ad96el:
Superseded by: https://github.com/KILTprotocol/kilt-node/pull/839

PR #652, opened 2024-06-04
fix: CLI 

Ad96el:
`cli.run.base` is used to run the node (providing no CLI argument), which is not the case for benchmarking, exporting WASM, or other tools. The `SharedParameter` struct was empty, even if the flags were set in the CLI.

By using the `SharedParameters` from the subcommands, CLI flags are taken into account.

### DIFF
--- a/nodes/parachain/src/command.rs
+++ b/nodes/parachain/src/command.rs
@@ -158,14 +158,14 @@ pub(crate) fn run() -> sc_cli::Result<()> {
 		Some(Subcommand::Benchmark(cmd)) => {
 
 			let shared_params = match cmd {
-				BenchmarkCmd::Block(c) => &c.shared_params,  
-				BenchmarkCmd::Pallet(c) => &c.shared_params,  
+				BenchmarkCmd::Block(c) => &c.shared_params,
+				BenchmarkCmd::Pallet(c) => &c.shared_params,
 				BenchmarkCmd::Extrinsic(c) => &c.shared_params,
-				BenchmarkCmd::Machine(c) => &c.shared_params, 
-				BenchmarkCmd::Overhead(c) => &c.shared_params,  
+				BenchmarkCmd::Machine(c) => &c.shared_params,
+				BenchmarkCmd::Overhead(c) => &c.shared_params,
 				BenchmarkCmd::Storage(c) => &c.shared_params,
 			};
- 
+
 			let (_, runtime) = get_selected_chainspec(shared_params)?;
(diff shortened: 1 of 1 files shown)


abdulmth (approved):
LGTM

nodes/parachain/src/command.rs
```diff
@@ -121,7 +121,7 @@ pub(crate) fn run() -> sc_cli::Result<()> {
 			})
 		}
 		Some(Subcommand::ExportGenesisState(cmd)) => {
-			let (chain_spec_id, runtime) = get_selected_chainspec(&cli)?;
+			let (chain_spec_id, runtime) = get_selected_chainspec(&cmd.shared_params)?;
 			let spec = cli.load_spec(chain_spec_id.as_str())?;
 
 			println!("Dispatching task for spec id: {chain_spec_id}.");
```
abdulmth, line 127: There is this print call too, not sure if it should be removed.

abdulmth (dismissed):

nodes/parachain/src/command.rs
Ad96el: It was already there. I just removed the debug prints that I needed to troubleshoot the issue. So, I think it's fine now.

Ad96el:
@ntn-x2 for your information

abdulmth (approved):

nodes/parachain/src/command.rs
```diff
@@ -35,16 +35,16 @@ use crate::{
 
 // Returns the provided (`--chain`, <selected_runtime>) given only a reference
 // to the global `Cli` object.
```
ntn-x2, line 37: The comment must be updated as well.

nodes/parachain/src/command.rs
```diff
@@ -149,14 +149,24 @@ pub(crate) fn run() -> sc_cli::Result<()> {
...
+				let (chain_spec_id, _) = get_selected_chainspec(&cmd.shared_params)?;
 				let spec = cli.load_spec(chain_spec_id.as_str())?;
 
 				cmd.run(&*spec)
 			})
 		}
 		Some(Subcommand::Benchmark(cmd)) => {
-			let (_, runtime) = get_selected_chainspec(&cli)?;
+
+			let shared_params = match cmd {
```
ntn-x2, line 160: So there's no trait implementation that can be used on the `cmd` object to get the `shared_params` property? What about the `SubstrateCli` trait?

PR #658, opened 2024-06-17
feat: switch functionality

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3467.

Not meant to be merged until all other stacked PRs will be merged into this.

Adds the first round of features to the swap pallet, specifically it includes a swap pool that supports only the runtime local currency, swappable 1:1 (not configurable yet) with a configurable remote token. The XCM fees are paid in a local fungible token as injected by the runtime. The pallet also exposes two implementations of the `IsReserve` trait that must be added to the runtime to allow the configured tokens (to swap back to local currency and to pay for XCM fees) to be reserve-transferred to this chain. That code is not yet part of this PR.

Additional TODOs for the pallet is both within the code where applicable, and in the pallet `README.md` file.


Ad96el (commented):
LGTM. I have some minor questions, mainly to improve my understanding

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		SwapPairCreated {
+			remote_asset_id: VersionedAssetId,
+			remote_asset_reserve_location: VersionedMultiLocation,
```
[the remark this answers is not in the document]
ntn-x2: This is the location from which we receive an XCM message. `AssetId` has a concrete and abstract variant, but we deal with `MultiLocation`s when reasoning about the origin of an XCM message.

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+			T::SwapOrigin::ensure_origin(origin)?;
+
+			// 1. Verify swap pair has not already been set.
+			ensure!(!SwapPair::<T>::exists(), Error::<T>::SwapPairAlreadyExisting);
+			// 2. Verify that total issuance >= circulating supply and take the difference
+			//    as the amount of assets we control on destination.
+			let locked_supply = total_issuance
+				.checked_sub(circulating_supply)
+				.ok_or(Error::<T>::InvalidInput)?;
+			// 3. Verify the pool account does not exist and has enough local assets to
```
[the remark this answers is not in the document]
ntn-x2: The comment is wrong, yes, since we need to check that the account has enough funds. But this is a pre-condition: a pool for an asset cannot be created if its to-be pool account does not have enough assets to cover all the swaps, so it needs to be pre-funded with a regular transfer by whoever.

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+				)?;
+			if withdrawn_fees != vec![remote_fee_asset_v3.clone()].into() {
+				log::error!(
+					target: LOG_TARGET,
+					"Withdrawn fees {:?} does not match expected fee {:?}.",
+					withdrawn_fees,
+					remote_fee_asset_v3
+				);
+				return Err(DispatchError::from(Error::<T>::Internal));
+			}
```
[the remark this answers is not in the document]
ntn-x2: No. What we withdraw here are the bridged DOTs. So we simply need to "take" them from there, since we will use them on AssetHub. We don't transfer them anywhere on our chain. Is it clear?

pallets/pallet-asset-swap/src/xcm/match_fungible.rs
```diff
@@ -0,0 +1,56 @@
...
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use frame_support::ensure;
+use sp_std::marker::PhantomData;
+use xcm::v3::{AssetId, Fungibility, MultiAsset, MultiLocation};
+use xcm_executor::traits::{Error as XcmExecutorError, MatchesFungibles};
+
+use crate::{Config, SwapPair, SwapPairInfoOf, LOG_TARGET};
+
+pub struct MatchesSwapPairXcmFeeAsset<T>(PhantomData<T>);
```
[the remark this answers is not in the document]
ntn-x2: This is used in the XCM config (upcoming PR). We need to pass this as the `TransactAsset` of the XCM config to allow the subject to pay for fees in DOTs.

pallets/pallet-asset-swap/src/xcm/reserve.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use frame_support::traits::ContainsPair;
+use sp_std::marker::PhantomData;
+use xcm::prelude::{AssetId, MultiAsset, MultiLocation};
+
+use crate::{Config, SwapPair, LOG_TARGET};
+
+pub struct ReserveTransfersOfXcmFeeAssetAndRemoteAsset<T>(PhantomData<T>);
```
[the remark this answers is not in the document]
ntn-x2: This is used in the XCM fee to allow reserve-based transfers from the configured remote location of the relevant tokens that we need, namely the token to be swapped back for KILTs, and the token to be used for XCM fees. If we don't allow those to enter our network, people will never be able to do any swaps.

pallets/pallet-asset-swap/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Sorry Not completely.  
Is the `withdraw_asset` function sending a XCM message to AssetHub, where the `withdrawn_fees` are taken from the sovereign account?

pallets/pallet-asset-swap/src/xcm/transact.rs
```diff
@@ -0,0 +1,111 @@
...
+				))?;
+		SwapPair::<T>::try_mutate(|entry| {
+			let SwapPairInfoOf::<T> {
+				remote_asset_balance, ..
+			} = entry
+				.as_mut()
+				.ok_or(XcmError::FailedToTransactAsset("SwapPair should not be None."))?;
+			*remote_asset_balance = new_remote_balance;
+			Ok::<_, XcmError>(())
+		})?;
```
Ad96el, line 107: I am not sure about this one. I know we checked that the XCM message can be transmitted to AssetHub, but I assume those checks are mainly local, verifying whether a channel exists or not. I am concerned that some messages on AssetHub might not be executed or might throw an error in specific edge cases. If this happens, the invariant could get corrupted.

pallets/pallet-asset-swap/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: No. Here we are withdrawing assets from the user's local DOT balance. Since we know that we will use that on AssetHub to pay for the fees. But the user having that many DOTs means they previously sent them to our sovereign account, so we are using DOTs he previously sent us. So now what we do here is simply saying: "I'm reducing your balance by this much since we will use this much on AssetHub for the wrapped token transfer". It's essentially the same as burning those many DOTs from our local representation.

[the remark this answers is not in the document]
ntn-x2: But we already own those DOTs on AssetHub, as they have been reserve-transferred to us alraedy.

pallets/pallet-asset-swap/src/xcm/transact.rs
ntn-x2: This cannot happen. We are 100% sure the transfer happened remotely if we are even receiving the XCM message here, so we can safely increase the counter. The opposite is not true tho, meaning that we are optimistically reducing the balance here before the full transfer of eKILTs happens on AssetHub. I thought about this. There would need to be a "ping-back" mechanism where we tell KILT, from AssetHub, that something went ok or wrong. We could use XCM message hashes, and temporarily store a map of hash -> amount locally. Every time we get a successful message, we know the transfer happened, so we are N KILTs "poorer" on AssetHub. I will mark this as something to do for the pallet V2.

ntn-x2: This is called when a user is trying to "reserve-deposit" eKILTs in Spiritnet. For each eKILT that AssetHub guarantees us has been moved to our sovereign account there, we release one KILT from the pool account. So the condition you are talking about has already been verified by AH before dispatching the message that we are in the middle of processing here.

Ad96el: Yes, you are right. We are here 100% save. The fees will also be paid by the KILT transferred by the XCM message, so this will also not fail. 

The optimistic reducing seems also fine, but the user needs somehow a mechanism to claim back his KILT, which are not processed on AH. I am sure this will be part of v2.

ntn-x2: That would be an asset trap, or we can simply add instructions to deposit whatever is left after an error to the user's balance.

pallets/pallet-asset-swap/src/lib.rs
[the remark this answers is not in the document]
Ad96el: I see okay. I was assuming the `remote_asset_reserve_location` had another purpose

pallets/pallet-asset-swap/src/lib.rs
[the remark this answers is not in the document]
Ad96el: I see okay

pallets/pallet-asset-swap/src/xcm/transact.rs
Ad96el: This seems to me like an attack point, or not? A user can send a bunch of extrinsic, which will fail on AH.

Therefore, the `remote_asset_balance` will be so low that no one can do a swap. 

On AH, he can claim those coins and transfer them back to KILT, with no huge financial damage

Ad96el: Ah no if he transfers those back the remote_asset_balance will be increased again

Ad96el: Yeah, okay, he would only damage himself

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+			// 6. Compose and validate XCM message
+			let appendix: Xcm<()> = vec![
+				RefundSurplus,
+				DepositAsset {
+					assets: MultiAssetFilter::Wild(WildMultiAsset::All),
+					beneficiary: submitter_as_multilocation,
+				},
+			]
+			.into();
```
Ad96el, line 365: This seems incorrect to me. The beneficiary should be sovereign account on AH. This is the account which paid  the fee

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+			let remote_xcm: Xcm<()> = vec![
+				WithdrawAsset(remote_asset_fee_v3.clone().into()),
```
Ad96el, line 367: I guess on AH we are only allowed to pay the fees with dots. I think we have to still withdraw those assets.

pallets/pallet-asset-swap/src/lib.rs
ntn-x2: No. What we are doing here, is:
1. We take the fee from the user's balance on our chain
2. We withdraw the same amount of assets on AH from our SA for the fees
3. We refund back **to the user** any unused fees

Hopefully with the fee being as close as possible to the actual one, the refund won't be too much, but we're still refunding it back into the user's account.

pallets/pallet-asset-swap/src/lib.rs
ntn-x2: No, you can also pay with other assets. But here the `remote_asset_fee_v3` is set by KILT governance when setting a swap pair, and in the case of assets bridged to AH, that will be indeed the DOT token. But if the token is wrapped on a different chain, the asset that will be used for the fees might be different than the DOT. The pallet is not coupled to a specific asset.

pallets/pallet-asset-swap/src/lib.rs
Ad96el: I see. I misunderstood something

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+				status,
+				..
+			} = entry.as_mut().ok_or(Error::<T>::SwapPairNotFound)?;
+			let relevant_event = match new_status {
+				SwapPairStatus::Running => Event::<T>::SwapPairResumed {
+					remote_asset_id: remote_asset_id.clone(),
+				},
+				SwapPairStatus::Paused => Event::<T>::SwapPairPaused {
+					remote_asset_id: remote_asset_id.clone(),
+				},
```
Ad96el, line 520: A bit picky, but we should also add the remote_reserve_location in those events.

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+		<<T as Config>::LocalCurrency as InspectFungible<<T as frame_system::Config>::AccountId>>::Balance;
+	pub type SwapPairInfoOf<T> = SwapPairInfo<<T as frame_system::Config>::AccountId>;
+
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		const POOL_ADDRESS_GENERATION_ENTROPY: [u8; 8];
+
+		type AccountIdConverter: TryConvert<Self::AccountId, Junction>;
```
Ad96el, line 78: Could you please explain, why this is needed? This feels to me somehow overengineered

ntn-x2: It does feel, but then if you think about what the definition of `AccountId` is for `frame_system`, you realise you cannot assume it will always be `AccountId32`. For instance, a parachain using `AccountId20` would need to use a different converter. If we bake in the assumption that an account ID is always an `AccountId32`, then the pallet is not usable anywhere else.

pallets/pallet-asset-swap/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+mod swap;
+pub use swap::{SwapPairInfo, SwapPairStatus};
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+
+use ::xcm::{VersionedAssetId, VersionedMultiAsset, VersionedMultiLocation};
```
Ad96el, line 31: Is this intendant?

ntn-x2: Yes. We have a module called `xcm`. Here we want to use the `xcm` crate, so that's Rust syntax for it.

pallets/pallet-asset-swap/src/xcm/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+pub mod convert;
+pub use convert::AccountId32ToAccountId32JunctionConverter;
+
+pub mod r#match;
+pub use r#match::MatchesSwapPairXcmFeeFungibleAsset;
```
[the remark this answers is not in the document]
ntn-x2: Nope 😃 `match` is a reserved keyword, so referring to a module named `match`, that's the right syntax.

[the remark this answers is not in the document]
ntn-x2: Of course, if you think we should rename modules from verb to nouns (e.g., `matcher` instead of `match`) we could do that, although I'm not sure how it would look like for `transfer`.

pallets/pallet-asset-swap/src/lib.rs
ntn-x2: Also, the pallet supports a single pair, so there's not even confusion for any external indexers. I really don't see any benefits in also including the remote location here, but I'm open to change my mind.

Ad96el: I was assuming, once there are multiple swapPairs, it could get ambiguous, or is it out of discussion and we only support a single swapPair?

ntn-x2: If we would support multiple pairs within the same pallet, then we would probably add the local and remote asset IDs in the event, but still for the reserve location there would be no need, since it cannot change afterwards anyway. Indexers would probably stored it the first time they see it, and assume that will stay like that.

Ad96el: Yes, sorry. That's true. I was assuming the smart contract could be deployed on multiple EVM based chains, but the remote asset IDs would be different every time.

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
+# Asset switch pallet
+
+The asset switch pallet introduces the possibility for the chain local currency to be switched 1:1 with a remote asset at a remote destination, according to the provided configuration, using XCM.
+
+This is possible via the creation of a *switch pair*, which contains information about the identifier of the remote asset (e.g., its `MultiLocation`), the remote location on which the asset lives (and with which XCM communication takes place), the circulating supply of the remote asset which can be switched back for the local currency, and additional information relevant for the XCM communication, which are explained more in depth later on.
+
+## In a gist
```
ChrisChinchilla, line 7:
```suggestion
## Summary
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
+# Asset switch pallet
+
+The asset switch pallet introduces the possibility for the chain local currency to be switched 1:1 with a remote asset at a remote destination, according to the provided configuration, using XCM.
```
ChrisChinchilla, line 3:
```suggestion
The asset switch pallet allows for switching the chain local currency 1:1 with a remote asset at a remote destination, according to the provided configuration, and using XCM.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+This is possible via the creation of a *switch pair*, which contains information about the identifier of the remote asset (e.g., its `MultiLocation`), the remote location on which the asset lives (and with which XCM communication takes place), the circulating supply of the remote asset which can be switched back for the local currency, and additional information relevant for the XCM communication, which are explained more in depth later on.
```
ChrisChinchilla, line 5:
```suggestion
This is possible by creating a *switch pair*, which contains information about the remote asset's identifier (e.g., its MultiLocation`), the remote location where the asset lives (and with which XCM communication takes place), the circulating supply of the remote asset, which can be switched back for the local currency, and additional information relevant for the XCM communication, which is explained more in-depth later on.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## In a gist
+
+The pallet lets users on the source chain lock `N` tokens of the chain's native currency into a switch pair specific account, to receive `N` remote assets on the configured remote location, which are transferred to them from the source chain's sovereign account at destination.
+Then, the chain, which i
```
ChrisChinchilla, line 10: ?

```suggestion
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+    // Config
+}
+```
+
+## The `Config` trait
+
+As the pallet is generic over the runtime specifics, the `Config` trait requires the following configuration parameters passed to it:
+
+- `type AccountIdConverter: TryConvert<Self::AccountId, Junction>`: Because the `AccountId` type can be anything in the runtime, the converter is responsible for converting such a `AccountId` into a `Junction`, which is then used for some XCM processing.
+- `type AssetTransactor: TransactAsset`: This component is used when charging the extrinsic submitter with the XCM fees that the chain will pay at destination. For instance, if the transfer on the remote chain will cost 0.1 DOTs, the `AssetTransactor` might deduct 0.1 DOTs from the user's previously topped up balance on the source chain (more details below).
```
ChrisChinchilla, line 65:
```suggestion
- `type AssetTransactor: TransactAsset`: This component is used when charging the extrinsic submitter with the XCM fees that the chain will pay at the remote chain. For instance, if the transfer on the remote chain will cost 0.1 DOTs, the `AssetTransactor` might deduct 0.1 DOTs from the user's previously topped up balance on the source chain (more details below).
```

Again, if this is correct.

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+- `type FeeOrigin: EnsureOrigin<Self::RuntimeOrigin>`: The origin that can update the XCM fee to be paid for the transfer on destination.
```
ChrisChinchilla, line 66:
```suggestion
- `type FeeOrigin: EnsureOrigin<Self::RuntimeOrigin>`: The origin that can update the XCM fee to be paid for the transfer on the remote chain.
```

Again, if appropriate.

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+- `type LocalCurrency: MutateFungible<Self::AccountId>`: The chain's local currency.
+- `type PauseOrigin: EnsureOrigin<Self::RuntimeOrigin>`: The origin that can pause a switch pair, e.g., if a vulnerability is found.
+- `type RuntimeEvent: From<Event<Self, I>> + IsType<<Self as frame_system::Config>::RuntimeEvent>`: The aggregate `Event` type.
+- `type SubmitterOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>`: The origin that can call the `switch` extrinsic and perform the switch.
+- `type SwitchHooks: SwitchHooks<Self, I>`: Any additional runtime-specific logic that can be injected both before and after local tokens are exchanged for the remote assets, and before and after the remote assets are being converted into local tokens.
```
ChrisChinchilla, line 71:
```suggestion
- `type SwitchHooks: SwitchHooks<Self, I>`: Any additional runtime-specific logic that can be injected both before and after local tokens are exchanged for the remote assets, and before and after the remote assets are converted into local tokens.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+- `type SwitchOrigin: EnsureOrigin<Self::RuntimeOrigin>`: The origin that can set, resume, and delete a switch pair.
+- `type XcmRouter: SendXcm`: The component responsible for routing XCM messages to the switch pair remote location to perform the remote asset transfer from the chain's sovereign account to the specified beneficiary.
```
ChrisChinchilla, line 73: Is this the component, or does the comment relate to an XCM component?

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
...st be at least as large as `circulating_supply`. It is possible for `circulating_supply` to be `0`, in which case it means this chain controls all of the `total_issuance` of the remote asset, which can be obtained by locking a corresponding amount of local tokens via the `switch` call below. Furthermore, the pallet calculates the account that will hold the local tokens locked in exchange for remote tokens: this account is based on the pallet runtime name as returned by the `PalletInfoAccess` trait, and the value of `remote_asset_id`. The generated account must already have a balance of at least `circulating_supply`, ensuring that there are enough local tokens locked to satisfy all the requests to exchange the remote asset for local tokens. The balance of such account can be increased with a simple transfer, after obtaining the to-be-created switch pair pool account by interacting with the [asset-switch runtime API][asset-switch-runtime-api]: this requirement is bypassed with the `force_set_switch_pair` call. Only `SwitchOrigin` can call this, and in most cases it will most likely be a governance-based origin such as the one provided by referenda or collectives with high privileges.
```
ChrisChinchilla, line 95:
```suggestion
1. `pub fn set_switch_pair(origin: OriginFor<T>, reserve_location: Box<VersionedMultiLocation>, remote_asset_id: Box<VersionedAssetId>, remote_fee: Box<VersionedMultiAsset>, total_issuance: u128, circulating_supply: u128) -> DispatchResult`: Set a new switch pair between the local currency and the specified `remote_asset_id` on the `reserve_location`. The specified `total_issuance` includes both the `circulating_supply` (i.e., the remote asset amount that the chain does not control on the `reserve_location`) and the locked supply under the control of the chain's sovereign account on the `reserve_location`. For this reason, the value of `total_issuance` must be at least as large as `circulating_supply`. It is possible for `circulating_supply` to be `0`, in which case it means this chain controls all of the `total_issuance` of the remote asset, which can be obtained by locking a corresponding amount of local tokens via the `switch` call below.

Furthermore, the pallet calculates the account that will hold the local tokens locked in exchange for remote tokens. This account is based on the pallet runtime name as returned by the `PalletInfoAccess` trait and the value of `remote_asset_id`. The generated account must already have a balance of at least `circulating_supply`, ensuring enough local tokens are locked to satisfy all requests to exchange the remote asset for local tokens. The balance of such an account can be increased with a simple transfer after obtaining the to-be-created switch pair pool account by interacting with the [asset-switch runtime API][asset-switch-runtime-api].

This requirement can be bypassed with the `force_set_switch_pair` call. Only `SwitchOrigin` can call this, and in most cases it will most likely be a governance-based origin such as the one provided by referenda or collectives with high privileges.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+2. `pub fn force_set_switch_pair(origin: OriginFor<T>, reserve_location: Box<VersionedMultiLocation>, remote_asset_id: Box<VersionedAssetId>, remote_fee: Box<VersionedMultiAsset>, total_issuance: u128, circulating_supply: u128) -> DispatchResult`: It does the same as the `set_switch_pair`, but it skips the check over the switch pair pool account balance, and it requires the `root` origin for the call to be dispatched.
```
ChrisChinchilla, line 96:
```suggestion
2. `pub fn force_set_switch_pair(origin: OriginFor<T>, reserve_location: Box<VersionedMultiLocation>, remote_asset_id: Box<VersionedAssetId>, remote_fee: Box<VersionedMultiAsset>, total_issuance: u128, circulating_supply: u128) -> DispatchResult`: The same as the `set_switch_pair`, but skips the check over the switch pair pool account balance, and requires the `root` origin for the call to be dispatched.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+3. `pub fn force_unset_switch_pair(origin: OriginFor<T>) -> DispatchResult`: Forcibly remove a previously-stored switch pair. This operation can only be accomplished by the `root` origin. **Any intermediate state, such as local tokens locked in the switch pair pool or remote assets that are not switchable anymore for local tokens must be taken care of with subsequent governance operations.**
```
ChrisChinchilla, line 97:
```suggestion
3. `pub fn force_unset_switch_pair(origin: OriginFor<T>) -> DispatchResult`: Forcibly remove a previously-stored switch pair. This operation can only be called by the `root` origin.

**Any intermediate state, such as local tokens locked in the switch pair pool or remote assets that are not switchable anymore for local tokens, must be taken care of with subsequent governance operations.**
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+4. `pub fn pause_switch_pair(origin: OriginFor<T>) -> DispatchResult`: Allows the `PauseOrigin` to immediately pause switches in both directions.
+5. `pub fn resume_switch_pair(origin: OriginFor<T>) -> DispatchResult`: Allows the `SwitchOrigin` to resume switches in both directions.
+6. `pub fn update_remote_fee(origin: OriginFor<T>, new: Box<VersionedMultiAsset>) -> DispatchResult`: Allows the `FeeOrigin` to update the required XCM fee to execute the transfer of remote asset on the reserve location from the chain's sovereign account to the beneficiary specified in the `switch` operation. For example, if the cost of sending an XCM message containing a `TransferAsset` instruction from the source chain to AssetHub (reserve location) changes from 0.1 DOTs to 0.2 DOTs, the fee will need to be updated accordingly to avoid transfers to fail on AssetHub, leaving the whole system in an inconsistent state. Since the pallet refunds any unused assets on the reserve location to the account initiating the switch on the source chain, it is not a big issue to overestimate this value here, since no funds will be burnt or unfairly taken from the user during the switch process.
```
ChrisChinchilla, line 100:
```suggestion
6. `pub fn update_remote_fee(origin: OriginFor<T>, new: Box<VersionedMultiAsset>) -> DispatchResult`: Allows the `FeeOrigin` to update the required XCM fee to execute the transfer of remote asset on the reserve location from the chain's sovereign account to the beneficiary specified in the `switch` operation.

For example, if the cost of sending an XCM message containing a `TransferAsset` instruction from the source chain to AssetHub (reserve location) changes from 0.1 DOTs to 0.2 DOTs, the fee will need to be updated accordingly to avoid transfers failing on AssetHub, leaving the whole system in an inconsistent state. Since the pallet refunds any unused assets on the reserve location to the account initiating the switch on the source chain, it is not a big issue to overestimate this value here since no funds will be burnt or unfairly taken from the user during the switch process.
```

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,129 @@
...
+	2. No switch pair is set or the switch pair is currently not allowing switches.
+	3. There are not enough locked remote assets on the `reserve_location` to cover the switch request. E.g., if the chain sovereign account on the `reserve_location` only controls `10` remote assets, users can only switch up to `10` local tokens. Once the limit is reached, there needs to be someone performing the reverse operation (remote -> local switch) to free up some remote tokens.
+	4. The switch pair `reserve_location` is not reachable from the source chain, because the configured `XcmRouter` returns an error (e.g., there is no XCM channel between the two chains).
+	5. The configured `SwitchHooks` returns an error in either the `pre-` or the `post-` switch checks.
+	6. The user does not have enough assets to pay for the required remote XCM fees as specified in the switch pair info and as returned by the configured `AssetTransactor`.
+
+## XCM components
+
+Because the switch functionality relies on XCM, the pallet provides a few XCM components that are to be included in a runtime in order to enable the whole set of interactions between the source chain and the configured remote reserve location.
```
ChrisChinchilla, line 111:
```suggestion
Because the switch functionality relies on XCM, the pallet provides a few XCM components that should be included in a runtime to enable the whole set of interactions between the source chain and the configured remote reserve location.
```

pallets/pallet-asset-switch/README.md
ntn-x2: This is the component.

pallets/pallet-asset-switch/README.md
ntn-x2: This is a bullet point, I don't think we can add whitespaces without breaking the list, can we?

ntn-x2:
@ChrisChinchilla I addressed most of your comments in a separate PR: https://github.com/KILTprotocol/kilt-node/pull/673.

pallets/pallet-asset-switch/README.md
ChrisChinchilla: Ah yes, it needs to be intended, though to be fair, this is very long for a bullet point, but a bit of whitespace will help.

ChrisChinchilla: I think that's right, GitHub makes it hard to suggest indents, if not, I'll just submit the change directly.

ntn-x2: Ok I'll merge this suggestion in the PR I opened.

runtimes/peregrine/src/tests.rs
```diff
@@ -41,15 +41,18 @@ use runtime_common::{
...
-#[test]
-fn call_size() {
-	assert!(
-		core::mem::size_of::<RuntimeCall>() <= 240,
-		"size of Call is more than 240 bytes: some calls have too big arguments, use Box to reduce \
-		the size of Call.
-		If the limit is too strong, maybe consider increase the limit to 300.",
-	);
-}
+// TODO: Uncomment if pallet_assets implements measures to reduce their `Call`
```
Ad96el, line 44: Is there already a PR? If yes, maybe link it here.

runtimes/peregrine/src/lib.rs
```diff
@@ -969,6 +976,61 @@ impl pallet_proxy::Config for Runtime {
...
+	type Extra = ();
+	type ForceOrigin = EnsureRoot<AccountId>;
+	type Freezer = ();
+	type MetadataDepositBase = ConstU128<0>;
+	type MetadataDepositPerByte = ConstU128<0>;
+	type RemoveItemsLimit = ConstU32<1_000>;
+	type RuntimeEvent = RuntimeEvent;
+	type StringLimit = ConstU32<4>;
+	// TODO: Change before release
+	type WeightInfo = ();
```
Ad96el, line 1031: Can we maybe benchmark the pallet already in this PR?

runtimes/peregrine/src/lib.rs
```diff
@@ -969,6 +976,61 @@ impl pallet_proxy::Config for Runtime {
...
+
+// No deposit is taken since creation is permissioned. Only the root origin can
+// create new assets, and the owner will be the treasury account.
+impl pallet_assets::Config for Runtime {
+	type ApprovalDeposit = ConstU128<0>;
+	type AssetAccountDeposit = ConstU128<0>;
+	type AssetDeposit = ConstU128<0>;
+	type AssetId = Location;
+	type AssetIdParameter = Location;
+	type Balance = u128;
```
Ad96el, line 1016: should we use the Balance type from runtime common?

runtimes/peregrine/src/lib.rs
```diff
@@ -969,6 +976,61 @@ impl pallet_proxy::Config for Runtime {
...
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = crate::benches::CreateFungibleForAssetSwitchPool1;
+}
+
+// No deposit is taken since creation is permissioned. Only the root origin can
+// create new assets, and the owner will be the treasury account.
+impl pallet_assets::Config for Runtime {
+	type ApprovalDeposit = ConstU128<0>;
+	type AssetAccountDeposit = ConstU128<0>;
+	type AssetDeposit = ConstU128<0>;
```
Ad96el, line 1013: I guess all these constants should go to runtime common. Spiritnet will prob use the same

runtimes/peregrine/src/lib.rs
```diff
@@ -969,6 +976,61 @@ impl pallet_proxy::Config for Runtime {
...
+	type AssetId = Location;
+	type AssetIdParameter = Location;
+	type Balance = u128;
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = runtime_common::asset_switch::NoopBenchmarkHelper;
```
Ad96el, line 1018: Would prefer to have the feature guided types at the end of the config implementation

runtimes/peregrine/src/asset_switch/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// Check requiring the beneficiary be a single `AccountId32` junction
+/// containing the same account ID as the account on this chain initiating the
+/// switch.
+pub struct RestrictswitchDestinationToSelf;
+
+impl SwitchHooks<Runtime, KiltToEKiltSwitchPallet> for RestrictswitchDestinationToSelf {
+	type Error = Error;
+
+	fn pre_local_to_remote_switch(
```
Ad96el, line 38: should we maybe add a unit test for this function?

Ad96el (commented):
Some minor comments

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,141 @@
+# Asset switch pallet
+
+The asset switch pallet allows for switching the chain local currency 1:1 with a remote asset at a remote destination, according to the provided configuration, and using XCM.
```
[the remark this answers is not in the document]
ntn-x2: Yes, that's true, but 99% of the cases it will be local currency. I would not make things even more confusing, especially in the first line.

pallets/pallet-asset-switch/src/lib.rs
```diff
@@ -0,0 +1,686 @@
...
+		LocalCurrencyBalanceOf<T, I>: Into<u128>,
+	{
+		/// Set a new switch pair.
+		///
+		/// See the crate's README for more.
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as Config<I>>::WeightInfo::set_switch_pair())]
+		pub fn set_switch_pair(
+			origin: OriginFor<T>,
+			remote_asset_total_supply: u128,
```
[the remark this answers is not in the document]
ntn-x2: Yes but we can't do anything about it. Also the fungibility of XCM assets uses `u128`. So that's the maximum that we can safely assume can be represented over XCM.

runtimes/peregrine/src/tests.rs
ntn-x2: No, there's not one, as far as I could find.

runtimes/peregrine/src/lib.rs
ntn-x2: I would move them once we implement the stuff in the Spiritnet runtime as well. Not really a concern for this PR.

runtimes/peregrine/src/asset_switch/mod.rs
ntn-x2: Nah, I think we don't need to at this stage. If we add more complex logic then yes.

pallets/pallet-asset-switch/src/lib.rs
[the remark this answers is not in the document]
Ad96el: Ah yeah true. Fungible is set to u128

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+This is possible by creating a *switch pair*, which contains information about the remote asset's identifier (e.g., its MultiLocation`), the remote location where the asset lives (and with which XCM communication takes place), the circulating supply of the remote asset, which can be switched back for the local currency, and additional information relevant for the XCM communication, which is explained more in-depth later on.
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/692/commits/c93fd69b8a5e68eb5269c1bf5bec3c1bb5141739

pallets/pallet-asset-switch/README.md
```diff
@@ -0,0 +1,141 @@
...
+* The sovereign account of the source chain at destination owns the remote assets in the amount that is specified when the switch pair is created. The validation of this requirement is delegated to the source chain governance in the act leading to the creation of the switch pair, as the source chain itself has currently no means of doing that.
+* Similarly, the pallet has currently no way to verify that a transfer of remote tokens from the chain sovereign account to the **specified** beneficiary has been completed successfully on the remote location, hence it takes an optimistic approach given that all the verifiable preconditions for the switch are verified on the source chain, i.e., the chain from which the switch originates. Any unexpected issues in the transfer on the remote location will most likely require intervention of the source chain governance to re-balance the state and make it consistent again.
+
+## Add the pallet to the runtime
+
+Add the following line to the runtime `Cargo.toml` dependencies section:
+
+```toml
+pallet-asset-switch = {git = "https://github.com/KILTprotocol/kilt-node.git", branch = "release-1.13.0"}
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/692/commits/d8b810f0792970fa0ac8b19d8b5c159830b31410

runtimes/peregrine/src/lib.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/692/commits/f64a63ecc4dd6a7d19fdeea70d2d29c43ceb94a6

runtimes/peregrine/src/lib.rs
ntn-x2: I like how you think: https://github.com/KILTprotocol/kilt-node/pull/692/commits/9237211194e8e5687db496af0cfb5ecd66c8abef.

runtimes/peregrine/src/lib.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/692/commits/68cb4a6741c2ba2d5d4591b7f1cd4367a83ac0b1

ntn-x2:
Comments address in this PR: https://github.com/KILTprotocol/kilt-node/pull/692.

runtimes/peregrine/src/lib.rs
Ad96el: Seeing and adapting

Ad96el (approved):
:partying_face:

PR #660, opened 2024-06-26
fix: swap pallet fixes

ntn-x2:
Fixes the swap pallet after testing with a local Peregrine setup. Based on top of https://github.com/KILTprotocol/kilt-node/pull/658.

PR #661, opened 2024-06-27
fix: properly handle dust from XCM execution

ntn-x2:
The numbers did not match up between what the swap pallet thought the remote balance was, and what the actual remote balance is. This is because there was some dust in the trader `drop()` function we did not handle. That is fixed now, so that whenever a eKILT -> KILT swap happens, the following two things also happen:

1. The amount of remote balance is increased not only by the amount of eKILTs (after XCM fees are taken) being swapped, but also by the amount of fees taken as part of the XCM operation. Now, the real value and the tracked value in the swap pallet always stay consistent.
2. The same corresponding amount of KILTs that are now not swappable because some eKILTs have been used to pay for XCM fees are sent to a configurable account ID (which in the [Peregrine configuration](https://github.com/KILTprotocol/kilt-node/pull/659) is the treasury account). So now the balance of the pool account matches exactly the number of eKILTs that have been swapped back to KILTs.

Based on top of https://github.com/KILTprotocol/kilt-node/pull/660.

### DIFF
--- a/pallets/pallet-asset-swap/src/xcm/trade.rs
+++ b/pallets/pallet-asset-swap/src/xcm/trade.rs
@@ -18,13 +18,10 @@
 
 pub use xcm_fee_asset::UsingComponentsForXcmFeeAsset;
 mod xcm_fee_asset {
-	use frame_support::{traits::fungibles::Inspect, weights::WeightToFee as WeightToFeeT};
-	use sp_runtime::{
-		traits::{Saturating, Zero},
-		SaturatedConversion,
-	};
+	use frame_support::{ensure, weights::WeightToFee as WeightToFeeT};
+	use sp_runtime::traits::Zero;
 	use sp_std::marker::PhantomData;
-	use xcm::v3::{AssetId, Error, MultiAsset, Weight, XcmContext};
+	use xcm::v3::{AssetId, Error, MultiAsset, Weight, XcmContext, XcmHash};
 	use xcm_executor::{traits::WeightTrader, Assets};
 
 	use crate::{Config, SwapPair};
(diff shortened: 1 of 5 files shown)


pallets/pallet-asset-swap/src/xcm/trade.rs
```diff
@@ -36,71 +33,120 @@ mod xcm_fee_asset {
...
-		Fungibles::Balance,
-		PhantomData<(T, WeightToFee)>,
-	);
+	pub struct UsingComponentsForXcmFeeAsset<T, WeightToFee>
+	where
+		T: Config,
+	{
+		remaining_weight: Weight,
+		remaining_fungible_balance: u128,
+		consumed_xcm_hash: Option<XcmHash>,
```
Ad96el, line 42: Do you think tracking the xcm message is necessary? I would assume the messages are processed sequential. So there is no way two xcm message can collide

pallets/pallet-asset-swap/src/xcm/trade.rs
```diff
@@ -109,56 +155,147 @@ mod swap_pair_remote_asset {
...
+						e
+					});
+
+					// No error should ever be thrown from inside this block.
+					SwapPair::<T>::mutate(|entry| {
+						let Some(entry) = entry.as_mut() else {
+							log::error!(target: LOG_TARGET, "Stored swap pair should not be None but it is.");
+							return;
+						};
+						entry.remote_asset_balance = entry
```
Ad96el, line 292: I think a * is missing, no?

pallets/pallet-asset-swap/src/xcm/trade.rs
```diff
@@ -109,56 +155,147 @@ mod swap_pair_remote_asset {
...
+	// the full amount on the remote location.
+	impl<T, WeightToFee, FeeDestinationAccount> Drop
+		for UsingComponentsForSwapPairRemoteAsset<T, WeightToFee, FeeDestinationAccount>
+	where
+		T: Config,
+		FeeDestinationAccount: Get<T::AccountId>,
+	{
+		fn drop(&mut self) {
+			log::trace!(target: LOG_TARGET, "Drop with remaining {:?}", (self.consumed_xcm_hash, self.remaining_fungible_balance, self.remaining_weight, &self.swap_pair));
+			match (self.remaining_fungible_balance, &self.swap_pair) {
```
Ad96el, line 268: Not a big fan of this match statement. I think a if is better suited

pallets/pallet-asset-swap/src/xcm/trade.rs
```diff
@@ -109,56 +155,147 @@ mod swap_pair_remote_asset {
...
+
+					// No error should ever be thrown from inside this block.
+					SwapPair::<T>::mutate(|entry| {
+						let Some(entry) = entry.as_mut() else {
+							log::error!(target: LOG_TARGET, "Stored swap pair should not be None but it is.");
+							return;
+						};
+						entry.remote_asset_balance = entry
+							.remote_asset_balance
+							.saturating_add(self.remaining_fungible_balance);
```
Ad96el, line 294: Maybe it is worth to have an error message, if we reached the numeric bounds. Not sure. It should never happen

Ad96el (commented):
LGTM, Just some understanding questions

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: Nope, that's not the correct assumption. If you see how the `impl_for_tuples` works, it sequentially calls `buy_weight` and then sequentially calls `refund_weight` and then `drop` in the end. So given the super-specific use cases in which these traders are to be used, we need to refund (and refund in drop) only if the very same asset has been withdrawn, which is why we need to have an `Option::Some` in there. Parity traders do something similar as well.

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: It's a `u128`. I also was assuming so, but apparently you cannot dereference a `u128`. And since the tests are passing, I assume we are changing the storage directly here 😄

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: Yeah I could add an error log. In the `drop` we can't fail, so it does not change much. But I'll add a `warn` log in there, perhaps.

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: Basically, we want to avoid the situation in which one trader is used to buy weight, and another one to refund it. We can control that by populating the state of the trader that was used to buy weight.

Ad96el: Ah, okay. I looked into the PolkadotSdk and couldn't find something like that. That's why I was bringing it up here. Maybe I wasn't looking attentive enough

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/661/commits/ddff4eec5507f603d528e55c0ca2e2f0274cac94

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: Not sure the added complexity compensates the benefits, but here it is: https://github.com/KILTprotocol/kilt-node/pull/661/commits/4066b94a9cc7bbeba5a283e99b2ca90707cfdbc7

pallets/pallet-asset-swap/src/xcm/trade.rs
ntn-x2: You need to check the [runtimes repo](https://github.com/polkadot-fellows/runtimes)

Ad96el: ah okay thank you

pallets/pallet-asset-swap/src/xcm/trade.rs
Ad96el: Yeah, maybe not. It should anyway never happen. Sorry for the extra work

ntn-x2: I'll revert this and then merge this PR 👍

PR #663, opened 2024-06-28
feat: support runtime-injected logic for swaps

ntn-x2:
WIP.

Based on top of https://github.com/KILTprotocol/kilt-node/pull/659.

## Checklist

- [x] Add hook to disallow swaps to a different remote beneficiary

Given the current XCM executor logic, specifically the appending of a `ClearOrigin` instruction right after a reserve-based transfer, it is not possible to make any judgments about the origin funds are coming from to our chain, hence it is not possible to apply any restrictions to those.

### DIFF
--- a/pallets/pallet-asset-swap/README.md
+++ b/pallets/pallet-asset-swap/README.md
@@ -2,8 +2,6 @@ WIP.
 
 # TODO
 
-* [REQUIRED FOR V2] Add hook to check the swap parameters (restricting where remote assets can be sent to).
-* [REQUIRED FOR V2] Add constraints about which beneficiary people can send their tokens to.
 * [REQUIRED FOR V2] Make sure XCM fee asset can only be used for local XCM fees when transferring the asset itself, if possible.
 * [REQUIRED FOR V2] Unit test the different XCM components.
 * [REQUIRED FOR V2] Add try-runtime support.
--- a/pallets/pallet-asset-swap/src/traits.rs
+++ b/pallets/pallet-asset-swap/src/traits.rs
@@ -37,6 +37,10 @@ where
 		to: &VersionedMultiLocation,
 		amount: LocalCurrencyBalanceOf<T>,
 	) -> Result<(), Self::Error>;
+
+	fn pre_remote_to_local_swap(to: &T::AccountId, amount: u128) -> Result<(), Self::Error>;
+
+	fn post_remote_to_local_swap(to: &T::AccountId, amount: u128) -> Result<(), Self::Error>;
 }
 
 impl<T> SwapHooks<T> for ()
@@ -58,6 +62,14 @@ where
 		_to: &VersionedMultiLocation,
 		_amount: LocalCurrencyBalanceOf<T>,
 	) -> Result<(), Self::Error> {
+		Ok(())
+	}
+
+	fn pre_remote_to_local_swap(_to: &<T>::AccountId, _amount: u128) -> Result<(), Self::Error> {
+		Ok(())
+	}
+
+	fn post_remote_to_local_swap(_to: &<T>::AccountId, _amount: u128) -> Result<(), Self::Error> {
 		Ok(())
 	}
 }
--- a/pallets/pallet-asset-swap/src/traits.rs
+++ b/pallets/pallet-asset-swap/src/traits.rs
@@ -1,0 +1,63 @@
+// KILT Blockchain – https://botlabs.org
+// Copyright (C) 2019-2024 BOTLabs GmbH
+
+// The KILT Blockchain is free software: you can redistribute it and/or modify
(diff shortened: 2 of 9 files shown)


pallets/pallet-asset-swap/README.md
```diff
@@ -2,9 +2,6 @@ WIP.
 
 # TODO
 
-* [REQUIRED FOR V1] Implement `Drop` for XCM transactors.
```
ntn-x2, line 5: Change cherry-picked from https://github.com/KILTprotocol/kilt-node/pull/661.

pallets/pallet-asset-swap/src/traits.rs
```diff
@@ -0,0 +1,75 @@
...
+
+	fn post_local_to_remote_swap(
+		from: &T::AccountId,
+		to: &VersionedMultiLocation,
+		amount: LocalCurrencyBalanceOf<T>,
+	) -> Result<(), Self::Error>;
+
+	fn pre_remote_to_local_swap(to: &T::AccountId, amount: u128) -> Result<(), Self::Error>;
+
+	fn post_remote_to_local_swap(to: &T::AccountId, amount: u128) -> Result<(), Self::Error>;
```
Ad96el, line 43: Why is the `from` not important here?

runtimes/peregrine/src/asset_swap/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+};
+
+use crate::Runtime;
+
+const LOG_TARGET: &str = "runtime::peregrine::asset-swap::RestrictTransfersToSameUser";
+
+/// Check requiring the beneficiary be a single `AccountId32` junction
+/// containing the same account ID as the account on this chain initiating the
+/// swap.
+pub struct RestrictSwapDestinationToSelf;
```
Ad96el, line 34: Do you think, this `SwapHook` implementation could be interesting for other teams?

pallets/pallet-asset-swap/src/traits.rs
Ad96el, line 43: I also wonder, do we really need that? Please correct me if I am wrong, but it seems to me those functions are not directly used in the pallet itself, only in some XCM related structs. If someone wants to do some pre- and post-remote to local checks, they can implement it by themselves or not?

Ad96el: It doesn't hurt if they would remain. It just feels a bit strange to me.

ntn-x2: Because unfortunately we don't have information about that. What we get from AH is "deposit these assets into this account", nothing else.

runtimes/peregrine/src/asset_swap/mod.rs
ntn-x2: We'll find out 😊 I wouldn't commit to have it into other crates for now, to be honest. We can always export it there if we see many projects will want to enforce the same requirement.

pallets/pallet-asset-swap/src/traits.rs
ntn-x2: I mean, sure, you can find the right XCM component to do what you want to do. Or simply call into the post hook after everything else happens. I think, as you also say, that it's nice to have pre and post for both local -> remote and remote -> local. There might be cases where you might want to do something else with this information, and unnecessarily restricting it would be a wasted opportunity, I think.

Ad96el: So we also don't know, from which parachain the message is coming?

ntn-x2: We only know that in the `Barrier`. After that, the info is cleared with the `ClearOrigin` instruction. This will change in the future, where the `ClearOrigin` instruction will most likely be replaced by a `DescendOrigin` instruction that the source chain adds to the message, so that we know the `MultiLocation` on the source chain that "authored" the XCM message, but right now the only place where the information about the source chain is available is in `Barrier`s, and that is unfortunately way too early to draw any kind of conclusions or call into any hooks.

Ad96el: Okay, but what would happen if a runtime has two instances of the pallet with a different `reserved_location`? The XCM config would use the first struct which is implementing the `swapHook` trait and maybe doing the wrong checks or not?

ntn-x2: The `SwapHook` trait takes also the pallet instance as a generic parameter, so if the first pallet has a different reserve location, the trait implementation would be able to tell the difference. It would then move on to the second implementation, which would then return the expected result.

Ad96el: I see

Ad96el (approved):

PR #666, opened 2024-07-01
chore: swap -> switch renaming

ntn-x2:
As per internal discussion.

Based on top of https://github.com/KILTprotocol/kilt-node/pull/664.


Ad96el (approved):
LGTM. Couldn't find any appearances of "swap"

PR #669, opened 2024-07-15
fix: xcm swap pallet 

Ad96el:
## fixes https://github.com/KILTprotocol/ticket/issues/3497

### DIFF
--- a/pallets/pallet-asset-switch/src/tests/mod.rs
+++ b/pallets/pallet-asset-switch/src/tests/mod.rs
@@ -36,6 +36,6 @@ fn assert_supply_invariant(
 ) {
 	// The pool must ALWAYS have enough local tokens to exchange for remote tokens
 	assert!(circulating_supply.into() <= Balances::usable_balance(pool_address) as u128);
-	// The amount of remote tokens controlled by the chain must always match the remote balance.
+	// The amount of calculated remote tokens controlled by the chain must always match the remote balance (including its ED).
 	assert_eq!(total_supply.into() - circulating_supply.into(), remote_balance.into());
 }
--- a/pallets/pallet-asset-switch/src/tests/mod.rs
+++ b/pallets/pallet-asset-switch/src/tests/mod.rs
@@ -28,10 +28,14 @@ mod set_switch_pair;
 mod switch;
 mod update_remote_fee;
 
-fn assert_total_supply_invariant(
+fn assert_supply_invariant(
 	total_supply: impl Into<u128>,
+	circulating_supply: impl Copy + Into<u128>,
 	remote_balance: impl Into<u128>,
 	pool_address: &AccountId32,
 ) {
-	assert!(total_supply.into() - remote_balance.into() <= Balances::usable_balance(pool_address) as u128);
+	// The pool must ALWAYS have enough local tokens to exchange for remote tokens
+	assert!(circulating_supply.into() <= Balances::usable_balance(pool_address) as u128);
+	// The amount of remote tokens controlled by the chain must always match the remote balance.
+	assert_eq!(total_supply.into() - circulating_supply.into(), remote_balance.into());
 }
--- a/pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
+++ b/pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
@@ -326,7 +326,7 @@ fn fails_on_invalid_supply_values() {
 			Error::<MockRuntime>::InvalidInput
 		);
 	});
-	// Circulating supply + minimum balance > total issuance
+	// Circulating supply - total issuance < min remote balance
 	ExtBuilder::default().build().execute_with(|| {
 		assert_noop!(
 			Pallet::<MockRuntime>::set_switch_pair(
--- a/pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
+++ b/pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
@@ -25,7 +25,7 @@ use crate::{
 		ExtBuilder, MockRuntime, NewSwitchPairInfo, System, ASSET_HUB_LOCATION, REMOTE_ERC20_ASSET_ID, XCM_ASSET_FEE,
 	},
 	switch::SwitchPairStatus,
-	tests::assert_total_supply_invariant,
+	tests::assert_supply_invariant,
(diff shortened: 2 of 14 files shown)


pallets/pallet-asset-switch/src/xcm/transact.rs
```diff
@@ -87,7 +87,7 @@ where
 			&switch_pair.pool_account,
 			&beneficiary,
 			fungible_amount_as_currency_balance,
-			Preservation::Expendable,
+			Preservation::Preserve,
```
ntn-x2, line 90: Does this now mean that the account must have ED + the amount of remote tokens? Or is that the purpose of `Preservation::Preserve`? What happens if we bring the balance to `ED + 1` and try to remove `2`? Does it go to `ED - 1`? Can it go all the way to `0`? If so, what happens then?

Ad96el: This means the account needs at least the ED. If the balance of an account is getting below the ED by the transfer, the `LocalCurrency::transfer` will fail. I think this is exactly what we want, no?

ntn-x2: Then we need to change the requirement for the pool account to be ED + circulating supply, which is not the case now.

ntn-x2: So there's no way to not have an ED requirement and still avoid the account from being dusted?

Ad96el: https://support.polkadot.network/support/solutions/articles/65000168651-what-is-the-existential-deposit-

Ad96el: I think, this is exactly what we need here

ntn-x2: Yeah but I'm sure the balances pallet has a way to mark accounts as "special", no?

ntn-x2:
Also, we need some more unit tests for the dust case.

ntn-x2:
@Ad96el please check the changes starting from commit https://github.com/KILTprotocol/kilt-node/pull/669/commits/4f3e41d8d490726396927ebebf58c91facc035ee.

pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
```diff
@@ -44,6 +44,7 @@ fn successful() {
 				Box::new(XCM_ASSET_FEE.into()),
 				u64::MAX as u128,
 				1_000,
+				0,
```
Ad96el, line 47: Is the zero here intendant? I expected to see a one here.

pallets/pallet-asset-switch/src/tests/switch.rs
```diff
@@ -320,12 +327,37 @@ fn fails_on_not_enough_user_local_balance() {
...
+		.execute_with(|| {
+			assert_noop!(
+				Pallet::<MockRuntime>::switch(
+					RawOrigin::Signed(user.clone()).into(),
+					100_000,
+					Box::new(ASSET_HUB_LOCATION.into())
+				),
+				Error::<MockRuntime>::UserSwitchBalance
+			);
+		});
```
Ad96el, line 348: Why is this test case failing?

pallets/pallet-asset-switch/src/tests/mod.rs
```diff
@@ -28,10 +28,14 @@ mod set_switch_pair;
...
+	circulating_supply: impl Copy + Into<u128>,
 	remote_balance: impl Into<u128>,
 	pool_address: &AccountId32,
 ) {
-	assert!(total_supply.into() - remote_balance.into() <= Balances::usable_balance(pool_address) as u128);
+	// The pool must ALWAYS have enough local tokens to exchange for remote tokens
+	assert!(circulating_supply.into() <= Balances::usable_balance(pool_address) as u128);
+	// The amount of calculated remote tokens controlled by the chain must always match the remote balance (including its ED).
+	assert_eq!(total_supply.into() - circulating_supply.into(), remote_balance.into());
 }
```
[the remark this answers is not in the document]
Ad96el: Or maybe better, the struct itself should implement it

Ad96el (commented):
LGTM 

1 question for clarification and 1 suggestion

pallets/pallet-asset-switch/src/tests/mod.rs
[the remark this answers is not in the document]
ntn-x2: Let me know if this comment is addressed in the new PR that I opened.

pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
ntn-x2: Why?

pallets/pallet-asset-switch/src/tests/switch.rs
ntn-x2: Because the user has 1 token that is in a hold state, hence they cannot perform the switch. Same for the case below it, which uses reserves.

pallets/pallet-asset-switch/src/tests/mod.rs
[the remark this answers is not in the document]
Ad96el: I will resolve it here, since it is already covered in the try-runtime PR

pallets/pallet-asset-switch/src/tests/set_switch_pair.rs
Ad96el: Ah the 1 from this [line](https://github.com/KILTprotocol/kilt-node/pull/669/files/36a43c521ec976baffe1f8a5280cf14a7a81e148#diff-dd0d7d477c7a0ce1248f0b2fa1ada994993b2bc2cc6b45b2979876c1b02420cdR37) is for the ED?

ntn-x2: Yes, we give `1_000` + the local ED to the pool account. While in this line you commented, `0` is the ED at destination.

Ad96el: I see

Ad96el:
I can not approve it, because this is my PR

PR #671, opened 2024-07-16
bench: add switch pallet benchmarks

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3511.

Adds benchmars to the asset-switch pallet. The benchmark logic could definitely be improved and be made more generic, but I think for the time being this is a good compromise. If we see that other projects are struggling to run the benchmarks with their runtime, we can think about how to improve them.


pallets/pallet-asset-switch/src/benchmarking.rs
```diff
@@ -0,0 +1,236 @@
...
+	fn setup() -> Option<BenchmarkInfo>;
+}
+
+impl BenchmarkHelper for () {
+	fn setup() -> Option<BenchmarkInfo> {
+		None
+	}
+}
+
+#[frame_benchmarking::v2::instance_benchmarks(where LocalCurrencyBalanceOf<T, I>: Into<u128>)]
```
Ad96el, line 41: Maybe import it

pallets/pallet-asset-switch/src/lib.rs
```diff
@@ -489,11 +500,19 @@ pub mod pallet {
...
+			cfg_if::cfg_if! {
+				if #[cfg(not(feature = "runtime-benchmarks"))] {
+					T::XcmRouter::deliver(xcm_ticket.0).map_err(|e| {
+						log::info!("Failed to deliver ticket with error {:?}", e);
+						DispatchError::from(Error::<T, I>::Xcm)
+					})?;
+				} else {
+					log::trace!(target: LOG_TARGET, "Running benchmarks. Message will not be delivered to destination.");
+					drop(xcm_ticket);
+				}
```
Ad96el, line 514: I think we can improve that. 
Is it not possible to mock a channel with `ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests`? A reachable destination can be configured in `BenchmarkHelper`. The destination is then set in the storage map of the pallet. 

Therefore, we don't need a special benchmark logic in the pallet and the calculated weights are more accurate

pallets/pallet-asset-switch/src/benchmarking.rs
Ad96el, line 41: a bit picky but I would prob import it

pallets/pallet-asset-switch/src/benchmarking.rs
```diff
@@ -0,0 +1,236 @@
...
+	};
+
+	const RESERVE_LOCATION: MultiLocation = MultiLocation {
+		parents: 1,
+		interior: Junctions::X1(Junction::Parachain(1_000)),
+	};
+	const REMOTE_ASSET_ID: AssetId = AssetId::Concrete(RESERVE_LOCATION);
+	const REMOTE_FEE: MultiAsset = MultiAsset {
+		id: REMOTE_ASSET_ID,
+		fun: Fungibility::Fungible(100_000),
```
Ad96el, line 66: Not sure, but I would probably prefer to have those const values in the default implementation of the `BenchmarkHelper`

pallets/pallet-asset-switch/src/tests/force_set_switch_pair.rs
```diff
@@ -59,7 +59,11 @@ fn successful() {
 				status: SwitchPairStatus::Paused,
 			};
 			assert_eq!(switch_pair, Some(expected_switch_pair.clone()));
-			assert_total_supply_invariant(u64::MAX, expected_switch_pair.remote_asset_balance, &pool_account_address);
+			assert_total_supply_invariant(
```
Ad96el, line 62: I assume, these are only fmt changes, right?

runtimes/peregrine/src/asset_switch/mod.rs
```diff
@@ -42,19 +41,30 @@ impl SwitchHooks<Runtime, KiltToEKiltSwitchPallet> for RestrictswitchDestination
...
+		cfg_if::cfg_if! {
+			if #[cfg(feature = "runtime-benchmarks")] {
+				// Clippy complaints the variable is not used with this feature on, otherwise.
+				let _ = is_beneficiary_self;
+				Ok(())
+			} else {
+				frame_support::ensure!(is_beneficiary_self, Error::NotToSelf);
+				Ok(())
+			}
+		}
```
Ad96el, line 67: Why is this needed? Can we not specify, in the benchmarks, that the receiver is the sender?

Ad96el (commented):
LGTM. 2 or 3 improve suggestions

pallets/pallet-asset-switch/src/benchmarking.rs
ntn-x2: But the purpose of these defaults is if the benchmark helper does not want to have anything to do with the setup, which would return `None`. In that case we assume "anything will do", and we set a default ourselves. Alternatively we could force the helper to return a value, but then we can't use `()` without attaching some more assumptions to it.

pallets/pallet-asset-switch/src/tests/force_set_switch_pair.rs
ntn-x2: Yeah but they're gone in the `try-runtime` PR: https://github.com/KILTprotocol/kilt-node/pull/674.

runtimes/peregrine/src/asset_switch/mod.rs
ntn-x2: You mean using the same account in the benchmarks to avoid doing this here?

Ad96el: yes

pallets/pallet-asset-switch/src/benchmarking.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/671/commits/c6f0b5e8d70c1ded508b7830f924a8f6fe5d9fb7

runtimes/peregrine/src/asset_switch/mod.rs
ntn-x2: I changed the logic, and allow the implementor to pass more info, none of them required. https://github.com/KILTprotocol/kilt-node/pull/671/commits/b9d28e5376ad30f1aae95f9b3992ad45572bc298

I like it much more now, thanks very much for the suggestions.

pallets/pallet-asset-switch/src/benchmarking.rs
ntn-x2: See comment below: https://github.com/KILTprotocol/kilt-node/pull/671#discussion_r1684460262

pallets/pallet-asset-switch/src/lib.rs
ntn-x2: Benchmarks seem to pass after the refactoring, thanks a lot! https://github.com/KILTprotocol/kilt-node/pull/671/commits/2ae571ee40d522c2ed1f05962ab85f394bc88ab0

runtimes/peregrine/src/asset_switch/mod.rs
Ad96el: Maybe it is overengineered, but what do you think of using the [derive_builder](https://docs.rs/derive_builder/latest/derive_builder/)?

ntn-x2: Mhh yeah but in the benchmarks we still want to convert any `None` into an actual value, so we still need a struct that has `Option`s, and one that does not. The builder pattern would be applied to the `OptionalBenchmarkInfo` struct, but we still internally need to convert that into a `BenchmarkInfo` one. Not sure it would actually bring that much value, given that this is info the trait implementer should return. If they need or wish, they could have a builder on one of their type, and then simply implement `Into<OptionalBenchmarkInfo>` for `MyType`.

ntn-x2: I personally think the trait should be as agnostic as possible over the details of its implementer. So if they want to use a builder pattern, they're free to do so. All the trait should care, is the return type of the implementation.

Ad96el: I see okay

Ad96el (approved):
Very nice!

PR #672, opened 2024-07-17
chore: Add stake.plus bootnodes

senseless:
Add stake.plus bootnodes


ntn-x2:
CC @ggera

ggera (approved):
LGTM!  Thanks & It would be nice if you enabled telemetry on your nodes so that users can see in real-time that they are online.

PR #676, opened 2024-07-25
chore: CI fixes

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3495, based on top of https://github.com/KILTprotocol/kilt-node/pull/675.

I had to remove the tests for the `RuntimeCall` size, since the newly introduced `pallet_assets` has a size of 896 bytes, which is huge. I noticed that none of the chains where the pallet is deployed and uses `MultiLocation` as asset ID has this test anymore... Maybe that will change in the future again.
Also, the `try-runtime` tests fail the idempotency tests for the same `pallet_assets` pallet. This should be solved by the `VersionedMigration` struct, but in 1.1 (the version this PR is based on) it was still marked as experimental. We might be able to solve that once we rebase on top of 1.7.


runtimes/peregrine/src/lib.rs
```diff
@@ -1166,7 +1166,9 @@ pub type Executive = frame_executive::Executive<
 	Runtime,
 	// Executes pallet hooks in the order of definition in construct_runtime
 	AllPalletsWithSystem,
-	(),
+	// Unfortunately, this migration fails the idempotency tests, and the `VersionedMigration` struct is still marked
+	// as experimental as of now.
+	pallet_assets::migration::v1::MigrateToV1<Runtime>,
```
Ad96el, line 1171: Why do we need the migration at all? We have no storage entries. I think if we just set the storage version, we are good to go.

Ad96el (commented):
There should be a struct to init the pallet version. I do not think we need the migration

runtimes/peregrine/src/lib.rs
ntn-x2: How would you do that? There's no extrinsic, so it would have to be a runtime migration. What this migration does, is to simply set the new storage version, since there's no other storage elements to migrate. Or maybe I am not getting your point?

Ad96el: Yes, it would be a runtime migration to set up the storage version. We have this struct: https://github.com/KILTprotocol/kilt-node/blob/develop/runtimes/common/src/migrations.rs#L29 

Which can be utilized as: `pallet_assets::Pallet::<T>::current_storage_version().put::<pallet_assets::Pallet<T>>();`

ntn-x2: I mean, the provided migration does exactly the same. Why would we need a new custom migration just for that?

ntn-x2: We might get green CI flag for try-runtime tho. I will use this migration on Monday before merging, you convinced me.

Ad96el: Yes okay. I also played a bit around, and I got the try-runtime tests passing by initializing the storage.

If you face this error: 

`[2024-07-26T14:19:04Z ERROR try-runtime::cli] failed to generate compact proof: TrieError(InvalidStateRoot(0x64c4b3358b6f671d81f8421100f5b73d1b1cbf931fc986ea8a0178db60cd7a0a))`

Update the try-runtime cli. There was a bug https://github.com/paritytech/try-runtime-cli/issues/76

ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/676/commits/d2fcad5405d911cc9e024c7c8a8aa7c49a078807

Ad96el (approved):

PR #684, opened 2024-08-01
docs: Add GitHub docs autogen

ChrisChinchilla:
Hacky way to cleanup disk space yet it works
Deployed at [this link ](https://kiltprotocol.github.io/kilt-node/did/index.html) (Custom link possible).


.github/workflows/docs-publish.yml
```diff
@@ -0,0 +1,40 @@
+name: Build and Deploy Chain docs
+
+on:
+  push:
+    branches:
+      - develop
```
ntn-x2, line 6: So now all versions will override each other since everything is deployed on the same branch? Could we perhaps name the branch differently depending on what we are pushing? Or just remove the `master` deployment and only have `develop` and the latest release, which are also the only two docs we would care about.

ggera: Removed `master`

ggera: Removed `master`

ntn-x2: Should we still name the deployment branches different between `develop` and releases?

ggera: I dynamically generate the deployment branch `gh-pages-${{ github.ref_name }}`, I hope that is not an issue for the linking @ChrisChinchilla

ntn-x2 (approved):

.github/workflows/docs-publish.yml
ChrisChinchilla: @ggera Dunno, I guess we'll see

PR #690, opened 2024-08-13
fix: asset switch runtime API returns a `Result`

ntn-x2:
As per the code review suggestion.


runtimes/peregrine/src/asset_switch/runtime_api.rs
Ad96el, line 1: Probably rename this file to runtime_api_error.

Ad96el, line 1: Not sure if it should live here. I guess spiritnet should use the same error type. Maybe move it to runtime/common/errors

ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/690/commits/1a5a80a908456c602e4f63cc0736b129a5357663

PR #697, opened 2024-08-19
fix: chainspec generation for DIP templates

ntn-x2:
Generating plain chainspecs worked, but adding `--raw` (and hence calling into the runtime) was failing.


Ad96el (approved):

ntn-x2:
Clippy and checks pass, so I'll merge.

PR #698, opened 2024-08-20
ci: Migrate Gitlab tests to GitHub runner

ggera:
## fixes KILTProtocol/ticket#3345
Build docker image only when tests on gh passed
## Checklist:

    * Link the PR/Ticket here


.github/workflows/check-clippy-all-features.yml
```diff
@@ -0,0 +1,58 @@
+name: Check Clippy all-features
+
+on:
+  pull_request:
+
+jobs:
+  get-head-commit-msg:
```
ntn-x2, line 7: I forgot to mention this when reviewing the other PR. It's not nice that this pops up as its own job in the Actions view. Can we merge it with the actual step below, and bail out early if the condition is not met? The name of the job would remain the "main" one.

.github/workflows/check-clippy-all-features.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Get Head Commit Message
+        id: step-head-commit
+        run: echo "headCommitMsg=$(git show -s --format=%s)" >> "$GITHUB_OUTPUT"
+
+  check-clippy-all-features:
```
ntn-x2, line 20: We can do a matrix strategy as in https://github.com/kilt-jam/jam-client/blob/master/.github/workflows/clippy.yaml.

.github/workflows/check-clippy-all-features.yml
```diff
@@ -0,0 +1,58 @@
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: $ ${{ github.job }}-${{ github.ref_name }}-clippy-all-features-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            $ ${{ github.job }}-${{ github.ref_name }}-clippy-all-features-
+
+      - name: Set up Docker Buildx
```
ntn-x2, line 45: Why do we need this? Doesn't GH support the `container` instruction?

.gitlab-ci.yml
```diff
@@ -1,27 +1,31 @@
-default:
-  interruptible: true
-
-.check_skip_rust:
+build-images:
```
ntn-x2, line 1: Can we move WASM building and artifacts as well to GL? SO we have all checks and tests on GH, and everything we publish we do via GL, instead of having to jump between the two.

.github/workflows/build-images.yml
ntn-x2, line 1: Rather `trigger-build-images-gitlab` or something to make it clear.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,22 @@
...
+  trigger_gitlab:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger GitLab CI
+        env:
+          GITLAB_TRIGGER_TOKEN: ${{ secrets.GITLAB_TRIGGER_TOKEN }}
+          GITLAB_PROJECT_ID: ${{ secrets.GITLAB_PROJECT_ID }}
+        run: |
+          curl -X POST \
```
ntn-x2, line 19: Very nice!

ntn-x2 (changes requested):
I have to say I like this structure much more. Few points:

1. The get-head-commit can be a simple step and not a full job, as it kinda pollutes the space for the actions that are actually meaningful. And since we can't re-use it anyway, I would suggest to make it a simple step for all jobs that need it, and reference it as a step output instead.
2. Already mentioned in a comment, but let's aim to have checks, tests and anything that does not result in a build on GH, and everything else on GL. These include Docker images, WASM blobs, and perhaps even the docs, if possible, since we can check them on GH and then deploy them from GL. If deploying the docs from GL is too complicated, I am willing to accept that we deploy them from GH, althought I wouldn't be super happy about it.
3. Not sure we need the Dockerx stuff in GH anymore, since we can specify a container to run the action within, and environment variables as well as part of the step or job or even workflow definition.

I will take a more thorough review once these are addressed, but as I said in the beginning, this looks already much better compared to what we now have, which is already much much better than what we had last week 🔥

ntn-x2 (changes requested):
I have to say I like this structure much more. Few points:

1. The get-head-commit can be a simple step and not a full job, as it kinda pollutes the space for the actions that are actually meaningful. And since we can't re-use it anyway, I would suggest to make it a simple step for all jobs that need it, and reference it as a step output instead.
2. Already mentioned in a comment, but let's aim to have checks, tests and anything that does not result in a build on GH, and everything else on GL. These include Docker images, WASM blobs, and perhaps even the docs, if possible, since we can check them on GH and then deploy them from GL. If deploying the docs from GL is too complicated, I am willing to accept that we deploy them from GH, althought I wouldn't be super happy about it.
3. Not sure we need the Dockerx stuff in GH anymore, since we can specify a container to run the action within, and environment variables as well as part of the step or job or even workflow definition.

I will take a more thorough review once these are addressed, but as I said in the beginning, this looks already much better compared to what we now have, which is already much much better than what we had last week 🔥

.github/workflows/check-clippy-all-features.yml
ggera: Condition will be mentioned in each step yet yes this looks better

.gitlab-ci.yml
ggera: done, only docs publishing will stay on gh

.gitlab-ci.yml
```diff
@@ -1,27 +1,31 @@
...
+  timeout: 2 hours
+  image:
+    name: kiltprotocol/kilt-ci:2.7.31
+    entrypoint: [""]
+  stage: build
+  only:
```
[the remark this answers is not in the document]
ggera: Yup, I modified the tigger to run only when workflows passwed on specific branches 
```  workflow_run:
    workflows: ["Clippy-checks", "Cargo-tests", "Try-runtime", "Integration-tests"]
    branches:
      - develop
      - master
      - 'refs/tags/[0-9]+.[0-9]+.[0-9]+*'```

.github/workflows/check-clippy-all-features.yml
ggera: nice, applied it in tests as well

PR #700, opened 2024-08-23
docs: remove branch auto name for now as it breaks build

ChrisChinchilla:
We attempted to create a different github pages branch for every version, but that doesn't work, so to unblock progress for now, this PR switches back to default GitHub way and we will find a better solution soon.


ggera (approved):
Considering https://github.com/KILTprotocol/ticket/issues/3583

PR #702, opened 2024-08-27
chore: fix benchmarking

Ad96el:
# Fixes [#3591](https://github.com/KILTprotocol/ticket/issues/3591)

This PR fixes the benchmarking logic. Instead of piping the output into a specific file, the default value is used, which is the pallet name itself. This change is necessary for pallets with multiple instances in the runtime.

While running the benchmarks, some extrinsics failed due to insufficient funds of the submitter. I'm not sure of the underlying issue, but the tests are still passing. Simply increasing the funds should resolve it.


Ad96el:
Noo, this pr is unrelated to the actions. @ggera can you maybe support?

ChrisChinchilla:
@Ad96el He already is

ntn-x2 (approved):

PR #708, opened 2024-08-30
fix: gitlab release pipeline

ntn-x2:
The Gitlab project ID did not need to be a secret, so I made it a repo variable. Everything else seems to work, at least the Pipeline is properly triggered with the latest HEAD.


ggera (approved):
Awesome :rocket:

PR #723, opened 2024-09-09
chore: post 1.14 release

ntn-x2:
* Remove executed runtime migrations
* Remove DMP pallet + migration (makes also `try-runtime` still fail since pallet version is `2` but storage version after cleanup is `0`)


integration-tests/chopsticks/src/network/polkadot.ts
```diff
@@ -27,11 +27,7 @@ export function removeDisputesAndMessageQueues() {
 		ParasDisputes: {
 			// those can makes block building super slow
 			$removePrefix: ['disputes'],
-		},
-		Dmp: {
```
ntn-x2, line 31: @Ad96el maybe this needs to be replaced by some other storage entry in the XCMP pallet?

Ad96el:
The metadata for Opendid, Dena dive binary, and the attester service needs an update. @abdulmth

integration-tests/chopsticks/src/network/polkadot.ts
Ad96el: Not sure. The pallet is still available on the relay chain, and I also checked the implementation in chopsticks. They still have this replacement in their testing [repository](https://github.com/AcalaNetwork/e2e-tests/blob/master/networks/polkadot.ts). I would probably leave it like that.

runtimes/peregrine/src/lib.rs
```diff
@@ -1201,11 +1200,6 @@ pub type Executive = frame_executive::Executive<
 	Runtime,
 	// Executes pallet hooks in the order of definition in construct_runtime
 	AllPalletsWithSystem,
```
Ad96el, line 1202: I think we have to add a migration to remove the pallet, right?

integration-tests/chopsticks/src/network/polkadot.ts
ntn-x2: Yes but this pallet won't be in the runtime anymore. It does not make sense to have it in here. Either an alternative storage is used, or this makes no sense to me. I think they probably just forgot.

runtimes/peregrine/src/lib.rs
ntn-x2: Good catch: https://github.com/KILTprotocol/kilt-node/pull/723/commits/8057f3be0689bfe14eb9ea8eb781b64ef25f098f.

integration-tests/chopsticks/src/network/polkadot.ts
ntn-x2: Sorry didn't see this is relaychain stuff. I added them back: https://github.com/KILTprotocol/kilt-node/pull/723/commits/2d5bfb3b76912c61641286866564a8add21ca779

Ad96el (approved):
LGTM

PR #725, opened 2024-09-11
fix: more accurate weights for staking heavy operations

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3564.

## How to test

Cd into `chopsticks`, then `yarn install` and `yarn spawn`. Try to call `execute_scheduled_reward_change` with a regular account and `set_inflation` with sudo.


chopsticks/config.yaml
```diff
@@ -0,0 +1,17 @@
+endpoint:
+    # We fetch relevant storage elements from Spiritnet even tho we are running Peregrine (with sudo).
+    - wss://spiritnet.kilt.io
+db: db/peregrine.sqlite
+runtime-log-level: 5
+port: 50002
+wasm-override: /home/antonio/Developer/kilt-node/target/debug/wbuild/peregrine-runtime/peregrine_runtime.wasm
```
Ad96el, line 7: Can we maybe use the relative path?

Ad96el: Or is this file only for testing, and we delete it after?

chopsticks/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+import-storage:
+    System:
+        Account:
+            - - - 4seWojfEHrk5YKPahdErazQ3CWEHZYi6NV4gKz5AaejWbRPJ
```
Ad96el, line 12: Is this Alice?

Ad96el (commented):
LGTM, one minor comment for chopsticks

Ad96el:
I tested the `execute_scheduled_reward_change` in the chopsticks env

chopsticks/config.yaml
ntn-x2: Yes, I will delete these once you are able to verify the correctness of the solution.

chopsticks/config.yaml
ntn-x2: Was it? I think it's Alice but with Ed25519.

Ad96el (approved):
LGTM

PR #729, opened 2024-09-16
chore: remove experimental feature

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3593.

Removing the feature everywhere and using the pallet implementation without the feature is equivalent to the current solution.

What the aura pallet does without the feature enabled is taking the `MinimumPeriod` as returned by the timestamp pallet, and multiply that by two: see code [here](https://github.com/paritytech/polkadot-sdk/blob/09df373db9cd5dfed82c5cdb0736d417d54249e6/substrate/frame/aura/src/lib.rs#L252).

Everywhere, we set `SlotDuration` to be exactly 2x `MinimumPeriod` (which is 6s), totalling 12s. So basically the net result is that `SlotDuration` is set to `12s` everywhere.

Current values of `MinimuPeriod` are:
* DIP consumer template: https://github.com/KILTprotocol/kilt-node/blob/d9552b70160465eaedd7263db450aa70f49667df/dip-template/runtimes/dip-consumer/src/lib.rs#L275
* DIP provider template: https://github.com/KILTprotocol/kilt-node/blob/d9552b70160465eaedd7263db450aa70f49667df/dip-template/runtimes/dip-provider/src/lib.rs#L276
* Kestrel (minimum period 0,5 s, block time 1s): https://github.com/KILTprotocol/kilt-node/blob/d9552b70160465eaedd7263db450aa70f49667df/runtimes/kestrel/src/lib.rs#L230
* Peregrine: https://github.com/KILTprotocol/kilt-node/blob/d9552b70160465eaedd7263db450aa70f49667df/runtimes/peregrine/src/lib.rs#L178
* Spiritnet: https://github.com/KILTprotocol/kilt-node/blob/d9552b70160465eaedd7263db450aa70f49667df/runtimes/spiritnet/src/lib.rs#L177C11-L177C62

I invite the reviewer to double-check that.

### DIFF
--- a/dip-template/runtimes/dip-consumer/Cargo.toml
+++ b/dip-template/runtimes/dip-consumer/Cargo.toml
@@ -33,7 +33,7 @@ frame-executive                            = { workspace = true }
 frame-support                              = { workspace = true }
 frame-system                               = { workspace = true }
 frame-system-rpc-runtime-api               = { workspace = true }
-pallet-aura                                = { workspace = true, features = ["experimental"] }
+pallet-aura                                = { workspace = true }
 pallet-authorship                          = { workspace = true }
 pallet-balances                            = { workspace = true }
 pallet-session                             = { workspace = true }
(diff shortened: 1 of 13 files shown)


Ad96el (approved):
LGTM

PR #731, opened 2024-09-16
test: test runtime benchmarks

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3201.

Actual runtime benchmarks are blocked by https://github.com/paritytech/polkadot-sdk/issues/3555.


scripts/run_benches_for_runtime.sh
```diff
@@ -57,16 +60,28 @@ echo "[+] Running all runtime benchmarks for $runtime --chain=$chain"
 
 cargo build $standard_args
 
+if [ $profile == "dev" ]; then
+    target_folder="debug"
+	# We care about benchmark correctness, not accuracy.
+	steps=1
```
Ad96el, line 66: I think steps must be at least 2.

.github/workflows/check-code.yml
```diff
@@ -218,69 +314,14 @@ jobs:
...
-      - name: Run `try-runtime`
-        run: |
-          ./try-runtime \
-          --runtime=./target/release/wbuild/${{ matrix.runtime }}-runtime/${{ matrix.runtime }}_runtime.compact.compressed.wasm \
-          on-runtime-upgrade \
-            --disable-spec-version-check \
-            --checks=all \
-          live \
-            --uri=wss://${{ matrix.runtime }}.kilt.io
+            bash -c "bash scripts/run_benches_for_runtime.sh ${{ matrix.runtime }}-dev dev > /dev/null"
```
Ad96el, line 327: Not sure, if we would find errors with this approach. If one pallet is throwing an error, the script would continue with the next pallet, we needed to look up in the logs to find possible issues.

.github/workflows/check-code.yml
```diff
@@ -13,192 +13,288 @@ concurrency:
...
+  #         ./try-runtime \
+  #         --runtime=./target/release/wbuild/${{ matrix.runtime }}-runtime/${{ matrix.runtime }}_runtime.compact.compressed.wasm \
+  #         on-runtime-upgrade \
+  #           --disable-spec-version-check \
+  #           --checks=all \
+  #         live \
+  #           --uri=wss://${{ matrix.runtime }}.kilt.io
+
+  runtime-benchmarks:
+    name: Test benchmarks
```
Ad96el, line 289: I think it is not necessary to run the benchmarks on every PR. Only updating polkadot dependencies can cause issues in the benchmarking logic. To speed up the CI maybe we should trigger them, maybe manually? I have also concerns to automate that based on the PR title.

.github/workflows/check-code.yml
```diff
@@ -13,192 +13,288 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  get-commit-head:
-    name: Get HEAD commit message
+  # get-commit-head:
```
Ad96el, line 16: is it intended, to comment them all out?

.github/workflows/check-code.yml
ntn-x2: It was for testing. I changed it to run only for `push` events, so on `master`, `develop`, and for tags. I think it does not hurt to have it there. Any other solution would most likely end up being too complicated, unless you have a solution in mind.

.github/workflows/check-code.yml
ntn-x2: Nope. It was for testing. I reverted that.

.github/workflows/check-code.yml
ntn-x2: You are right, the script was considering only the result of the last benchmarked pallet. I changed that [here](https://github.com/KILTprotocol/kilt-node/pull/731/commits/792ec806bc69707956e6ccef1e064e696afb0ae5), and made sure that we bail out early as long as one benchmark fails. I tested it by changing the benchmark helper of the switch pallet to make it fail (which it did), and also to try to run the `sudo` pallet for peregrine after it, but it did not because the script exited on the first failure.

scripts/run_benches_for_runtime.sh
```diff
@@ -44,32 +47,42 @@ pallets=(
...
 
 cargo build $standard_args
 
+if [ $profile == "dev" ]; then
+    target_folder="debug"
+	# We care about benchmark correctness, not accuracy.
+	additional_args="--steps=2 --repeat=1 --default-pov-mode=ignored --no-verify"
+else
+    target_folder=$profile
+	additional_args="--header=\"HEADER-GPL\" --template=\".maintain/runtime-weight-template.hbs\" --output=\"./runtimes/${runtime}/src/weights/\""
```
Ad96el, line 67: steps=50 and repeat=20 are the default values right? Otherwise, they are missing here

scripts/run_benches_for_runtime.sh
```diff
@@ -44,32 +47,42 @@ pallets=(
...
+fi
+
 for pallet in "${pallets[@]}"; do
 	echo "Runtime: $runtime. Pallet: $pallet"
 	# shellcheck disable=SC2086
-	./target/release/kilt-parachain benchmark pallet \
-		--template=".maintain/runtime-weight-template.hbs" \
-		--header="HEADER-GPL" \
+	./target/$target_folder/kilt-parachain benchmark pallet \
 		--wasm-execution=compiled \
```
Ad96el, line 74: I think we can remove this flag

.github/workflows/check-code.yml
Ad96el: I see. I think we should maybe change that to `pull_request`?
According to the documentation, the pipeline will be triggered when a PR is created or updated. It would be a minor improvement, allowing someone to test things without using up CI resources. But we can also leave it like that. No strong opinion

scripts/run_benches_for_runtime.sh
ntn-x2: Yes, that's why I removed them.

.github/workflows/check-code.yml
ntn-x2: Exactly, we don't want to run in on every PR, right? Yet, we should be able to catch any breaking changes before we start the release process, that's why I thought having it run on pushes onto our two main branches is a good idea. Plus, if the benchmark command breaks, we should avoid push those Docker images out, that's why it's part of the same workflow.

scripts/run_benches_for_runtime.sh
ntn-x2: Cleaned both up since `--wasm-execution=compiled` is the default and the `--execution=wasm` deprecated. I adjusted also the pallet benchmarking command to do the same: https://github.com/KILTprotocol/kilt-node/pull/731/commits/d8a43d450837b99ae27072ecfc8b2662c1d36042.

.github/workflows/check-code.yml
Ad96el: I agree. Should be fine.

Ad96el (approved):

PR #734, opened 2024-09-20
fix: benchmark test CI

ntn-x2:
Ops.

### DIFF
--- a/.github/workflows/check-code.yml
+++ b/.github/workflows/check-code.yml
@@ -321,7 +321,7 @@ jobs:
       # Configured by the Docker image. We can't change this unless the image does it.
       CARGO_HOME: /usr/local/cargo
     needs: cargo-clippy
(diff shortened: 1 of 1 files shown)


.github/workflows/check-code.yml
```diff
@@ -332,13 +317,11 @@ jobs:
...
     name: Test runtime benchmarks
     runs-on: ubuntu-latest
-    container:
-      image: paritytech/ci-unified:bullseye-1.74.0
     env:
       # Configured by the Docker image. We can't change this unless the image does it.
       CARGO_HOME: /usr/local/cargo
     needs: cargo-clippy
-    if: ${{ github.event_name == 'push'}}
+    # if: ${{ github.event_name == 'push'}}
```
ntn-x2, line 324: Need to revert this once the step passes in this PR.

PR #735, opened 2024-09-20
feat: allow easy integration of the deposit storage for `MutateHold` requirements

ntn-x2:
WIP. Fixes https://github.com/KILTprotocol/ticket/issues/3607.


ntn-x2:
Unfortunately it is not possible, because of the strong limitations on the `RuntimeHoldReason` associated type of the `InspectHold` trait. The pallet deposit storage needs namespaces when holding some funds, and the trait does not make it possible to use the namespace type as the hold reason.

PR #736, opened 2024-09-23
[zombienet] replace chain-spec

pepoviola:
cc: @ntn-x2 
Replace relaychain chain-spec (polkadot-local), make `spiritnet` test working as expected.
Thx!


ntn-x2:
Found the culprit! Thanks a lot @pepoviola!

PR #714, opened 2024-09-03
chore: add zombienet config files for quick spin-ups

ntn-x2:
Quick way of spinning up a parachain instance using actual node software and consensus rather than mocking stuff with Chopsticks.

PR #740, opened 2024-09-25
chore: add extensible chopsticks setup

ntn-x2:
To spin up Chopsticks-based networks using actual data.

It's based on top of https://github.com/KILTprotocol/kilt-node/pull/714.

For more info about how to use this new utility, check the `chopsticks/README.md` file.


ntn-x2:
@Ad96el as for the [other](https://github.com/KILTprotocol/kilt-node/pull/714) PR, please provide a review when you have time. I will go ahead and merge this one now as I need it.

Ad96el (commented):
Overall, LGTM.

To be honest, I envision a different setup. Instead of introducing another TypeScript subproject into the kilt-node, I would prefer to see these functionalities combined with the Chopstick setup used for the e2e tests. Many of the functionalities used in the e2e tests can also define the network. A minimal CLI can be introduced, similar to what's done here, to spin up a network. This means that instead of using YAML files to describe the network, the configuration would be defined in TypeScript, which should fit well with the team's current skill set.

I also think using TypeScript is more flexible than YAML files. Basic functionalities, like overwriting the storage of an already running network, are not possible with YAML.

Ad96el:
Yes, for now it's a great intermediate solution. I created a [ticket](https://github.com/orgs/KILTprotocol/projects/24/views/1?pane=issue&itemId=82613542&issue=KILTprotocol%7Cticket%7C3634) to track that.

PR #742, opened 2024-09-28
fix: spiritnet config for the switch pallet

ntn-x2:


ntn-x2:
@Ad96el for your info. I will merge and cherry-pick on the release branch for a new runtime.

PR #746, opened 2024-10-08
feat: runtime API for switch XCM

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3628.

* Moves most of the versioned -> v4 logic into a `TryFrom` implementation
* Moves the XCM message creation into its own function
* Re-use the function in a new runtime API method for the switch pallet


pallets/pallet-asset-switch/src/lib.rs
```diff
@@ -838,6 +756,87 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
+		// 			contain either only the XCM fee token in case of successful transfer, or the
+		//	 		XCM fee token + the amount of funds supposed to be transferred.
+		// 		3.2 Deposit the un-transferred asset (only if the transfer failed) back into
+		// our account.
+		//		3.3 Refund any surplus weight.
+		//		3.4 Deposit the remaining XCM fee assets in the user's account.
+		// 4. Withdraw the requested asset (this operation should be infallible since we
+		//    have full control of this balance)
+		// 5. Try to deposit the withdrawn asset into the user's account. This operation
+		//    could fail and the error is handled in the appendix.
```
Ad96el, line 816: Should the docu maybe be moved as a docstring to the beginning of the function?

pallets/pallet-asset-switch/src/lib.rs
```diff
@@ -479,33 +481,7 @@ pub mod pallet {
...
-			let destination_v4: Location = switch_pair.remote_reserve_location.clone().try_into().map_err(|e| {
-				log::error!(
-					target: LOG_TARGET,
-					"Failed to convert remote reserve location {:?} into v4 `Location` with error {:?}",
-					switch_pair.remote_reserve_location,
-					e
-				);
-				DispatchError::from(Error::<T, I>::Internal)
-			})?;
+			let switch_pair_v4 = SwitchPairInfoV4::<_, T, I>::try_from(switch_pair).map_err(DispatchError::from)?;
```
Ad96el, line 484: I like these changes, but I'm thinking, that maybe the xcm version should be defined in the config? We are currently forcing version 4. Other teams might use a different version. Making the pallet generic against the xcm version requires more changes, which might be not worth it in regard to the effort.

Ad96el (commented):
Two minor comments

pallets/pallet-asset-switch/src/lib.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/746/commits/53b5184c3f6c814334f53e104c5b894c640f0bc5

Ad96el (approved):

PR #753, opened 2024-10-10
[REVIEW PR] Release 1.14.5 -> 1.14.4

ntn-x2:
Changes compared to the last release for this PR.


pallets/attestation/src/tests/delete.rs
```diff
@@ -37,57 +37,9 @@ fn test_remove() {
...
-	ExtBuilder::default()
-		.with_balances(vec![(ACCOUNT_00, <Test as Config>::Deposit::get() * 100)])
-		.with_ctypes(vec![(attestation.ctype_hash, attester.clone())])
-		.with_attestations(vec![(claim_hash, attestation)])
-		.build_and_execute_with_sanity_tests(|| {
-			assert_ok!(Attestation::remove(
-				DoubleOrigin(ACCOUNT_00, attester.clone()).into(),
-				claim_hash,
-				None
+				authorization_info
```
Ad96el, line 40: Release 1.14.4 contains the broken attestation events?

Ad96el: Ah, wait 1.14.4 only contained fixes for the node right?

ntn-x2: This looks like a GitHub bug. The two files are exactly the same if you open them on each respective branch. They also don't show as diff in my VS Code. Uh, that makes the whole point of this review kinda worthless. Maybe I can try to adjust the history, let's see.

ntn-x2: I would suggest you look at the PR from VS Code, which seems to be better at showing actual diffs.

Ad96el (approved):

PR #761, opened 2024-10-18
feat: reset team and freeze assets

Ad96el:
## fixes 

This PR introduces the reset team and the ability to freeze currencies. In the creation step, each currency requires a team, where the issuer, admin, and freezer must be specified. Additionally, the tradability of each currency can be activated or deactivated.

In the latest Polkadot SDK version, the reset team functionality is already implemented in `pallet_assets`. For now, the trait has been copied into the bonded pallet crate. This will need to be updated once the trait becomes available in the Polkadot SDK.

### DIFF
--- a/pallets/pallet-bonded-coins/src/curves.rs
+++ b/pallets/pallet-bonded-coins/src/curves.rs
@@ -8,6 +8,26 @@ use substrate_fixed::{
 	types::I9F23,
 };
 
+pub trait ParameterValidation {
+	fn are_parameters_valid(&self) -> bool;
+}
+
+pub trait BondingFunction<F: FixedSigned + PartialOrd> {
+	/// Calculates the cost of the curve between two points.
+	///
+	/// # Parameters
+	/// - `low`: The lower bound of the range for which the cost is to be calculated.
+	/// - `high`: The upper bound of the range for which the cost is to be calculated.
+	///
+	/// # Returns
+	/// - `Ok(F)`: The calculated cost if the operation is successful.
+	/// - `Err(ArithmeticError)`: An error if the calculation fails due to arithmetic issues.
+	///
+	/// # Errors
+	/// This function will return an `ArithmeticError` if the calculation cannot be performed.
+	fn calculate_costs(&self, low: F, high: F) -> Result<F, ArithmeticError>;
+}
+
 /// An enumeration representing different types of bonding curves.
 ///
 /// This enum is generic over the type `F`, which represents the type of the coefficients used in the bonding functions.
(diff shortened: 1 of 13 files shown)


pallets/pallet-bonded-coins/src/curves.rs
```diff
@@ -8,6 +8,26 @@ use substrate_fixed::{
 	types::I9F23,
 };
 
+pub trait ParameterValidation {
+	fn are_parameters_valid(&self) -> bool;
```
ntn-x2, line 12: This somewhat goes against Rust idiomatic guidelines. Ideally you should not have an instance of `&self` if this check would return `false`. You should implement something like `TryFrom` in this case.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -222,9 +233,13 @@ pub mod pallet {
 			NextAssetId::<T>::set(next_asset_id);
 
 			// Should never fail.
-			let currency_ids =
-				BoundedVec::<FungiblesAssetIdOf<T>, T::MaxCurrencies>::try_from(currency_ids_vec.clone())
-					.map_err(|_| Error::<T>::CurrenciesNumber)?;
+			let currency_ids = BoundedVec::<FungiblesAssetIdOf<T>, T::MaxCurrencies>::try_from(
```
ntn-x2, line 236: Why not returning the bounded vec directly from `generate_sequential_asset_ids`?

ntn-x2 (commented):
Also here I ignored tests and focused on the few lines of codes that are not test-related.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -252,22 +267,25 @@ pub mod pallet {
...
+				T::Fungibles::reset_team(
+					asset_id.clone(),
+					pool_id.clone().into(),
+					entry.team.admin.clone(),
+					entry.team.issuer.clone(),
+					entry.team.freezer.clone(),
+				)?;
+
+				if !entry.tradable {
+					T::Fungibles::freeze_asset(pool_id.clone().into(), asset_id.clone())?;
```
rflechtner, line 279: unfortunately, this will lead to all mint/burn operations failing, as these all happen after an 'asset-is-live' check in the assets pallet. You will have to unfreeze the asset before any change to balances in the pallet. An alternative is placing locks on any amount minted I guess.

Ad96el:
Superseded by: https://github.com/KILTprotocol/kilt-node/pull/768. I addressed all comments in the new PR.

PR #762, opened 2024-10-18
chore: more lints

ntn-x2:
Add more lints to our codebase. I have taken all lints that are allowed by default, and made either either warnings or strong denies. **I have not added any lints from the [pedantic](https://rust-lang.github.io/rust-clippy/master/index.html?groups=pedantic) group, as those tend to be opinionated and I don't think should be enforced on a shared codebase.**

Unfortunately, there is an [open cargo issue](https://github.com/rust-lang/cargo/issues/8170) which makes it impossible to enable/disable lints based on the `#[cfg(test)]` or `#[cfg(feature = "...")]` features, hence I have had to enable lints for either the whole codebase (`[target.'cfg(all())']`), or for runtime-only code `[target.'cfg(target_arch = "wasm32")']`.

The changeset is huge but just because I tackled all the new lints generated, so reviewing the new lints (in case some are unwanted or in case I missed some) would be sufficient, since the CI has been updated to verify all lints are properly addressed.

Last but not least, I had to enable the `-Dwarnings` flag for ALL clippy invocations, so also on local machines, because it's not possible to pass custom `--cfg` either to rustc via cargo, e.g., `cargo clippy -- --cfg ci` and then have a `[target.'cfg(ci)']` entry in the config file. This might go away in future versions of cargo (we are currently on 1.74), but until then we either never fail on the CI, or we fail for everything also locally, and I think the latter is a better choice. This is also because rustc takes additional flags either from the `RUSTFLAGS` env variable or the `.cargo/config.toml` file. Hence specifying `RUSTFLAGS="-Dwarnings"` in the CI would not enable any additional lints as the env variable would take precedence over the config file, rendering them useless.


dip-template/nodes/dip-consumer/src/chain_spec.rs
```diff
@@ -16,6 +16,10 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
+// Triggered by the `ChainSpecGroup` derive macro used for the custom chainspec
+// extension.
+#![allow(clippy::derive_partial_eq_without_eq)]
```
Ad96el, line 21: the reason, why it is allowed, should maybe go into the allow. We can also enforce that with: https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes_without_reason

.cargo/config.toml
```diff
@@ -1,5 +1,58 @@
...
+  "-Wclippy::ref_patterns",
+  "-Wclippy::rest_pat_in_fully_bound_structs",
+  "-Wclippy::suspicious_operation_groupings",
+  "-Wclippy::type_repetition_in_bounds",
+  "-Wclippy::unnecessary_self_imports",
+  "-Wclippy::unnecessary_struct_initialization",
+  "-Wclippy::unneeded_field_pattern",
+  "-Wclippy::unused_peekable",
+  "-Wclippy::useless_let_if_seq",
+  "-Wclippy::wildcard_dependencies",
```
Ad96el, line 41: For a better testing, we can also add: https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_result_states. 

Then we are enforced to check which error is thrown. We can still maintain the old behavior by assert_eq(result, Ok(_))

.cargo/config.toml
```diff
@@ -1,5 +1,58 @@
...
+  "-Dwarnings",
+  "-Wclippy::as_underscore",
+  "-Wclippy::branches_sharing_code",
+  "-Wclippy::clear_with_drain",
+  "-Wclippy::clone_on_ref_ptr",
+  "-Wclippy::collection_is_never_read",
+  "-Wclippy::derive_partial_eq_without_eq",
+  "-Wclippy::else_if_without_else",
+  "-Wclippy::empty_drop",
+  "-Wclippy::empty_structs_with_brackets",
```
Ad96el, line 17: We should do the same for enums: [empty_enum_variants_with_brackets ](https://rust-lang.github.io/rust-clippy/master/index.html#empty_enum_variants_with_brackets)

.cargo/config.toml
```diff
@@ -1,5 +1,58 @@
-# Deployment runtime lints
-[target.'cfg(all(target_arch = "wasm32", not(test)))']
+# Unfortunately we cannot add any lints based on `cfg(test)` nor `cfg(feature = "feat")`: https://github.com/rust-lang/cargo/issues/8170.
+# Hence, we can only define lints that are applied codebase-wide (including tests), and lints that are applied only to runtime code.
+
+# Codebase-wide lints.
+[target.'cfg(all())']
+rustflags = [
```
Ad96el, line 6: No strong opinion on that but I thing, we have some single match statements: https://rust-lang.github.io/rust-clippy/master/index.html#single_match

Ad96el, line 6: If I remember correctly, we have agreed, that tests should always live in a tests module. We can enforce that with: https://rust-lang.github.io/rust-clippy/master/index.html#tests_outside_test_module

Ad96el (commented):
Some suggestions on additional lints. 
I like the changes and I think they are a good step to enforce us to write cleaner code. 
I mainly focused on the `config.toml` as you described. Just one question about the introduced Arc usage on the client side.

dip-template/nodes/dip-provider/src/service.rs
```diff
@@ -118,13 +118,13 @@ pub fn new_partial(
 		config.role.is_authority().into(),
 		config.prometheus_registry(),
 		task_manager.spawn_essential_handle(),
-		client.clone(),
+		Arc::clone(&client),
```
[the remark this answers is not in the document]
ntn-x2: I did not, and I will. If Clippy threw a warning, I am fairly confident it was an `Arc`, but I will double check.

.cargo/config.toml
ntn-x2: Only available in 1.77 onwards.

.cargo/config.toml
ntn-x2: I will double-check this, but I was stumbling upon some false positives. Let me check that again.

dip-template/nodes/dip-provider/src/service.rs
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/blob/f057961868d2d865a8d3ff4a5f977cc5e381472a/nodes/parachain/src/service.rs#L159

.cargo/config.toml
ntn-x2: Plus it's already commented in the config file, so it just needs to be uncommented once we upgrade the compiler version.

.cargo/config.toml
ntn-x2: I'm not sure this is a good idea. We would fail to catch if the enum has new variants, and we would simply fail. I think having explicit enumeration is almost always better.

dip-template/nodes/dip-provider/src/service.rs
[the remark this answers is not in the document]
Ad96el: Okay funny. client.clone() is then syntactically the same. Which lint is throwing the error?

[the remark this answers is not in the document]
ntn-x2: I think the code is expecting an `Arc<dyn _>` so a trait object. I could also fix it that way, but honestly it's client code that we will barely touch in the future.

.cargo/config.toml
Ad96el: Oh yes, I somehow missed that

dip-template/nodes/dip-provider/src/service.rs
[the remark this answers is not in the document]
Ad96el: Yes thats true. I think changing it back is not worth it. I was just wondering why the linter is complaining.

dip-template/nodes/dip-consumer/src/chain_spec.rs
ntn-x2: Unfortunately allow and expect reasons were still unstable in 1.74, because they were only stabilised in 1.81, hence we cannot use them here. I added two more lints that can be uncommented if and when we upgrade our toolchain to 1.81: https://github.com/KILTprotocol/kilt-node/pull/762/commits/7b2ef121459d012a3ccdb084ea4f5808d489cab4.

.cargo/config.toml
ntn-x2: I like this very much! https://github.com/KILTprotocol/kilt-node/pull/762/commits/a0e2f9b04f36224f880e194ff55cc1b574c956b8

.cargo/config.toml
ntn-x2: Ok I found out why. Most pallets use the old benchmarking syntax which mixes tests and non tests together, triggering this lint a lot. I will open a new PR that adds the new lint and migrates the old benchmarking logic to the new one, which should also solve the problem with the lints.

Ad96el: gl hf in the staking pallet

Ad96el (approved):

ntn-x2:
@rflechtner if you were to rebase on `develop`, you might be surprised by the amount of new lints that are generated. This is for the better 😄

.cargo/config.toml
ntn-x2: Example of running clippy with the new lint:

```
error: this function marked with #[test] is outside a #[cfg(test)] module
   --> pallets/attestation/src/benchmarking.rs:31:1
    |
31  | / benchmarks! {
32  | |     where_clause {
33  | |         where
34  | |         T: core::fmt::Debug,
...   |
203 | |     )
204 | | }
    | |_^
    |
    = note: move it to a testing module marked with #[cfg(test)]
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tests_outside_test_module
    = note: `-D clippy::tests-outside-test-module` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::tests_outside_test_module)]`
    = note: this error originates in the macro `$crate::impl_bench_name_tests` which comes from the expansion of the macro `benchmarks` (in Nightly builds, run with -Z macro-backtrace for more info)
```

ntn-x2: Slapping a `#[cfg(test)]` above `impl_benchmark_test_suite!`, gives the following:

```
error: no rules expected the token `cfg`
   --> pallets/attestation/src/benchmarking.rs:199:4
    |
199 |     #[cfg(test)]
    |       ^^^ no rules expected this token in macro call
    |
note: while trying to match `skip_meta`
   --> /home/antonio/.cargo/git/checkouts/polkadot-sdk-cff69157b985ed76/e046d3b/substrate/frame/benchmarking/src/v1.rs:321:5
    |
321 |         #[skip_meta]
```

ntn-x2: I guess you did not run it with `--all-features`?

PR #768, opened 2024-10-22
feat: create pool

Ad96el:
##  Fixes: [#3657](https://github.com/KILTprotocol/ticket/issues/3657)

This PR introduces the feature to create new bonded pools.

For each newly created currency, the metadata are updated, and the team is reset. The currently implemented trait will be removed once the Polkadot version provides it.

The generated pool_id is also used as a proxy account where the collateral is accumulated.

The deposit system may change in the future, but a basic implementation is provided for now.


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,8 +174,91 @@ pub mod pallet {
...
+					T::DepositPerCurrency::get()
+						.saturating_mul(currency_length.saturated_into())
+						.saturated_into(),
+				),
+			)?;
+
+			currencies
+				.iter()
+				.enumerate()
+				.try_for_each(|(idx, entry)| -> DispatchResult {
```
Ad96el, line 218: Move this closure to its own function. Benchmarking would be simpler.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,8 +174,91 @@ pub mod pallet {
...
+			let who = T::PoolCreateOrigin::ensure_origin(origin)?;
+
+			let currency_length = currencies.len();
+
+			ensure!(
+				(1..=(T::MaxCurrencies::get()).saturated_into()).contains(&currency_length),
+				Error::<T>::CurrenciesNumber
+			);
+
+			let checked_curve = curve.try_into().map_err(|_| Error::<T>::InvalidCoefficients)?;
```
Ad96el, line 196: This operation can still fail. Depending on the selected `CurveparameterType`, the int bits may not fit. The error from `try_into` is ignored. I think it doesn't really matter to know which parameter is wrong.

pallets/pallet-bonded-coins/src/traits.rs
```diff
@@ -0,0 +1,20 @@
+use frame_support::{dispatch::DispatchResult, traits::fungibles::Inspect};
+
+/// Copy from the Polkadot SDK. once we are at version 1.13.0, we can remove this.
+pub trait ResetTeam<AccountId>: Inspect<AccountId> {
```
ntn-x2, line 4: Why does the trait itself need to require `Inspect<AccountId>`?

pallets/pallet-bonded-coins/src/types.rs
```diff
@@ -98,3 +98,10 @@ pub struct TokenMeta<Balance, Symbol, Name> {
 	pub symbol: Symbol,
 	pub min_balance: Balance,
 }
+
+#[derive(Debug, Encode, Decode, Clone, PartialEq, TypeInfo)]
+pub struct Team<AccountId> {
```
ntn-x2, line 103:
```suggestion
pub struct PoolManagingTeam<AccountId> {
```

or something along those lines. Please use meaningful names for types and variables.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -151,13 +155,14 @@ pub mod pallet {
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
-		Todo,
+		PoolCreated(T::PoolId),
```
ntn-x2, line 158: Please use structs for events payload.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -151,13 +155,14 @@ pub mod pallet {
...
 	}
 
-	// Errors inform users that something went wrong.
 	#[pallet::error]
 	pub enum Error<T> {
-		Todo,
+		CurrenciesNumber,
```
ntn-x2, line 163: Can you perhaps use a more descriptive name for this error?

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,8 +174,91 @@ pub mod pallet {
...
+			currencies: BoundedVec<TokenMetaOf<T>, T::MaxCurrencies>,
+			state: PoolStatus<Locks>,
+			denomination: u8,
+			pool_manager: AccountIdOf<T>,
+			transferable: bool,
+			team: Team<AccountIdOf<T>>,
+		) -> DispatchResult {
+			let who = T::PoolCreateOrigin::ensure_origin(origin)?;
+
+			let currency_length = currencies.len();
```
ntn-x2, line 189: If you take a `BoundedVec` in input, it is already guaranteed to be the max length, so you don't need to perform these checks. Different is, if you take a regular `Vec` as input, which is not recommended anyway.

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: I would rename the error `InvalidInput`, since as far as this pallet is concerned, it does not know exactly what in the conversion failed: it just knows that it could not convert the input to the expected type.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,8 +174,91 @@ pub mod pallet {
 	impl<T: Config> Pallet<T> {
 		#[pallet::call_index(0)]
 		#[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
-		pub fn create_pool(_origin: OriginFor<T>) -> DispatchResult {
-			todo!()
+		pub fn create_pool(
+			origin: OriginFor<T>,
+			curve: CurveInput<CurveParameterInputOf<T>>,
+			currencies: BoundedVec<TokenMetaOf<T>, T::MaxCurrencies>,
+			state: PoolStatus<Locks>,
```
ntn-x2, line 181: Is it possible to create a pool with a `Destroying` state?

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,8 +174,91 @@ pub mod pallet {
...
+			denomination: u8,
+			pool_manager: AccountIdOf<T>,
+			transferable: bool,
+			team: Team<AccountIdOf<T>>,
```
ntn-x2, line 185: Honestly, this feels like quite a big extrinsic. Have you guys considered splitting this up into multiple, atomic, extrinsics, that can be batched together to achieve the same result? Especially setting the team, for instance. Once the pool is created with a manager, the next extrinsic could be called. I think it makes it easier. We've done this mistake once in the DID pallet and it's been biting us back to this day.

ntn-x2: The `transferable` property I understand if it cannot be changed after creation, but for the team I would imagine there could be need to update its members. Same for the pool state. I would suggest to take out of this extrinsic anything that could be updated in a different extrinsic.

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: Still, good luck benchmarking this 😃 Since as far as I understand it's not a requirement that all currencies have sequential IDs, have you considered exposing one more extrinsic `add_currency` which would do one iteration of this foor loop, and a `finalize_curve` or something like that which prevents new currency additions?

ntn-x2: The power of tx batching is real.

ntn-x2: Of course, as long as the `create_pool` slims down to just setting the basic curve parameters like denomination and admin team.

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: If not, then this should not be a parameter and the state enum should have a default state, kinda like we do for the switch pallet.

rflechtner: that's the way I had set it up but SOMEBODY though they could just mess with my beautiful beautiful code !!!!11111

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: yeah 100% agree and we had talked about this prior. there really is no reason to not set this as part of dedicated extrinsics

rflechtner: concerns team and pool manager, for transferable it's probably better we keep if for now

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Is it that difficult to benchmark? I would benchmark the closure separately and the transaction without any currencies. The final weight would be the length of the currencies * the weight from the closure + the base weight from the transaction. That should be pretty accurate.

Ad96el: Writing smaller atomic tx is ofc another option, but I would say it is not as complicated yet to do that

ntn-x2: The way you would benchmark it is actually just giving it input currencies from 1 to the max limit, that's it. The implementation of the `Fungibles` for the pallet should be part of the benchmarked logic, so I don't think you need to split it up and benchmark the closure separately.

ntn-x2: So yeah, for input vecs probably it's ok to benchmark it like that, but for other extrinsics which iterate over stored vecs or other collections, it's better to expose them via atomic transactions that operate directly on a single item.

ntn-x2: So yeah, benchmarking this might actually not be that hard, although I would still suggest to take out of the extrinsic signature stuff that could be done in its own extrinsic, like setting the team, for instance. Then, I would also not split this up at all, but simply benchmark the whole extrinsic based on the length of the input vec, nothing else.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Fair point: [b01e5f6](https://github.com/KILTprotocol/kilt-node/pull/768/commits/b01e5f6fa289a6c99c1d96f120301b818b76c576)

pallets/pallet-bonded-coins/src/types.rs
Ad96el: Renamed in: [8d203c9](https://github.com/KILTprotocol/kilt-node/pull/768/commits/8d203c9413603fb8da1dbddbf75e117eda1d7b51)

pallets/pallet-bonded-coins/src/traits.rs
Ad96el: Don't know. Doesn't really make sense. I just copied it over from the polkadot SDK

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: renamed the error to: [633e119](https://github.com/KILTprotocol/kilt-node/pull/768/commits/633e119696a15443164ebc652088950e15e3b047)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Renamed in: [e401cef](https://github.com/KILTprotocol/kilt-node/pull/768/commits/e401cef05fe348907c08b9894c1e7c5e1cfe5740)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I moved the reset tx to its own extrinsic in: [0f8c286](https://github.com/KILTprotocol/kilt-node/pull/768/commits/0f8c286505b02e167123d6d89164437d34a17fe9)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I need to investigate that. I can imagine, benchmarking the worst case can exceed the resources of a block, depending on how many currencies we might allow. For now, I moved out the reset of the team to its own tx: [0f8c286](https://github.com/KILTprotocol/kilt-node/pull/768/commits/0f8c286505b02e167123d6d89164437d34a17fe9)

ntn-x2: Yes but if we calculate weights based on the length of the input parameter, and that would make a tx too big, then the tx is ACTUALLY too big. That's not a worst case weight, that is the actual weight the tx takes, since we take the exact size of the input in consideration.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I introduced a default value: [341ddf8](https://github.com/KILTprotocol/kilt-node/pull/768/commits/341ddf8a33ef12d53dede9f9adca2821ac4737a3)

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -151,13 +155,16 @@ pub mod pallet {
...
 
-	// Errors inform users that something went wrong.
 	#[pallet::error]
 	pub enum Error<T> {
-		Todo,
+		IndexOutOfBounds,
+		InvalidInput,
+		Internal,
+		UnknownPool,
+		UnAuthorized,
```
rflechtner, line 167: I think the assets pallet has pretty good descriptive error names (https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#645), so I tried to re-use them where our semantics are similar. They call this one `NoPermission`, which I also like.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,66 +176,178 @@ pub mod pallet {
...
 		#[pallet::call_index(0)]
 		#[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
-		pub fn create_pool(_origin: OriginFor<T>) -> DispatchResult {
-			todo!()
+		pub fn create_pool(
+			origin: OriginFor<T>,
+			curve: CurveInput<CurveParameterInputOf<T>>,
+			currencies: BoundedVec<TokenMetaOf<T>, T::MaxCurrencies>,
+			denomination: u8,
+			pool_manager: AccountIdOf<T>,
```
rflechtner, line 184: I think we wanted to move this to its own extrinsic? (`set_manager` maybe?)

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,66 +176,178 @@ pub mod pallet {
...
+			transferable: bool,
+		) -> DispatchResult {
+			let who = T::PoolCreateOrigin::ensure_origin(origin)?;
+
+			let currency_length = currencies.len();
+
+			let checked_curve = curve.try_into().map_err(|_| Error::<T>::InvalidInput)?;
+
+			let current_asset_id = NextAssetId::<T>::get();
```
rflechtner, line 193: @ntn-x2 suggested to have users supply the asset ids to avoid the situation that we reach the numerical bounds of the asset id type and cannot safely wrap around because some of the ids we would be trying to use then are still in use. Did we decide this is a non-issue?

On a related note, the assets pallet now has the exact same functionality - not sure it exposes it via any public trait though.

rflechtner: Maybe this can be a callback / hook instead, so we can implement this in the runtime to rely on the asset pallet's id increment solution. This would also avoid the pretty bad situation where the `NextAssetId` gets registered without this pallet knowing (e.g., via the asset pallet's extrinsics), which would break pool creation until governance steps in and force-increments this.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,66 +176,178 @@ pub mod pallet {
...
+					Ok(())
+				})?;
+
+			// Touch the pool account in order to be able to transfer the collateral currency to it
+			T::CollateralCurrency::touch(T::CollateralAssetId::get(), &pool_id.clone().into(), &who)?;
+
+			Pools::<T>::set(
+				&pool_id,
+				Some(PoolDetails::new(
+					pool_manager,
```
rflechtner, line 235: btw I think we need a manager and an owner, bc we want to allow 'unmanaged' pools (nobody has special permission) but still we need an owner to send and excess collateral to - and we may still grant the owner the right to destroy a pool.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -169,66 +176,178 @@ pub mod pallet {
...
+			let asset_id = pool_details
+				.bonded_currencies
+				.get(currency_idx as usize)
+				.ok_or(Error::<T>::IndexOutOfBounds)?;
+
+			T::Fungibles::reset_team(
+				asset_id.to_owned(),
+				pool_id.into(),
+				team.admin,
+				team.issuer,
```
rflechtner, line 273: issuer should always be the pool account, otherwise they can mint coins which are not backed by any collateral

pallets/pallet-bonded-coins/src/types.rs
```diff
@@ -9,8 +9,9 @@ pub struct Locks {
 	pub allow_swap: bool,
 }
 
-#[derive(Clone, Encode, Decode, PartialEq, Eq, TypeInfo, MaxEncodedLen, Debug)]
+#[derive(Clone, Encode, Decode, PartialEq, Eq, TypeInfo, MaxEncodedLen, Debug, Default)]
 pub enum PoolStatus<LockType> {
+	#[default]
```
rflechtner, line 14: default should be locked I'd say?

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: Made a proposal for that on #772

pallets/pallet-bonded-coins/src/types.rs
rflechtner: another proposal on #772 (355a49efd65cbda909183ea7f7949a3f5925b1f9)

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: If the asset ID in here is strictly depending on some conditions of the assets pallet, then I would recommend making it a injectable hook for exactly the reason where a new asset class is created in that pallet and this pallet is not aware of it.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Ah okay. I would maybe wait for this change, until the [wireframe PR](https://github.com/KILTprotocol/kilt-node/pull/772) is merged. Otherwise, I get some unnecessary merge conflicts.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Updated the error name in: [32f5988](https://github.com/KILTprotocol/kilt-node/pull/768/commits/32f5988b11ef763b7c81d918c878e2490703a8f5)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Ups... right: [32f5988](https://github.com/KILTprotocol/kilt-node/pull/768/commits/32f5988b11ef763b7c81d918c878e2490703a8f5)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: > Did we decide this is a non-issue?

Yes, we did. This would only happen if there were already over 4 million bonded currencies. If that ever becomes a problem, it would mean we’re very successful,

> This would also avoid the pretty bad situation where the NextAssetId gets registered without this pallet knowing (e.g., via the asset pallet's extrinsics)

This shouldn’t be an issue. In the runtime, only root can create new assets. Of course, if there’s a public referendum to create a new asset in the pallet, it should be blocked. 


> If the asset ID in here is strictly depending on some conditions of the assets pallet, then I would recommend making it a injectable hook 

I’m worried that adding a hook could cause issues later. Right now, there’s no `NextAssetId` struct in the fungibles pallet, but this feature is coming in version 14. We need this storage map temporarily. If we remove the storage value completely and allow users to define any asset ID not in use, we wouldn’t be able to migrate to this feature later. We’d also lose control over which assets exist.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I will also wait tackling that until the proposal is merged. Some for: https://github.com/KILTprotocol/kilt-node/pull/768#discussion_r1820572239

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Added tx in: [7f9a319](https://github.com/KILTprotocol/kilt-node/pull/768/commits/7f9a31966ead7d4e74c6acf20ff5e0cf968e3a36)

Ad96el: Or should I check, if who is owner? Should the owner be able to update the manager, or should the manager have the power?

Ad96el: I would argue, the manager should have the power. Follows the logical flow of a root in substrate

rflechtner: The manager should be able to set the new manager. This follows from our idea to have manager-less pools, where the owner is set to None - and then can never be re-set again. This way users can be certain that there is no account that has special powers for the pool.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Yes, that would be possible and also my preferred way. I would keep the `NexAssetId` storage value, until we hit a version, where the feature is implemented, and we can migrate over and use a hook.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+						.saturating_mul(currency_length.saturated_into())
+						.saturated_into(),
+				),
+			)?;
+
+			currencies
+				.iter()
+				.enumerate()
+				.try_for_each(|(idx, entry)| -> DispatchResult {
+					let asset_id: &T::AssetId = currency_ids.get(idx).ok_or(Error::<T>::IndexOutOfBounds)?;
```
rflechtner, line 219:
```suggestion
			currencies
				.into_iter()
				.zip(currency_ids.into_iter())
				.try_for_each(|(entry, asset_id)| -> DispatchResult {
```
That should do the trick?

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+					T::Fungibles::create(asset_id.clone(), pool_id.clone().into(), false, entry.min_balance)?;
+
+					// set metadata for new asset class
+					T::Fungibles::set(
+						asset_id.clone(),
+						&pool_id.clone().into(),
+						entry.name.clone().into(),
+						entry.symbol.clone().into(),
```
rflechtner, line 227: with into_iter and maybe proper destructuring `asset_id`, `entry.name` and `entry.symbol` can all be consumed

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+						.saturated_into(),
+				),
+			)?;
+
+			currencies
+				.iter()
+				.enumerate()
+				.try_for_each(|(idx, entry)| -> DispatchResult {
+					let asset_id: &T::AssetId = currency_ids.get(idx).ok_or(Error::<T>::IndexOutOfBounds)?;
+					T::Fungibles::create(asset_id.clone(), pool_id.clone().into(), false, entry.min_balance)?;
```
rflechtner, line 220: may be worth casting pool_id to pool_account once before entering the loop

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+
+		#[pallet::call_index(2)]
+		#[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
+		pub fn reset_team(
+			origin: OriginFor<T>,
+			pool_id: T::PoolId,
+			team: PoolManagingTeam<AccountIdOf<T>>,
+			currency_idx: u32,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
rflechtner, line 258: All the signed origins should now be migrated to the configurable `DefaultOrigin`
```suggestion
			let who = T::DefaultOrigin::ensure_origin(origin)?;
```

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+			Pools::<T>::try_mutate(pool_id, |maybe_entry| -> DispatchResult {
+				if let Some(entry) = maybe_entry {
+					ensure!(entry.is_manager(&who), Error::<T>::NoPermission);
+					entry.manager = manager;
+					Ok(())
+				} else {
+					Err(Error::<T>::UnknownPool.into())
+				}
```
rflechtner, line 295: works but isn't this more readable:

```suggestion
				let entry = maybe_entry.as_mut().ok_or(Error::<T>::UnknownPool)?;
				ensure!(entry.is_manager(&who), Error::<T>::NoPermission);
				entry.manager = manager;
				Ok(()) // This may actually be optional
```

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
-	impl<T: Config> Pallet<T> {}
+	impl<T: Config> Pallet<T> {
+		fn generate_sequential_asset_ids(
+			mut start_id: T::AssetId,
+			count: usize,
+		) -> Result<(BoundedCurrencyVec<T>, T::AssetId), Error<T>> {
+			let mut currency_ids_vec = Vec::new();
+			for _ in 0..count {
+				currency_ids_vec.push(start_id.clone());
+				start_id = start_id.saturating_plus_one();
```
rflechtner, line 364:
```suggestion
				start_id.saturating_inc();
```

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -175,66 +182,193 @@ pub mod pallet {
...
+			}
+
+			let currency_array =
+				BoundedVec::<FungiblesAssetIdOf<T>, T::MaxCurrencies>::try_from(currency_ids_vec.clone())
```
rflechtner, line 368: why does this have to be cloned?

pallets/pallet-bonded-coins/src/types.rs
rflechtner: this conflicts with the implementation on #772 and needs to be reverted

Ad96el: Ops, [5677282](https://github.com/KILTprotocol/kilt-node/pull/768/commits/5677282d8308eaa6fbb972f933ec20a3e6d30ad0)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: introduced in: [30cbeb3](https://github.com/KILTprotocol/kilt-node/pull/768/commits/30cbeb34d5d6e3ae8a7561c38beef14206755a5d)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: love it: [30cbeb3](https://github.com/KILTprotocol/kilt-node/pull/768/commits/30cbeb34d5d6e3ae8a7561c38beef14206755a5d)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: [8f89de9](https://github.com/KILTprotocol/kilt-node/pull/768/commits/8f89de90fad830a20894cb74fca73ea0962a6c82)

rflechtner (approved):
da 💣

Ad96el:
@ntn-x2 I would merge this PR. If you still find some issues, please comment on this PR. I will address them in a new one.

PR #769, opened 2024-10-24
chore: update pallets to benchmarks v2

ntn-x2:


pallets/did/src/origin.rs
```diff
@@ -123,13 +123,12 @@ where
 	}
 }
 
-#[cfg(all(test, feature = "runtime-benchmarks"))]
+#[cfg(test)]
 mod tests {
-	use super::EnsureDidOrigin;
-
+	#[cfg(feature = "runtime-benchmarks")]
 	#[test]
```
Ad96el, line 129: This seems a bit odd to me

Ad96el (commented):
LGTM, nice changes

pallets/did/src/origin.rs
Ad96el: Was it previously, throwing warnings?

ntn-x2: Yes, because it meant the `tests` module was not used only for tests but also for something else, and it was indeed throwing a lint. Ideally I agree with the lint, a `tests` module should only be used for tests. Then if something from inside the module is used when testing some feature, then it should be feature-gated inside the module.

Ad96el: I see

Ad96el (approved):

PR #772, opened 2024-10-29
refactor: bonded coins wireframe adjustments

rflechtner:
Combines some changes that result from discussions on the feature branches but are more general than the individual features.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -118,7 +118,7 @@ pub mod pallet {
 		/// Who can create new bonded currency pools.
 		type PoolCreateOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
 		/// The type used for pool ids
-		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<Self::AccountId>;
+		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<[u8; 32]>;
 
 		/// The type used for asset ids. This is the type of the bonded currencies.
 		type AssetId: Parameter + Member + FullCodec + TypeInfo + MaxEncodedLen + Saturating + One + Default;
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -124,7 +124,7 @@ pub mod pallet {
 		/// The origin for permissioned operations (force_* transactions).
 		type ForceOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 		/// The type used for pool ids
-		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<[u8; 32]>;
+		type PoolId: Parameter + MaxEncodedLen + Into<[u8; 32]>;
 
 		/// The type used for asset ids. This is the type of the bonded currencies.
 		type AssetId: Parameter + Member + FullCodec + TypeInfo + MaxEncodedLen + Saturating + One + Default;
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -80,6 +80,8 @@ pub mod pallet {
 
 	pub(crate) type PassiveSupply<T> = Vec<T>;
 
+	const LOG_TARGET: &str = "runtime::pallet-bonded-coins";
+
 	/// Configure the pallet by specifying the parameters and types on which it depends.
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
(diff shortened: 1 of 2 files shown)


pallets/pallet-bonded-coins/src/types.rs
```diff
@@ -36,7 +36,8 @@ impl<LockType> PoolStatus<LockType> {
 
 #[derive(Clone, Encode, Decode, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
 pub struct PoolDetails<AccountId, ParametrizedCurve, Currencies> {
-	pub manager: AccountId,
+	pub owner: AccountId,
+	pub manager: Option<AccountId>,
```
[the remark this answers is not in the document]
ntn-x2: Why not make use of Rust's strong type system here? I think it's an option is the right type here.

pallets/pallet-bonded-coins/src/types.rs
```diff
@@ -45,47 +46,46 @@ pub struct PoolDetails<AccountId, ParametrizedCurve, Currencies> {
...
+	pub fn new(owner: AccountId, curve: ParametrizedCurve, bonded_currencies: Currencies, transferable: bool) -> Self {
 		Self {
-			manager,
+			manager: Some(owner.clone()),
+			owner,
 			curve,
 			bonded_currencies,
 			transferable,
-			state,
+			state: PoolStatus::Locked(Default::default()),
```
ntn-x2, line 58: This should be the `Default` implementation of `PoolStatus`, I'd say.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -115,10 +115,14 @@ pub mod pallet {
 
 		/// The asset id of the collateral currency.
 		type CollateralAssetId: Get<CollateralAssetIdOf<Self>>;
-		/// Who can create new bonded currency pools.
+		/// The origin for most permissionless and priviledged operations.
+		type DefaultOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
```
ntn-x2, line 119: Where would this origin be used? For what extrinsics?

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -115,10 +115,14 @@ pub mod pallet {
...
+		/// The dedicated origin for creating new bonded currency pools (typically permissionless).
 		type PoolCreateOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+		/// The origin for permissioned operations (force_* transactions).
+		type ForceOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 		/// The type used for pool ids
-		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<Self::AccountId>;
+		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<[u8; 32]>;
```
ntn-x2, line 126: `From<[u8; 32]>` is then not necessary. You can create a `[u8; 32]` with this trait and then call `AccountId::from(<[u8; 32]>)`. `PoolId` is not required to deal with `From<[u8; 32]>`, unless I am missing something.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: For most of them, like mint_into, burn_into, start_destroy ... everything that's not creation or a force origin

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: I also can't tell you what that was for, but I assumed it had a purpose. @Ad96el why do we need to be able to create a pool id from a 32 byte string?

pallets/pallet-bonded-coins/src/types.rs
rflechtner: Yeah guess I'll just cherry-pick 9ae15d2310bb6a7f53b8e57cf7e972820f5f674b then. I had it implemented that way, can't tell you when exactly this disappeared.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I am also not sure anymore. We can get rid of the trait bound, I guess.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: that's what I took from our conversations, that you wanted a configurable origin for all of these things?

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: c0722ea8bd2fdc02a1da7ba98e0a852d3069e743

pallets/pallet-bonded-coins/src/types.rs
rflechtner: 835cfeacce6711ee77758a5db6e75b59fa690cf2

PR #782, opened 2024-11-05
feat: support for unique linking

ntn-x2:
Part of https://github.com/KILTprotocol/ticket/issues/3650. Built on top of https://github.com/KILTprotocol/kilt-node/pull/781.

## Trade-off

I chose to go this way instead of providing an optional counter, because providing a counter would require one of the following two approaches:

1. Transform the storage double map into a counted one, requiring a migration also for our currently-deployed pallet, which I wanted to avoid
2. Not use a counter, but iterate every time to make sure there are still "spots" left for the current DID. This would require changing the benchmarking logic as now we have a potentially unbounded iteration happening. I also wanted to avoid that.

Hence, the solution was to provide a somehow more limited feature of simply specifying whether the links are expected to be unique per DID or not. This, as long as we set `false` for our deployed pallets would not require any storage migration, and does not require any changes in the benchmarks, so I found it a good compromise.

### DIFF
--- a/pallets/pallet-did-lookup/src/lib.rs
+++ b/pallets/pallet-did-lookup/src/lib.rs
@@ -131,6 +131,11 @@ pub mod pallet {
 
 		/// Migration manager to handle new created entries
 		type BalanceMigrationManager: BalanceMigrationManager<AccountIdOf<Self>, BalanceOf<Self, I>>;
+
+		// Flag specifying whether there should only ever be a single account <-> DID
+		// link, or multiple.
+		#[pallet::constant]
+		type UniqueLinkingEnabled: Get<bool>;
 	}
 
 	#[pallet::pallet]
@@ -196,6 +201,9 @@ pub mod pallet {
 		///
 		/// NOTE: this will only be returned if the storage has inconsistencies.
 		Migration,
+		/// The deployed pallet supports a single account <-> link, which has
+		/// already been previously created for the provided DID.
+		LinkExisting,
 	}
 
 	#[pallet::genesis_config]
@@ -445,6 +453,19 @@ pub mod pallet {
 				deposit,
 				did: did_identifier.clone(),
 			};
+
+			// We don't care if an account is linked to a new DID (handled below by
+			// replacing the old with the new), but we care that if the unique linking flag
+			// is on, the DID can have a single account linked.
+			let is_did_already_linked = ConnectedAccounts::<T, I>::iter_key_prefix(&did_identifier)
+				.next()
+				.is_some();
+			let is_unique_flag_enabled = <T as Config<I>>::UniqueLinkingEnabled::get();
+			// Either flag not enabled, or if it is, the DID is not already linked.
+			ensure!(
+				!is_unique_flag_enabled || !is_did_already_linked,
+				Error::<T, I>::LinkExisting
+			);
 
 			LinkableAccountDepositCollector::<T, I>::create_deposit(
 				record.clone().deposit.owner,
--- a/pallets/pallet-did-lookup/src/lib.rs
+++ b/pallets/pallet-did-lookup/src/lib.rs
@@ -454,18 +454,13 @@ pub mod pallet {
 				did: did_identifier.clone(),
 			};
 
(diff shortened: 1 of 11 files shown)


pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -131,6 +131,11 @@ pub mod pallet {
 
 		/// Migration manager to handle new created entries
 		type BalanceMigrationManager: BalanceMigrationManager<AccountIdOf<Self>, BalanceOf<Self, I>>;
+
+		// Flag specifying whether there should only ever be a single account <-> DID
+		// link, or multiple.
+		#[pallet::constant]
+		type UniqueLinkingEnabled: Get<bool>;
```
ntn-x2, line 138: I made this a `Get<bool>` instead of a const to both allow clients to query whether this pallet enforces unique links or not, and to also make testing easier since this value can be dynamically mocked, as opposed to a const.

pallets/pallet-did-lookup/src/lib.rs
```diff
@@ -446,6 +454,19 @@ pub mod pallet {
...
+			// is on, the DID can have a single account linked.
+			let is_did_already_linked = ConnectedAccounts::<T, I>::iter_key_prefix(&did_identifier)
+				.next()
+				.is_some();
+			let is_unique_flag_enabled = <T as Config<I>>::UniqueLinkingEnabled::get();
+			// Either flag not enabled, or if it is, the DID is not already linked.
+			ensure!(
+				!is_unique_flag_enabled || !is_did_already_linked,
+				Error::<T, I>::LinkExisting
+			);
```
Ad96el, line 468: Is this right? We only need to check  `ConnectedAccounts`, if `is_unique_flag_enabaled` is true, or not?

pallets/pallet-did-lookup/src/mock.rs
```diff
@@ -109,6 +110,23 @@ parameter_types! {
...
+	fn set(flag: bool) {
+		FlagStorage::set(flag)
+	}
+}
+
+impl Get<bool> for UniqueLinkEnabledFlag {
+	fn get() -> bool {
+		FlagStorage::get()
+	}
+}
```
Ad96el, line 128: sneaky

Ad96el (commented):
Some questions.

pallets/pallet-did-lookup/src/mock.rs
```diff
@@ -163,6 +182,7 @@ pub struct ExtBuilder {
 	balances: Vec<(AccountId, Balance)>,
 	/// list of connection (sender, did, connected address)
 	connections: Vec<(AccountId, SubjectId, LinkableAccountId)>,
+	unique_flag: Option<()>,
```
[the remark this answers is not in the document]
ntn-x2: Yeah I mean the exposed function lets you set this flag to true, and by default it's false anyway. But I switched to a `bool` in a new commit: https://github.com/KILTprotocol/kilt-node/pull/782/commits/21e97e75b8c07c1d2ca8165bcfb77f956e5cfb99

pallets/pallet-did-lookup/src/lib.rs
ntn-x2: This reads as "Either the unique flag is not on (in which case we don't care), or (if it's on) make sure the DID is not already linked". It's equivalent to `(is_unique_flag_enabled && !is_did_already_linked)`.

Ad96el: Yes, I was more considering the performance. If the flag is enabled, then we don't need to check if the did has already an association. But this is probably a bit nitpicky.

Ad96el (approved):

pallets/pallet-did-lookup/src/lib.rs
ntn-x2: Yeah good point, I'll change that.

ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/782/commits/f853e2bf22ad70fbb88e8b0c37dcf4761195581e

PR #784, opened 2024-11-05
feat: new pallets into runtime

ntn-x2:
Part of https://github.com/KILTprotocol/ticket/issues/3650, built on top of https://github.com/KILTprotocol/kilt-node/pull/782.

I left a few comments on the review to help the reviewer understanding the context of this changeset.

## Checklist

- [x] Add dotnames pallet to both runtimes
- [x] Add second linking deployment with unique linking to both runtimes
- [ ] ~Add new runtime API for batch resolution to both runtimes~ -> will do in a separate PR along with a new runtime API for unique linking resolution


pallets/pallet-web3-names/src/lib.rs
```diff
@@ -225,7 +218,11 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config<I>, I: 'static> Pallet<T, I> {
+	impl<T: Config<I>, I: 'static> Pallet<T, I>
+	where
```
ntn-x2, line 222: I changed from `TryFrom<Vec<u8>, Error = Error<Self, I>>` to here, where the error should be representable as a pallet error, but not necessarily be a pallet error. This allow us, below, to define a new type for the dotname that does not directly depend on the runtime, hence it's much easier to unit test.

runtimes/common/src/dot_names/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+		match value {
+			DotNameValidationError::TooLong => Self::TooLong,
+			DotNameValidationError::TooShort => Self::TooShort,
+			DotNameValidationError::InvalidCharacter => Self::InvalidCharacter,
+		}
+	}
+}
+
+#[derive(Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct DotName<const MIN_LENGTH: u32, const MAX_LENGTH: u32>(BoundedVec<u8, ConstU32<MAX_LENGTH>>);
```
ntn-x2, line 53: All the changes were made to allow us to do this 😄 instead of the mess we currently have for the `AsciiWeb3Name` type we use for our own web3names.

runtimes/peregrine/src/lib.rs
```diff
@@ -707,6 +707,21 @@ impl pallet_did_lookup::Config for Runtime {
...
+	type Currency = Balances;
+	type Deposit = constants::did_lookup::DidLookupDeposit;
+	type DidIdentifier = DidIdentifier;
+	type EnsureOrigin = did::EnsureDidOrigin<DidIdentifier, AccountId>;
+	type OriginSuccess = did::DidRawOrigin<AccountId, DidIdentifier>;
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type UniqueLinkingEnabled = ConstBool<true>;
+	// TODO: Change
+	type WeightInfo = ();
```
Ad96el, line 722: can we already add some benchmarks?

ntn-x2: I mean I can copy the weights we already have. Will do that.

ntn-x2: Copied the weights here: https://github.com/KILTprotocol/kilt-node/pull/784/commits/01b3ce5d8c5392a9ff4ae1ce44bbab02e9683a73. I noticed that the benchmarks are now failing and the logic must be made generic. I'll do that in a separate PR. For now the weights should be fine.

Ad96el (approved):

PR #780, opened 2024-11-04
chore: mocks for bonding curve pallet

Ad96el:

PR #785, opened 2024-11-05
test: create pool

rflechtner:
## re/ KILTProtocol/ticket#3671

Adds unit tests for creating pools.
To be rebased and merged after https://github.com/KILTprotocol/kilt-node/pull/780

Tests currently failing (fix TBD in another PR):

- [ ] Exhausting Asset Id value space should result in `ArithmeticError::Overflow`
- [x] You should not be able to create a pool where a bonded currency is also the pool's collateral 

## Checklist:

    * Link the PR/Ticket here


pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+
+#[test]
+fn single_currency() {
+	let initial_balance = 100_000_000_000_000_000u128;
+	ExtBuilder::default()
+		.with_native_balances(vec![(ACCOUNT_00, initial_balance)])
+		.with_collaterals(vec![DEFAULT_COLLATERAL_CURRENCY_ID])
+		.build()
+		.execute_with(|| {
+			assert_eq!(NextAssetId::<Test>::get(), 0);
```
ntn-x2, line 21: We usually don't run assertions on state that is created in the mock. If you are to run these assertions, they should be put inside the mock `build` function, so that you are sure calling `build()` gives you the state you expect.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+			));
+
+			let pool_id = calculate_pool_id(&[0]);
+
+			let details = Pools::<Test>::get(&pool_id).unwrap();
+
+			assert!(details.is_owner(&ACCOUNT_00));
+			assert!(details.is_manager(&ACCOUNT_00));
+			assert!(details.transferable);
+			assert_eq!(details.state, PoolStatus::Locked(Default::default()));
```
ntn-x2, line 49: I think we want to math against the exact state, and not use `Default::default()`. If tomorrow the default canges, we would fail to spot the change in the tests.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+
+			assert_eq!(
+				Balances::free_balance(ACCOUNT_00),
+				initial_balance.sub(BondingPallet::calculate_pool_deposit(1))
+			);
+
+			System::assert_has_event(BondingPalletEvents::PoolCreated { id: pool_id.clone() }.into());
+
+			// TODO: check events or storage of linked pallets?
+			System::assert_has_event(
```
ntn-x2, line 65: So the `crate_pool` generates a `ForceCreated` event? Why is that?

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+			assert_err!(
+				BondingPallet::create_pool(
+					origin,
+					curve,
+					DEFAULT_COLLATERAL_CURRENCY_ID,
+					BoundedVec::truncate_from(vec![bonded_token]),
+					10,
+					true
+				),
+				AssetsPalletErrors::<Test>::Unknown
```
ntn-x2, line 215: Sorry I have missed follow-up on previous conversations. While the `Unknown` error here if a collateral does not exist? Shouldn't the error be more specific?

rflechtner: I also tripped here for a minute, but it's the assets pallet's `Unknown` error, which is meant to express that an asset is unknown:
```rust
		/// The given asset ID is unknown.
		Unknown,
```
Weird choice indeed because it looks like an 'unknown error', but when it shows up in our runtime as `Fungibles:Unknown` in combination with the docstring associated with it (which is carried over to the metadata afaik) it should become clearer.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: Turns out the assets pallet performs a 'force create' in the implementation of the `fungibles::create` trait method:
https://docs.rs/pallet-assets/40.0.0/src/pallet_assets/impl_fungibles.rs.html#187

ntn-x2: Ok but these are unit tests, you should not test storage and events generated by other pallets. Also because if you swap out the implementation of Fungibles with a different one, the events would be gone.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: makes sense, I added this to make sure we detect relevant changes to the mock state though - do we not do this ?

ntn-x2: Those changes should be detected inside the mock itself then, not here. Moving those assertions inside the `build` function, as we do here for instance: https://github.com/KILTprotocol/kilt-node/blob/4199b6fed27e37a1e8f774f0772e7667c82bf80d/pallets/pallet-asset-switch/src/mock.rs#L261

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+					depositor: ACCOUNT_00,
+				}
+				.into(),
+			);
+		});
+}
+
+#[test]
+fn multi_currency() {
+	let initial_balance = 100_000_000_000_000_000u128;
```
Ad96el, line 106: Maybe move the balance to mock. It is used multiple times

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: Yeah this section kinda goes beyond the scope of unit tests, but because the pallet is 80% calls to other pallets, we need to at least test that these are called correctly - which seems quite hard to do in Substrate? We'd need to somehow _actually_ mock the assets pallet (right now we use its real implementation to test against) and be able to run assertions on the arguments with which trait methods are being called, which I'm not sure we can do? So we need to make a call on how we want to test here - either mock away the fungibles trait implementation and capture calls, or expand the scope of the tests to be more integration-testy.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: yeah fair, 0423e17f1e8d35574f3912fa72c38e21e4c9c745

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
```diff
@@ -1 +1,251 @@
...
+	ExtBuilder::default()
+		.with_native_balances(vec![(ACCOUNT_00, initial_balance)])
+		.with_collaterals(vec![DEFAULT_COLLATERAL_CURRENCY_ID])
+		.build()
+		.execute_with(|| {
+			assert_eq!(NextAssetId::<Test>::get(), 0);
```
Ad96el, line 112: Can we also check the deposit? We have checked it for a single currency. For multiple, it should differ.

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: I'll just remove them for now (63b1b43689fe3f7c95b21837e1975d7d8e07616e). We're not setting the `NextAssetId` at all in the build method, which is probably an issue...

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: I can't seem to be able to link to the right line, but this is being checked with 
```rust
			assert_eq!(
 				Balances::free_balance(ACCOUNT_00),
 				initial_balance.sub(BondingPallet::calculate_pool_deposit(3))
 			);
```

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: I assume we have no equivalent to jests `spyOn` or auto-mock functionality here?

ntn-x2: No, that would have to be implemented in the trait mock implementation, where you keep track yourself. In general, I am not a very big fan of using existing stuff for unit tests, but rather mock everything I need to mock. In this case I think it's fine to use the assets pallet, but I would not check for the events generates, and rather use the same trait bounds you define in the config trait to verify that the conditions are met. Meaning, calling asserts only on functions that are exposed via the traits that the config object is required to implement. So checking the balance of a user, for instance, would not call the assets pallet directly, but would get the value via the trait that is responsible for returning that value. That would be a good compromise, I think.

rflechtner: makes sense! Let me see if I can figure out how to call into the asset pallet's fungibles trait from the tests

pallets/pallet-bonded-coins/src/tests/transactions/create_pool.rs
rflechtner: e458d8cd9aa2fd655b0db4d9921942d66e89af37

ntn-x2 (approved):
Good enough for me 👍

PR #786, opened 2024-11-05
feat: reset team on refund

rflechtner:
Suggested change to avoid unexpected changes to currencies and issuances while refunding. This resets all asset teams to the pool account when calling `start_refund`.

This is waiting for a runtime implementation of our new pallet so that we can see the impact on the transaction weight that this has.

## Checklist:

    * Link the PR/Ticket here


rflechtner:
> @rflechtner What is this?

We can pick this up after trialing the pallet on a testnet. We're currently not quite sure if we want to reset the asset admins when the pool moves to refunding state. This would do it, but would need to be benchmarked properly.

rflechtner:
> @rflechtner Can we also close this?

Mhh this is something that would still make sense, although I don't see that this gets any priority

PR #789, opened 2024-11-06
chore: benchmarks

Ad96el:
# fixes: [#3618](https://github.com/KILTprotocol/ticket/issues/3618)

build on top of: https://github.com/KILTprotocol/kilt-node/pull/799


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -360,11 +377,17 @@ pub mod pallet {
 
 			Self::deposit_event(Event::PoolCreated { id: pool_id });
 
-			Ok(())
+			let currency_index = currency_length.saturated_into();
+			Ok(Some(match checked_curve {
```
ntn-x2, line 381: You return the same pre-computed weight, so this is not necessary.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -529,11 +555,21 @@ pub mod pallet {
 				T::Fungibles::freeze(target_currency_id, &beneficiary).map_err(|freeze_error| freeze_error.into())?;
 			}
 
-			Ok(())
+			Ok(Some(match pool_details.curve {
```
ntn-x2, line 558: This one makes sense instead.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -758,32 +804,44 @@ pub mod pallet {
 				Self::deposit_event(Event::RefundComplete { id: pool_id });
 			}
 
-			Ok(())
+			Ok(Some(T::WeightInfo::refund_account(currency_count.to_owned())).into())
 		}
 
 		#[pallet::call_index(11)]
 		#[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
```
ntn-x2, line 811: Weight not updated.

scripts/run_benches_for_pallets.sh
```diff
@@ -19,17 +19,19 @@ pallets=(
 	pallet-web3-names
 	parachain-staking
 	public-credentials
+	pallet-asset-switch
```
ntn-x2, line 22: Can you pull this out real quick in a separate PR?

scripts/run_benches_for_pallets.sh
```diff
@@ -19,17 +19,19 @@ pallets=(
...
 )
 
 // Add Peregrine-only pallets here!
 if [ "$runtime" = "peregrine" ]; then
   pallets+=(
-	pallet-asset-switch
-  )
+		pallet-bonded-coins
```
ntn-x2, line 28: The codebase still does not contain any runtime code, so this change should be introduced in that PR, not in here.

scripts/run_benches_for_pallets.sh
```diff
@@ -19,17 +19,19 @@ pallets=(
...
+	)
 fi
 
 echo "[+] Running all default weight benchmarks for $runtime --chain=$chain"
```
ntn-x2, line 33: This space is probably not necessary?

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+
+	fn create_bonded_asset<T: Config>(asset_id: T::AssetId) {
+		let pool_account = account("bonded_owner", 0, 0);
+		<T as Config>::Fungibles::create(asset_id, pool_account, false, 1u128.saturated_into())
+			.expect("Creating bonded asset should work");
+	}
+
+	fn create_bonded_currencies_in_range<T: Config>(c: u32, is_destroying: bool) -> Vec<FungiblesAssetIdOf<T>> {
+		let mut asset_ids = Vec::new();
+		for i in 0..c {
```
ntn-x2, line 105: I know it's very nit-picky, but I think it would be slitghly more readable to replace this with `1..=c`, to use the same logic as the linear parameter computation.

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+		let collateral_id = calculate_default_collateral_asset_id::<T>();
+		create_collateral_asset::<T>(collateral_id.clone());
+		collateral_id
+	}
+
+	fn set_collateral_balance<T: Config>(asset_id: CollateralAssetIdOf<T>, who: &AccountIdOf<T>, amount: u128)
+	where
+		<T as Config>::CollateralCurrencies: MutateFungibles<T::AccountId>,
+	{
+		T::CollateralCurrencies::set_balance(asset_id, who, amount.saturated_into());
```
ntn-x2, line 92: Should you also `assert` after this, to make sure it worked?

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+	fn calculate_default_collateral_asset_id<T: Config>() -> CollateralAssetIdOf<T> {
+		T::BenchmarkHelper::calculate_collateral_asset_id(u32::MAX)
+	}
+
+	fn create_default_collateral_asset<T: Config>() -> CollateralAssetIdOf<T>
+	where
+		<T as Config>::CollateralCurrencies: Create<T::AccountId>,
+	{
+		let collateral_id = calculate_default_collateral_asset_id::<T>();
+		create_collateral_asset::<T>(collateral_id.clone());
```
ntn-x2, line 84: Also here, should you do an `assert` to make sure the state is what you expect it to be?

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+	fn create_bonded_currencies_in_range<T: Config>(c: u32, is_destroying: bool) -> Vec<FungiblesAssetIdOf<T>> {
+		let mut asset_ids = Vec::new();
+		for i in 0..c {
+			let asset_id = T::BenchmarkHelper::calculate_bonded_asset_id(i);
+			asset_ids.push(asset_id.clone());
+			create_bonded_asset::<T>(asset_id.clone());
+			if is_destroying {
+				<T as Config>::Fungibles::start_destroy(asset_id, None).expect("Destroying should work");
+			}
+		}
```
ntn-x2, line 112: Asserts?

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+
+		asset_ids
+	}
+
+	fn set_fungible_balance<T: Config>(asset_id: FungiblesAssetIdOf<T>, who: &AccountIdOf<T>, amount: u128)
+	where
+		<T as Config>::Fungibles: MutateFungibles<T::AccountId>,
+	{
+		T::Fungibles::set_balance(asset_id, who, amount.saturated_into());
```
ntn-x2, line 121: Assert, especially since this is a non-failing function.

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+		let beneficiary = AccountIdLookupOf::<T>::from(account_origin.clone());
+		let amount_to_mint = 10u128;
+
+		#[extrinsic_call]
+		mint_into(
+			origin as T::RuntimeOrigin,
+			pool_id,
+			0,
+			beneficiary,
+			amount_to_mint.saturated_into(),
```
ntn-x2, line 420: No conversions in the benchmarked extrinsic itself, nor `::get()`. Same below in the other cases.

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -1 +1,855 @@
...
+		let pool_id = create_pool::<T>(
+			curve,
+			bonded_currencies.clone(),
+			Some(account_origin.clone()),
+			None,
+			None,
+		);
+
+		#[extrinsic_call]
+		_(origin as T::RuntimeOrigin, pool_id.clone(), T::MaxCurrencies::get());
```
ntn-x2, line 665: No CloNe EithEr!11!

ntn-x2 (changes requested):

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -696,11 +736,6 @@ pub mod pallet {
 			)?
 			.into();
 
-			if burnt.is_zero() {
-				// no funds available to be burnt on account; nothing to do here
-				return Ok(());
-			}
-
```
Ad96el, line 703: @rflechtner is it a breaking change, if I drop this if statement? Otherwise, I need a special treatment for those early returns in the benchmarking

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Oh yeah true: [a265987](https://github.com/KILTprotocol/kilt-node/pull/789/commits/a265987dbfef2f4b3ef9310f34625f7dbd9fc66a)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Ops: [7334d3b](https://github.com/KILTprotocol/kilt-node/pull/789/commits/7334d3b916518aba2cc9f4b7b474ed90665ecc07)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: I added some asserts in: [358f418](https://github.com/KILTprotocol/kilt-node/pull/789/commits/358f418d0dafb9d36a930edfeec5019aa970c6ae)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: added asserts in: [358f418](https://github.com/KILTprotocol/kilt-node/pull/789/commits/358f418d0dafb9d36a930edfeec5019aa970c6ae)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: This is quite challenging to verify. There is no trait available to access the state of an asset directly. However, I can introduce a storage alias to access the asset's storage and decode the result. The last byte of the encoded struct would reveal the state of the asset.

Direct access to the state is not possible because the properties are private

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: Added some asserts in: [358f418](https://github.com/KILTprotocol/kilt-node/pull/789/commits/358f418d0dafb9d36a930edfeec5019aa970c6ae)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: no clone, function calls, or conversations in tx: [44841bd](https://github.com/KILTprotocol/kilt-node/pull/789/commits/44841bdaa050c2179d0748ac59fbf9357845197f)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: [44841bd](https://github.com/KILTprotocol/kilt-node/pull/789/commits/44841bdaa050c2179d0748ac59fbf9357845197f)

scripts/run_benches_for_pallets.sh
Ad96el: fair point: https://github.com/KILTprotocol/kilt-node/pull/800

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: Can you double check that there's no other place where you mistakenly do this?

pallets/pallet-bonded-coins/src/benchmarking.rs
ntn-x2: But this would couple the benchmarking logic with the assets pallet, which is not necessarily true. If there's no way to verify this, just resolve this comment yourself.

scripts/run_benches_for_pallets.sh
Ad96el: Droped in: [8739367](https://github.com/KILTprotocol/kilt-node/pull/789/commits/87393679e7015a3bea258bf75433596e63e4e7d9)

scripts/run_benches_for_pallets.sh
Ad96el: I fmt the file with [shfmt](https://github.com/mvdan/sh#shfmt) in: [8739367](https://github.com/KILTprotocol/kilt-node/pull/789/commits/87393679e7015a3bea258bf75433596e63e4e7d9)

pallets/pallet-bonded-coins/Cargo.toml
```diff
@@ -47,13 +46,24 @@ runtime-benchmarks = [
 	"pallet-assets/runtime-benchmarks",
 	"pallet-balances/runtime-benchmarks",
 ]
+
 std = [
 	"frame-benchmarking/std",
```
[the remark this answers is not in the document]
Ad96el: Good point. Updated in: [3073c4d](https://github.com/KILTprotocol/kilt-node/pull/789/commits/3073c4daff258801be33d0438419936247b80a39)

I also checked the other pallets. What I see is that the runtime-benchmarks is not activated for frame-benchmarking: 

```
runtime-benchmarks = [
  "frame-benchmarking",
```

Is this a bug in the other pallets? I would have assumed, the feature is important. On substrate it is normally activated.

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: [470cad9](https://github.com/KILTprotocol/kilt-node/pull/789/commits/470cad95496101a3eb7873205c3196ebffd738fb)

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: Yeah, true… I can introduce another function in the benchmarkhelper, to check if a currency is on a destroying state. That would be generic enough against all possible pallets, handling fungibles

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I checked the other extrinsics. Seems fine from my POV

pallets/pallet-bonded-coins/src/benchmarking.rs
Ad96el: Like so: [03c19f0](https://github.com/KILTprotocol/kilt-node/pull/789/commits/03c19f008a7ecbe64c96f3108307225651a86034)

pallets/pallet-bonded-coins/Cargo.toml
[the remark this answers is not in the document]
ntn-x2: Apparently we also need to specify that with `dep:frame-benchmarking` to avoid creating a new feature for each optional dependency.

pallets/pallet-bonded-coins/src/benchmarking.rs
ntn-x2: No I personally think that is way overkill and risks to create a precedent where you require the trait to implement any test that you need in the benchmark. I would simply forget we had this conversation 😃

Ad96el: Was feeling the same, but wanted to provide a possible solution. I reverted the changes in: [3617907](https://github.com/KILTprotocol/kilt-node/pull/789/commits/361790783098b2c8697cfc669c3588c2cc6e31cd)

ntn-x2 (approved):
LGTM!

pallets/pallet-bonded-coins/Cargo.toml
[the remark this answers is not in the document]
Ad96el: > But it should be done in a different PR
yes thats for sure


> Apparently we also need to specify that with dep:frame-benchmarking to avoid creating a new feature for each optional dependency.

what do you mean exactly?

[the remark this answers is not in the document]
ntn-x2: As far as I understand from the cargo docs, this pallet should be compilable with the following command `cargo build --features frame-benchmarking`. That's because by default any optional dependency is also exposed as a feature that, when enabled, enabled the dependency. To avoid that, the dependency must be specified as `dep:<dependency>` in any of the existing features to block this behaviour. I haven't yet tested this myself, but it seems to be the case according to the docs.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -347,7 +364,7 @@ pub mod pallet {
 				&pool_id,
 				Some(PoolDetails::new(
 					who,
-					checked_curve,
+					checked_curve.clone(),
```
rflechtner, line 367: what has changed? why does this need to be cloned now?

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: it is indeed a breaking change, because this will lead to errors. But why don't you just use a weight like `Ok(Some(T::WeightInfo::refund_account(0)).into())` given that the amount of currencies up to this point does not matter at all?

rflechtner: the one below will also cause errors when removed, as per the comment included. Any reason you can't just use a weight of `Ok(Some(T::WeightInfo::refund_account(currency_count.to_owned())).into())` like for the final result? If we don't cover these cases in benchmarking at all (so we make sure there's always a balance and a collateral returned), we would simply overestimate the weight here, which shouldn't be an issue?

rflechtner: > it is indeed a breaking change, because this will lead to errors

Well actually it's already broken now because trying to unfreeze an account that does not exist will error. So we need to decide if trying to refund an account that does not exist should be treated as a no-op, which means fixing this, or if it should be treated as an error. Ultimately both work, you could just submit a batchAll if you wanted to ignore that some accounts in your list may not have needed refunding.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -728,22 +763,6 @@ pub mod pallet {
...
-					amount,
-					Provenance::Extant,
-				)
-				.into_result()
-				.is_err()
-			{
-				// Funds are burnt but the collateral received is not sufficient to be deposited
-				// to the account. This is tolerated as otherwise we could have edge cases where
-				// it's impossible to refund at least some accounts.
-				return Ok(());
```
rflechtner, line 744: this is a bigger problem, because we have two big edge cases:
1. There can be less collateral than bonded currency, which means that some could receive 0 refunds. Either we allow reaping these accounts (the way it currently works), or you need to refund other accounts first, hope that this has brought the total supply to 0, and if not, try refunding this account again.
2. Worse: You may be entitled to so little collateral that you cannot transfer it to your account due to minimum balance requirements. In that case your account cannot be refunded and the whole pool teardown is stuck.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Ah, yes, sorry. Was an artifact: [3b8b198](https://github.com/KILTprotocol/kilt-node/pull/789/commits/3b8b198cf9ec9ba04be197485bbbbe2a94b591c9)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I think, trying to refund an account that does not exist should throw an error. Feels semantically cleaner to me. But I have no strong opinion on that.

rflechtner: I'm also fine with it. You still have a way to do fire and forget if you opt for a batchAll

Ad96el: Added in: [2bced52](https://github.com/KILTprotocol/kilt-node/pull/789/commits/2bced5208c970b38a109a4c74063769d82407e16)

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: I reintroduced the check: [2bced52](https://github.com/KILTprotocol/kilt-node/pull/789/commits/2bced5208c970b38a109a4c74063769d82407e16)

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: 🤘

rflechtner (approved):
LGTM!

PR #791, opened 2024-11-07
test: mint and burn

rflechtner:
## re/ KILTProtocol/ticket#3671

Adds unit tests for burning and minting. Tests document _EXPECTED_ behaviour, which currently differs from the actual behaviour in a lot of places.

Currently failing:

- mint_into::mint_first_coin
  - Fails due to flooring of amount_to_mint during conversion to float; may be fixed by proper rounding, or ignored by changing the assertion to an `assert_eq_error_rate!` with error `1`.
- mint_into::mint_in_refunding_pool
  - Should give `PoolNotLive` error but yields `NoPermission` 
- mint_into::mint_large_quantity
  - Fails because we overflow when converting large amounts of collateral back to a u128. Can be fixed with conversion logic similar to the conversion to fixed. 
- mint_into::multiple_mints_vs_combined_mint
  - fails due to imprecision - not clear if this can be resolved 

- burn_into::burn_in_refunding_pool
  - Should give `PoolNotLive` error but yields `NoPermission` 
- burn_into::burn_large_quantity
  - Fails because we overflow when converting large amounts of collateral back to a u128. Can be fixed.
- burn_into::multiple_burns_vs_combined_burn
 - just like multiple vs single mint fails due to imprecision - not clear if this can be resolved 
- burn_into::multiple_mints_vs_combined_burn
  - Tries to burn more than was minted, yields `FundsUnavailable`. Must be fixed by improving rounding!
- burn_into::burn_with_frozen_balance
  - Fails because it tries to re-freeze the account after draining it
- burn_into::burn_to_other
  - Funds are burnt from the beneficiary instead of the caller, which is wrong  

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -850,6 +850,7 @@ pub mod pallet {
 				.ok_or(ArithmeticError::Overflow)?;
 
 			let real_costs = normalized_costs
+				// TODO: can easily overflow, causing test burn_large_quantity to fail
 				.checked_mul(CurveParameterTypeOf::<T>::from_num(collateral_denomination))
 				.ok_or(ArithmeticError::Overflow)?
 				// should never fail
--- a/pallets/pallet-bonded-coins/src/mock.rs
+++ b/pallets/pallet-bonded-coins/src/mock.rs
@@ -8,7 +8,7 @@ use frame_system::{EnsureRoot, EnsureSigned};
 use sp_core::U256;
 use sp_runtime::{
 	traits::{BlakeTwo256, IdentifyAccount, IdentityLookup, Verify},
-	BoundedVec, BuildStorage, MultiSignature,
+	BoundedVec, BuildStorage, MultiSignature, Permill,
 };
 use substrate_fixed::{
 	traits::{FixedSigned, FixedUnsigned},
@@ -44,6 +44,8 @@ pub(crate) const DEFAULT_COLLATERAL_CURRENCY_ID: AssetId = 0;
 pub(crate) const DEFAULT_COLLATERAL_DENOMINATION: u8 = 10;
 pub(crate) const DEFAULT_BONDED_DENOMINATION: u8 = 10;
 pub(crate) const ONE_HUNDRED_KILT: u128 = 100_000_000_000_000_000;
+
+pub(crate) const MAX_ERROR: Permill = Permill::from_perthousand(1);
 
 // helper functions
 pub fn assert_relative_eq<T: FixedSigned>(target: T, expected: T, epsilon: T) {
--- a/pallets/pallet-bonded-coins/src/mock.rs
+++ b/pallets/pallet-bonded-coins/src/mock.rs
@@ -8,7 +8,7 @@ use frame_system::{EnsureRoot, EnsureSigned};
 use sp_core::U256;
 use sp_runtime::{
 	traits::{BlakeTwo256, IdentifyAccount, IdentityLookup, Verify},
-	BoundedVec, BuildStorage, MultiSignature, PerThing,
+	BoundedVec, BuildStorage, MultiSignature,
 };
 use substrate_fixed::{
 	traits::{FixedSigned, FixedUnsigned},
@@ -53,18 +53,6 @@ pub fn assert_relative_eq<T: FixedSigned>(target: T, expected: T, epsilon: T) {
 		expected,
 		epsilon,
 		target
-	);
-}
-
-pub fn assert_balance_with_error_margin<T: PerThing>(target: u128, expected: u128, margin: T) {
-	let error_margin: u128 = margin.mul_ceil(expected);
-
-	assert!(
-		(target <= expected.saturating_add(error_margin)) && (target >= expected.saturating_sub(error_margin)),
-		"Expected {:?} to be in range {:?} +/- {:?}",
-		target,
-		expected,
-		error_margin
 	);
 }
 
--- a/pallets/pallet-bonded-coins/src/mock.rs
+++ b/pallets/pallet-bonded-coins/src/mock.rs
@@ -275,7 +275,7 @@ pub mod runtime {
 			.assimilate_storage(&mut storage)
 			.expect("assimilate should not fail");
 
-			let collateral_assets = self.collaterals.into_iter().map(|id| (id, ACCOUNT_99, false, 1));
+			let collateral_assets = self.collaterals.iter().map(|id| (*id, ACCOUNT_99, true, 1));
 
 			let all_assets: Vec<_> = self
 				.pools
@@ -306,6 +306,11 @@ pub mod runtime {
 							.map(|id| (*id, vec![], vec![], pool_details.denomination))
 							.collect::<Vec<(u32, Vec<u8>, Vec<u8>, u8)>>()
 					})
+					.chain(
+						self.collaterals
+							.into_iter()
+							.map(|id| (id, vec![], vec![], DEFAULT_COLLATERAL_DENOMINATION)),
+					)
 					.collect(),
 			}
 			.assimilate_storage(&mut storage)
--- a/pallets/pallet-bonded-coins/src/mock.rs
+++ b/pallets/pallet-bonded-coins/src/mock.rs
@@ -58,7 +58,7 @@ pub fn assert_relative_eq<T: FixedSigned>(target: T, expected: T, epsilon: T) {
 	);
 }
 
-pub(crate) fn collateral_at_supply(supply: u128) -> u128 {
+pub(crate) fn mocks_curve_get_collateral_at_supply(supply: u128) -> u128 {
 	let supply_u256 = U256::from(supply);
 	let sup_squared = supply_u256 * supply_u256;
 	// curve is f(x) = 4x + 3, resulting in f'(x) = 2x^2 + 3x.
--- a/pallets/pallet-bonded-coins/src/mock.rs
+++ b/pallets/pallet-bonded-coins/src/mock.rs
@@ -289,8 +289,8 @@ pub mod runtime {
 				.chain(collateral_assets)
 				.collect();
 
-			// NextAssetId is set to the maximum value of all collateral/bonded currency ids, plus one.
-			// If no currencies are created, it's set to 0.
+			// NextAssetId is set to the maximum value of all collateral/bonded currency
+			// ids, plus one. If no currencies are created, it's set to 0.
 			let next_asset_id = all_assets.iter().map(|(id, ..)| id).max().map_or(0, |id| id + 1);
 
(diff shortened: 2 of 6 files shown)


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -850,6 +850,7 @@ pub mod pallet {
 				.ok_or(ArithmeticError::Overflow)?;
 
 			let real_costs = normalized_costs
+				// TODO: can easily overflow, causing test burn_large_quantity to fail
```
rflechtner, line 853: This suffers from the same issues that convert_to_fixed suffered from: If we multiply then convert, this can easily overflow, but the other way around results in truncation. Should be solved just like the scaled conversion to fixed I'd say.

pallets/pallet-bonded-coins/src/mock.rs
```diff
@@ -277,7 +305,7 @@ pub mod runtime {
 			.assimilate_storage(&mut storage)
 			.expect("assimilate should not fail");
 
-			let collateral_assets = self.collaterals.into_iter().map(|id| (id, ACCOUNT_99, false, 1));
+			let collateral_assets = self.collaterals.iter().map(|id| (*id, ACCOUNT_99, true, 1));
```
rflechtner, line 308: we said collaterals should be sufficient, right?

pallets/pallet-bonded-coins/src/mock.rs
```diff
@@ -308,6 +336,11 @@ pub mod runtime {
 							.map(|id| (*id, vec![], vec![], pool_details.denomination))
 							.collect::<Vec<(u32, Vec<u8>, Vec<u8>, u8)>>()
 					})
+					.chain(
+						self.collaterals
+							.into_iter()
+							.map(|id| (id, vec![], vec![], DEFAULT_COLLATERAL_DENOMINATION)),
+					)
```
rflechtner, line 343: collateral denomination was not set

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
```diff
@@ -1 +1,705 @@
...
+
+			assert_eq!(
+				<Test as crate::Config>::Fungibles::total_balance(currency_id1, &ACCOUNT_00),
+				<Test as crate::Config>::Fungibles::total_balance(currency_id2, &ACCOUNT_00),
+			);
+
+			assert_eq!(
+				<Test as crate::Config>::CollateralCurrencies::total_balance(DEFAULT_COLLATERAL_CURRENCY_ID, &pool_id1),
+				<Test as crate::Config>::CollateralCurrencies::total_balance(DEFAULT_COLLATERAL_CURRENCY_ID, &pool_id2),
+			);
```
Ad96el, line 366: That really sucks :D

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
```diff
@@ -1 +1,705 @@
...
+			// check that we can mint again into a frozen account
+			assert_ok!(BondingPallet::mint_into(
+				origin,
+				pool_id,
+				0,
+				ACCOUNT_00,
+				amount_to_mint,
+				initial_collateral,
+				1
+			));
```
Ad96el, line 432: can we also check, if the reducible balance is still zero?

pallets/pallet-bonded-coins/src/mock.rs
```diff
@@ -45,15 +46,30 @@ pub(crate) const DEFAULT_BONDED_DENOMINATION: u8 = 10;
...
 		(target - expected).abs() <= epsilon,
-		"Expected {:?} but got {:?}",
+		"Expected {:?} +/- {:?} but got {:?}",
 		expected,
+		epsilon,
 		target
 	);
 }
 
+pub(crate) fn collateral_at_supply(supply: u128) -> u128 {
```
Ad96el, line 59: I would get rid of this function. I explained why in:

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
```diff
@@ -1 +1,726 @@
...
+				origin,
+				pool_id,
+				0,
+				ACCOUNT_00,
+				amount_to_burn,
+				expected_price - 1, // rounding down may be happening in the conversion to fixed
+				1
+			));
+
+			assert_eq_error_rate!(
```
Ad96el, line 66: Thats nice :grimacing: I would prob use the same macro for the curves tests.

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
```diff
@@ -1 +1,705 @@
...
+				pool_id.clone(),
+				0,
+				ACCOUNT_00,
+				amount_to_mint,
+				expected_price,
+				1
+			));
+
+			assert_eq_error_rate!(
+				<Test as crate::Config>::CollateralCurrencies::total_balance(
```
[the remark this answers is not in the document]
rflechtner: Yes but I chose to interact with it everywhere via the fungibles traits to maintain loose coupling - it doesn't really matter but it's also how the pallet interacts with it.

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
```diff
@@ -1 +1,705 @@
...
+
+#[test]
+fn mint_first_coin() {
+	let pool_id = calculate_pool_id(&[DEFAULT_BONDED_CURRENCY_ID]);
+
+	let curve = get_linear_bonding_curve();
+
+	let initial_collateral = 100u128;
+	let amount_to_mint = 1u128;
+	let expected_price = collateral_at_supply(amount_to_mint);
```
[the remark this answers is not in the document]
rflechtner: Good question, but at least in the tests it may make sense to check that we are not toooooo far off? Just to see that it performs close to expectations?

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
```diff
@@ -1 +1,705 @@
...
+use sp_runtime::{assert_eq_error_rate, ArithmeticError, Permill, TokenError};
+
+use crate::{
+	curves::{polynomial::PolynomialParameters, Curve},
+	mock::{runtime::*, *},
+	types::{Locks, PoolStatus},
+	Error,
+};
+
+const MAX_ERROR: Permill = Permill::from_perthousand(1);
```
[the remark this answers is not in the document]
rflechtner: makes sense

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
rflechtner: yeah..... we can see what we can do with rounding, but it's an issue

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
rflechtner: it does not seem to work for the Fixed type for some reason :(

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
```diff
@@ -1 +1,726 @@
...
+		.build()
+		.execute_with(|| {
+			let origin = RawOrigin::Signed(ACCOUNT_00).into();
+
+			assert_err!(
+				BondingPallet::burn_into(origin, pool_id, 0, ACCOUNT_00, 2, 0, 1), // Attempt to burn 2 units
+				TokenError::FundsUnavailable
+			);
+		});
+}
```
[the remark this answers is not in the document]
rflechtner: oh yeah that was an issue right?

pallets/pallet-bonded-coins/src/mock.rs
rflechtner: It always needs to be used with `assert_eq_error_rate!` of course

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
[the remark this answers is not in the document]
rflechtner: 8435ab5e368a89c847b251b1f45135116bdc43e4

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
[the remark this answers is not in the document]
rflechtner: 2bcc31db8788cd97c6bbeba3d8797e7cf2fc487b

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
rflechtner: 2bcc31db8788cd97c6bbeba3d8797e7cf2fc487b

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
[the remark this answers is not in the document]
Ad96el: I think we also need another tx `burn_from` or not? If the pool is permissioned, and the pool creator should delete some funds, there is not really a way right now.

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
Ad96el: Oh, okay. That's unfortunate.

pallets/pallet-bonded-coins/src/tests/transactions/burn_into.rs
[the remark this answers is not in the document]
Ad96el: Or we change the implementation. If the pool is locked, the bonded currencies are burned from the beneficiary, Otherwise by the caller. But not sure, if it is a good idea to overload the tx.

pallets/pallet-bonded-coins/src/tests/transactions/mint_into.rs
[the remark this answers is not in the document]
Ad96el: Yeah, it is the same. It would just reduce the lines and the complexity in the tests. 
I would prefer the change, but I will not stay in the way if you insist. Feel free to resolve the conversation.

pallets/pallet-bonded-coins/src/mock.rs
Ad96el: Could we then maybe rename the function? It is not really clear, that this function is calculating the collateral for the default linear bonding curve. Or make it part of the return type for the function, where the curve is created, but this is maybe a bit overengineered.

rflechtner: 41c70f501425b49dc750349b8b6f62084569c5c2

PR #794, opened 2024-11-11
fix: denomination arithmetic

rflechtner:
Adresses issues with precision stemming from integer division in `convert_to_fixed` and suggests a strategy to mitigate overflow problems on `refund_account`.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/curves/mod.rs
+++ b/pallets/pallet-bonded-coins/src/curves/mod.rs
@@ -111,6 +111,31 @@ pub(crate) fn convert_to_fixed<T: Config>(
 	let decimals = 10u128
 		.checked_pow(u32::from(denomination))
 		.ok_or(ArithmeticError::Overflow)?;
-	let scaled_x = x.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
-	scaled_x.checked_to_fixed().ok_or(ArithmeticError::Overflow)
+
+	// Divide `x` to separate the quotient and remainder
+	let quotient = x.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
+	let remainder = x.checked_rem_euclid(decimals).ok_or(ArithmeticError::DivisionByZero)?;
+
+	// Convert quotient to fixed-point
+	// This can overflow if capacity(u128) / 10^denomination > capacity(fixed)
+	let fixed_quotient: CurveParameterTypeOf<T> = quotient.checked_to_fixed().ok_or(ArithmeticError::Overflow)?;
+
+	// we can skip handling the remainder if it's 0
+	if remainder == 0 {
+		Ok(fixed_quotient)
+	} else {
+		// Convert remainder to fixed-point and scale it down by `decimals`
+		let fixed_remainder = remainder
+			// This can overflow if 10^denomination > capacity(float)
+			.checked_to_fixed::<CurveParameterTypeOf<T>>()
+			.ok_or(ArithmeticError::Overflow)?
+			// This WILL overflow if 10^denomination > capacity(float)
+			.checked_div(decimals.checked_to_fixed().ok_or(ArithmeticError::Overflow)?)
+			.ok_or(ArithmeticError::DivisionByZero)?;
+		// Combine both parts
+		fixed_quotient
+			// This WILL overflow if x / 10^denomination > capacity(float)
+			.checked_add(fixed_remainder)
+			.ok_or(ArithmeticError::Overflow)
+	}
 }
--- a/pallets/pallet-bonded-coins/src/curves/mod.rs
+++ b/pallets/pallet-bonded-coins/src/curves/mod.rs
@@ -5,6 +5,7 @@ pub(crate) mod square_root;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_arithmetic::ArithmeticError;
+use sp_runtime::traits::CheckedConversion;
 use sp_std::ops::{AddAssign, BitOrAssign, ShlAssign};
 use substrate_fixed::traits::{Fixed, FixedSigned, ToFixed};
 
@@ -104,10 +105,10 @@ fn calculate_accumulated_passive_issuance<Balance: Fixed>(passive_issuance: &[Ba
 		.fold(Balance::from_num(0), |sum, x| sum.saturating_add(*x))
 }
 
-pub(crate) fn convert_to_fixed<T: Config>(
-	x: u128,
-	denomination: u8,
-) -> Result<CurveParameterTypeOf<T>, ArithmeticError> {
+pub(crate) fn convert_to_fixed<T: Config>(x: u128, denomination: u8) -> Result<CurveParameterTypeOf<T>, ArithmeticError>
+where
+	<CurveParameterTypeOf<T> as Fixed>::Bits: TryFrom<u128>,
+{
 	let decimals = 10u128
 		.checked_pow(u32::from(denomination))
 		.ok_or(ArithmeticError::Overflow)?;
@@ -130,11 +131,13 @@ pub(crate) fn convert_to_fixed<T: Config>(
 	if remainder > 0 {
 		// Convert remainder to fixed-point and scale it down by `decimals`,
 		let fractional = remainder
-			// This can overflow if 10^denomination > capacity(float)
+			// This can overflow if 10^denomination > capacity(fixed)
 			.checked_to_fixed::<CurveParameterTypeOf<T>>()
 			.ok_or(ArithmeticError::Overflow)?
-			// This WILL overflow if 10^denomination > capacity(float)
-			.checked_div(decimals.checked_to_fixed().ok_or(ArithmeticError::Overflow)?)
+			// This would overflow if 10^denomination exceeds the capacity of the _underlying integer_ of the fixed
+			// (e.g., i128 for an I75F53). However, there is no denomination that is representable in a u128 but not in
+			// an i128.
+			.checked_div_int(decimals.checked_into().ok_or(ArithmeticError::Overflow)?)
 			.ok_or(ArithmeticError::DivisionByZero)?;
 
 		// Combine both parts
--- a/pallets/pallet-bonded-coins/src/curves/mod.rs
+++ b/pallets/pallet-bonded-coins/src/curves/mod.rs
@@ -112,30 +112,38 @@ pub(crate) fn convert_to_fixed<T: Config>(
 		.checked_pow(u32::from(denomination))
 		.ok_or(ArithmeticError::Overflow)?;
 
-	// Divide `x` to separate the quotient and remainder
-	let quotient = x.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
+	// Scale down x by the denomination using integer division, truncating any
+	// fractional parts of the result for now. Will overflow if down-scaling is not
+	// sufficient to bring x to a range representable in the fixed point format.
+	// This can happen if capacity(u128) / 10^denomination > capacity(fixed).
+	let mut scaled_x: CurveParameterTypeOf<T> = x
+		.checked_div(decimals)
+		.ok_or(ArithmeticError::DivisionByZero)?
+		.checked_to_fixed()
+		.ok_or(ArithmeticError::Overflow)?;
+
+	// Next we handle the remainder of the division
 	let remainder = x.checked_rem_euclid(decimals).ok_or(ArithmeticError::DivisionByZero)?;
 
-	// Convert quotient to fixed-point
-	// This can overflow if capacity(u128) / 10^denomination > capacity(fixed)
-	let fixed_quotient: CurveParameterTypeOf<T> = quotient.checked_to_fixed().ok_or(ArithmeticError::Overflow)?;
-
-	// we can skip handling the remainder if it's 0
-	if remainder == 0 {
-		Ok(fixed_quotient)
-	} else {
-		// Convert remainder to fixed-point and scale it down by `decimals`
-		let fixed_remainder = remainder
+	// If the remainder is not 0, convert to fixed, scale it down and add the
+	// resulting fractional part to the previous result
+	if remainder > 0 {
+		// Convert remainder to fixed-point and scale it down by `decimals`,
+		let fractional = remainder
 			// This can overflow if 10^denomination > capacity(float)
 			.checked_to_fixed::<CurveParameterTypeOf<T>>()
 			.ok_or(ArithmeticError::Overflow)?
 			// This WILL overflow if 10^denomination > capacity(float)
 			.checked_div(decimals.checked_to_fixed().ok_or(ArithmeticError::Overflow)?)
 			.ok_or(ArithmeticError::DivisionByZero)?;
+
 		// Combine both parts
-		fixed_quotient
-			// This WILL overflow if x / 10^denomination > capacity(float)
-			.checked_add(fixed_remainder)
-			.ok_or(ArithmeticError::Overflow)
-	}
+		scaled_x = scaled_x
+			// Overflow is theoretically impossible as we are adding a number < 1 to a fixed point where the fractional
+			// part is 0.
+			.checked_add(fractional)
+			.ok_or(ArithmeticError::Overflow)?;
+	};
+
+	Ok(scaled_x)
 }
--- a/pallets/pallet-bonded-coins/src/curves/mod.rs
+++ b/pallets/pallet-bonded-coins/src/curves/mod.rs
@@ -5,6 +5,7 @@ pub(crate) mod square_root;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_arithmetic::ArithmeticError;
+use sp_core::U256;
 use sp_runtime::traits::CheckedConversion;
 use sp_std::ops::{AddAssign, BitOrAssign, ShlAssign};
 use substrate_fixed::traits::{Fixed, FixedSigned, ToFixed};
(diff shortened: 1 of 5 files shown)


pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -111,6 +111,31 @@ pub(crate) fn convert_to_fixed<T: Config>(
 	let decimals = 10u128
 		.checked_pow(u32::from(denomination))
 		.ok_or(ArithmeticError::Overflow)?;
-	let scaled_x = x.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
```
rflechtner, line 114: This was dropping everything that is not a multiple of the denomination - that can't be the point of conversion to fixed, right @Ad96el ?

pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -111,6 +111,31 @@ pub(crate) fn convert_to_fixed<T: Config>(
...
-	scaled_x.checked_to_fixed().ok_or(ArithmeticError::Overflow)
+
+	// Divide `x` to separate the quotient and remainder
+	let quotient = x.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
+	let remainder = x.checked_rem_euclid(decimals).ok_or(ArithmeticError::DivisionByZero)?;
```
rflechtner, line 117: To avoid truncating amounts that are smaller than a multiple of the denomination, we need to convert to fixed first, then divide by the denomination. This of course can easily result in overflow, which defies the purpose of scaling by denomination.
This is why I'm adopting a combined strategy: On one leg, perform scaling as before, truncating fractionals, and on the other leg, take the remainder of the integer division, convert it to fixed, then divide by the denomination. This gives us the fractional part that was previously dropped, which we can now add to the result.

pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -111,6 +111,31 @@ pub(crate) fn convert_to_fixed<T: Config>(
...
+	if remainder == 0 {
+		Ok(fixed_quotient)
+	} else {
+		// Convert remainder to fixed-point and scale it down by `decimals`
+		let fixed_remainder = remainder
+			// This can overflow if 10^denomination > capacity(float)
+			.checked_to_fixed::<CurveParameterTypeOf<T>>()
+			.ok_or(ArithmeticError::Overflow)?
+			// This WILL overflow if 10^denomination > capacity(float)
+			.checked_div(decimals.checked_to_fixed().ok_or(ArithmeticError::Overflow)?)
```
rflechtner, line 133: The solution comes at the cost of one additional case where overflow can happen:

- Previously, overflow could happen where `capacity(balance) / 10^denomination > capacity(fixed)`, which for a u128 balance (capacity ~3.4e+38) and a I75F53 (capacity ~1.8e+22) meant that any denomination >= 17 is safe.
- With the current solution, overflow __will__ occur where `10^denomination > capacity(float)`, which for an I75F53 means that we have a hard cap at a denomination of 22.

We can possibly soften the hard cap, e.g. by iteratively dividing the remainder by 10 (or some other multiplicative component of `decimals` which can be represented), but denominations larger than the soft cap would progressively increase the probability of an overflow occurring. 

For example, with an I75F53 and a denomination of 23, any amount `~1.8e+22 < x < e+23` would result in a remainder that cannot be represented and thus in an overflow error.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -681,12 +681,35 @@ pub mod pallet {
...
+				// An overflow is also theoretically impossible, as burnt <= sum_of_issuances
+				.ok_or(Error::<T>::Internal)?;
+
+			// for the remainder, we multiply first and divide next
+			let remainder_part = total_collateral_issuance
+				.checked_rem(&sum_of_issuances)
+				.ok_or(Error::<T>::Internal)? // Should only fail on sum_of_issuances == 0, same as above
+				.checked_mul(&burnt)
+				// This can indeed overflow if (sum_of_issuances - 1) * burnt > capacity(u128)
+				.ok_or(ArithmeticError::Overflow)?
```
rflechtner, line 703: Very large sum_of_issuances make large remainders more likely, which can in turn result in a product that too large.
We can accept that refunds would sometimes fail, which would require accounts with lower balances to refund first (after which the situation may have changed).
Or we can look for fallback strategies, such as:
- saturating multiplication, which could mean that people receive only a fraction of the funds they would be entitled to
- scaling, which sacrifices precision
- partial refunds, where we only burn the amount of funds for which we can properly compute the collateral to be refunded, and leave the rest untouched.

pallets/pallet-bonded-coins/src/tests/curves/arithmetic.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+#[test]
+fn test_convert_to_fixed_too_large_denomination() {
+	let x = u128::MAX; // around 3.4e+38
+	let denomination = 30u8; // I75F53 should handle around 1.8e+22, this can't be represented
+
+	let result = convert_to_fixed::<Test>(x, denomination);
+	assert_ok!(result);
+}
```
rflechtner, line 110: this test fails, as a reminder that the proposed solution is still imperfect

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -681,12 +681,35 @@ pub mod pallet {
...
-			let amount = burnt
-				.checked_mul(&total_collateral_issuance)
-				.ok_or(ArithmeticError::Overflow)? // TODO: do we need a fallback if this fails?
+			defensive_assert!(
+				sum_of_issuances >= burnt,
+				"burnt amount exceeds the total supply of all bonded currencies"
+			);
+
+			// divide first to avoid overflow, remainder will be handled seperately
+			let divisible_part = total_collateral_issuance
```
rflechtner, line 690: I realised that the approach of handling remainders separately can also improve the calculations of how much collateral is refunded, preventing overflow in many (but not all) cases.

pallets/pallet-bonded-coins/src/curves/mod.rs
Ad96el: Yes, this is a bug. Initially, I have first converted X into a fixed and then divided it by the denomination.

pallets/pallet-bonded-coins/src/curves/mod.rs
Ad96el: I kinda have the feeling, this is a bit overengineered: maybe I am wrong. The additional bits gained by the second step, are anyway dropped in the combination step, if the denomination is bigger than the fractional part of the Float. 

> This of course can easily result in overflow

That's right, but also quite an edge case, right? We kinda have some numeric bounds. A user can also provide an amount, which would not fit into an u128, then we have the same problem. I would argue the possible values are big enough

rflechtner: > I kinda have the feeling, this is a bit overengineered: maybe I am wrong.

What do you not agree with, the problem analysis or the proposed solution? 

__On the problem:__
I don't actually think we disagree here, but just to make sure: I do think it needs to be addressed, because it effectively means you cannot use sub-units at all; everything is rounded down. You cannot mint 1.4 xyz-coins - you'd be minting 1. You cannot burn 0.5 - this would just error. If there is any supply on any of the currencies that's below the denomination, it is going to be completely ignored in the calculations.

If that is what we want, _why even have a denomination at all_? If units are non-divisible, we may just assume a denomination of 0 everywhere. This would avoid a number of nasty bugs, for example the fact that you currently pay for the rounded down amount (e.g., 1 xyz), but what's minted is the exact amount you specified (1.4 xyz).

__On the solution:__
Is your preference simply to just convert to fixed first, then divide by the denomination, and accept the potential to overflow? Are you saying that trying to solve the overflow issue is overkill? 
In that case, we could handle amounts of up to ~1.8e+22. For a denomination of 10 that seems enough(~1.8 Trillion), for denomination of 15 that would mean we'd overflow at around 18 Million.

> I would argue the possible values are big enough

Possibly, I wouldn't really know how to make a judgement on this. If our balances would be u64 values, certainly; given that they are u128 we'd be wasting a ton of space that we can't effectively make use of. 

Still, if we accept these limits, we somehow need to enforce them, meaning that no bonded currency can have a supply > 2^74 (capacity of the fixed type we're using). Right now there is nothing preventing you from minting coins until the pool is past this limit, after which every attempt to mint or burn will fail and it's stuck completely. I wouldn't call that an edge case, but rather a nasty-ass bug.

>  A user can also provide an amount, which would not fit into an u128, then we have the same problem.

They actually can't; any transaction where the `amount_to_{mint,burn}` parameter is not exactly 128 bits long is invalid and will never be included in a block.

> The additional bits gained by the second step, are anyway dropped in the combination step, if the denomination is bigger than the fractional part of the Float.

I don't understand this at all, can you elaborate? And you mean Fixed, right? Right now we're effectively working with integers, and the fractional part is always 0, with my proposed there would be a non-zero fractional part, so I don't exactly get how these "extra bits are being dropped"?

A quick example is this test case:

```rust
 #[test]
 fn test_convert_to_fixed_with_remainder() {
 	let x = 1050u128;
 	let denomination = 2u8; // 10^2 = 100

 	let result = convert_to_fixed::<Test>(x, denomination).unwrap();
 	let expected = Float::from_num(10.5); // 1050 / 100 = 10.5

 	assert_eq!(result, expected);
 }
```

Before my change, this would error, as the function returned `10`; now it returns the correct result, `10.5`.

Ad96el: **On the problem**

I totally agree with you. This is a bug which we have to address. 0.5 UNITs have to be representable, which was also the case in the previous implementation (I hope). No discussion. 

**On the Solution** 

We should address all overflows. But I personally think, it is enough, to mint 18m coins for a denomination of 15 in a single step - if a user wants to mint more, he needs to provide a smaller denomination. The selection of the denomination should be an active decision, based on the user's need. Currently, nothing is stopping the user to provide a denomination of 100. It does not really make sense, but he can do that. 


> If our balances would be u64 values, certainly; given that they are u128 we'd be wasting a ton of space that we can't effectively make use of.

I don't really get this point.

> after which every attempt to mint or burn will fail and it's stuck completely.

That's not true. Every coin minted is also burnable. The user is not stuck, he just hit the numeric bounds for his pool, which he has to consider in the pool creation. 

Maybe instead of calculating the normalized amount of an u128, the tx should directly take the normalized amount. Then we have no issues, we are faster, and the user really gets what he wants (no rounding at all).

Ad96el: > I don't understand this at all, can you elaborate? And you mean Fixed, right? 

Yes, I mean fixed. By the current Float Implementation, we have 16 digits for the frac part. If the denomination is 17 for instance, we would drop the LSB

rflechtner: > 0.5 UNITs have to be representable, which was also the case in the previous implementation (I hope).

They weren't, that's why this PR exists. I couldn't get unit tests to pass because of that. 


> We should address all overflows. But I personally think, it is enough, to mint 18m coins for a denomination of 15 in a single step - if a user wants to mint more, he needs to provide a smaller denomination.

Agree, but unfortunately that's not the real problem we're facing (see below)


> > If our balances would be u64 values, certainly; given that they are u128 we'd be wasting a ton of space that we can't effectively make use of.

The supply that any currency may have before shit hits the fan is capped at 2^74 (see below). So we have 74 useable bits out of 128. A u64 balance is closer to what we can use, and would also naturally prevent balances to move into value ranges that break everything. 



> Every coin minted is also burnable. The user is not stuck, he just hit the numeric bounds for his pool, which he has to consider in the pool creation. 

Take a look at https://github.com/KILTprotocol/kilt-node/blob/pallet_bonded_coins/pallets/pallet-bonded-coins/src/lib.rs#L834. This is invoke during mint on the balance before mint, but not after. 
The result: the balance can increase to the point where this function can't handle it anymore, but the effects will only be seen on the next burn/mint, which will then fail regardless of the amount selected.  

> Maybe instead of calculating the normalized amount of an u128, the tx should directly take the normalized amount. Then we have no issues, we are faster, and the user really gets what he wants (no rounding at all). 

That's giving up on denominations. If there is not way to mint or burn 0.5 units, why even represent them?

pallets/pallet-bonded-coins/src/curves/mod.rs
rflechtner: Softened the cap on https://github.com/KILTprotocol/kilt-node/pull/794/commits/694334c178b2082d19b0f1e27a703585b9f865ae, which I think is a better way of scaling a fixed by an integer.
Still, I introduced a constant to limit allowable denominations on e94104e29ba16db3f7626488de705a92c00068b2.

pallets/pallet-bonded-coins/src/tests/curves/arithmetic.rs
rflechtner: Due to 694334c178b2082d19b0f1e27a703585b9f865ae this test now passes; but not for u128::MAX, which results in an overflowing remainder.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -129,11 +129,16 @@ pub mod pallet {
 		/// The maximum number of currencies allowed for a single pool.
 		#[pallet::constant]
 		type MaxCurrencies: Get<u32>;
-		/// The deposit required for each bonded currency.
 
 		#[pallet::constant]
 		type MaxStringLength: Get<u32>;
 
+		/// The maximum denomination that bonded currencies can use. To ensure
+		/// safe operation, this should be set so that 10^MaxDenomination <
```
ntn-x2, line 137: If you want to make sure that only these values are specified here, you can add an integrity test (available with the `try-runtime` feature) that only checks the values of the `Config` implementations for any given runtime, according to the specified rules. You can check an example here: https://github.com/paritytech/polkadot-sdk/blob/a875f1803253e469288f1a905e73861fb600bb28/polkadot/runtime/parachains/src/configuration.rs#L1258.

ntn-x2: This is to protect against malformed runtime-injected values inside a config trait, not too much about user-provided values.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -681,12 +688,35 @@ pub mod pallet {
...
+				"burnt amount exceeds the total supply of all bonded currencies"
+			);
+
+			// divide first to avoid overflow, remainder will be handled seperately
+			let divisible_part = total_collateral_issuance
+				.checked_div(&sum_of_issuances)
+				.ok_or(Error::<T>::Internal)? // because sum_of_issuances >= burnt > 0, this is theoretically impossible
+				.checked_mul(&burnt)
+				// An overflow is also theoretically impossible, as burnt <= sum_of_issuances
+				.ok_or(Error::<T>::Internal)?;
```
ntn-x2, line 702: `Internal` errors without an associated log is pretty much useless. Not sure if logging is supposed to be part of a separate PR or not, so I'll just mention it here.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -256,7 +261,7 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T>
 	where
-		<CurveParameterTypeOf<T> as Fixed>::Bits: Copy + ToFixed + AddAssign + BitOrAssign + ShlAssign,
+		<CurveParameterTypeOf<T> as Fixed>::Bits: Copy + ToFixed + AddAssign + BitOrAssign + ShlAssign + TryFrom<u128>,
```
ntn-x2, line 264: Where is the `TryFrom<u128>` trait used in the new code? I could not see it from the diff.

rflechtner: here: https://github.com/KILTprotocol/kilt-node/blob/e94104e29ba16db3f7626488de705a92c00068b2/pallets/pallet-bonded-coins/src/curves/mod.rs#L140

On this line, the u128 is first converted to the type underlying the fixed (e.g., an i128 in case of an I75U53) and then used for division.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: I've got the `defensive_assert!` which should do some logging in production builds, but other than that we haven't really set up logging yet.

ntn-x2: Ok then this is a thing to tackle when logging is added, which should happen at the very least any time an `Internal` error is thrown.

ntn-x2 (approved):

Ad96el (approved):

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: The new solution does not really put a limit on the maximum denomination - as long as `10^denomination` can be represented in a U256 - but if the denomination is larger than the precision of the fixed, we will experience truncation of balance amounts during scaling, which I documented here.

rflechtner: I also added an integrity test to check this, as suggested: 3a2306b615b0bab8fff0a07d5542e5e501c52287

pallets/pallet-bonded-coins/src/curves/mod.rs
rflechtner: Since 08b71c4c8966febb0c17df070b7b3e941f77115b we no longer have new cases of overflow introduced with this fix.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: With https://github.com/KILTprotocol/kilt-node/commit/08b71c4c8966febb0c17df070b7b3e941f77115b overflow issues are now a thing of the past.

pallets/pallet-bonded-coins/src/tests/curves/arithmetic.rs
rflechtner: Since https://github.com/KILTprotocol/kilt-node/commit/08b71c4c8966febb0c17df070b7b3e941f77115b large denominations do not result in an overflow any more. Note that of course truncation can still happen with large denominations if the fixed type is not able to offer enough digits behind the comma.

PR #796, opened 2024-11-13
refactor: use U256 for infallible scaling

rflechtner:
I realised that scaling may be done more elegantly and less likely to overflow by using a U256 to implement what essentially amounts to direct division on a fixed point number with >= 128 integer bits. This way the only overflow we can get is when the result (the scaled value) is too large to fit into the target type. 

In other words, the scaling logic is infallible, with the only error case being that the scaling factor is too small, so that the result cannot be represented.

I wanted to ask for your opinion on this solution before I apply it to my pending PR.    

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/curves/mod.rs
+++ b/pallets/pallet-bonded-coins/src/curves/mod.rs
@@ -5,6 +5,7 @@ pub(crate) mod square_root;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_arithmetic::ArithmeticError;
+use sp_core::U256;
 use sp_runtime::traits::CheckedConversion;
 use sp_std::ops::{AddAssign, BitOrAssign, ShlAssign};
 use substrate_fixed::traits::{Fixed, FixedSigned, ToFixed};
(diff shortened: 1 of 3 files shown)


pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -107,46 +108,26 @@ fn calculate_accumulated_passive_issuance<Balance: Fixed>(passive_issuance: &[Ba
...
-			.checked_add(fractional)
-			.ok_or(ArithmeticError::Overflow)?;
-	};
-
-	Ok(scaled_x)
+	// Convert to U256 so we have enough bits to perform lossless scaling.
+	let mut x_u256 = U256::from(x);
+	// Shift left to produce the representation that our fixed type would have (but
+	// with extra integer bits that would potentially not fit in the fixed type).
+	x_u256.shl_assign(CurveParameterTypeOf::<T>::frac_nbits());
```
rflechtner, line 120: Any idea if this can panic? I didn't find a `checked_` variant of shl, so probably not?

pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -107,46 +108,26 @@ fn calculate_accumulated_passive_issuance<Balance: Fixed>(passive_issuance: &[Ba
...
+	// Perform division. Due to the shift the precision/truncation is identical to
+	// division on the fixed type.
+	x_u256 = x_u256.checked_div(decimals).ok_or(ArithmeticError::DivisionByZero)?;
```
rflechtner, line 123: I'm realising this error case is also a theoretical impossibility, because even `10^0 = 1`

pallets/pallet-bonded-coins/src/curves/mod.rs
```diff
@@ -107,46 +108,26 @@ fn calculate_accumulated_passive_issuance<Balance: Fixed>(passive_issuance: &[Ba
 
 pub(crate) fn convert_to_fixed<T: Config>(x: u128, denomination: u8) -> Result<CurveParameterTypeOf<T>, ArithmeticError>
 where
-	<CurveParameterTypeOf<T> as Fixed>::Bits: TryFrom<u128>,
+	<CurveParameterTypeOf<T> as Fixed>::Bits: TryFrom<U256>,
```
rflechtner, line 111: We could in principle make this more generic by not fixing the type of x (an `Into<U256>` trait bound or similar is enough) and by making the type of the big integer configurable in the runtime, so you can match it to the chosen balance and fixed types.

The big integer only needs to be the size of the balance type plus the size of the fractional bits of the fixed, so for a `u64` balance and a `I75F53` fixed a `u128` would be sufficient for lossless conversion.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -261,7 +261,8 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T>
 	where
-		<CurveParameterTypeOf<T> as Fixed>::Bits: Copy + ToFixed + AddAssign + BitOrAssign + ShlAssign + TryFrom<u128>,
+		<CurveParameterTypeOf<T> as Fixed>::Bits: Copy + ToFixed + AddAssign + BitOrAssign + ShlAssign + TryFrom<U256>,
+		CollateralCurrenciesBalanceOf<T>: Into<U256> + TryFrom<U256>,
```
rflechtner, line 265: Using bigger integers is also a better fix for the overflow issues we had when computing the refund amounts. This second trait bound is for this fix.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -684,38 +686,25 @@ pub mod pallet {
 			let sum_of_issuances = pool_details
 				.bonded_currencies
 				.into_iter()
-				.fold(FungiblesBalanceOf::<T>::zero(), |sum, id| {
-					sum.saturating_add(T::Fungibles::total_issuance(id))
+				.fold(U256::from(0), |sum, id| {
+					sum.saturating_add(T::Fungibles::total_issuance(id).into())
```
rflechtner, line 690: this can now also no longer overflow

pallets/pallet-bonded-coins/src/curves/mod.rs
Ad96el: I think using the unchecked variant will give use some lint warnings. I think the best error is internal with a log message.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: We can also make this type configurable, it just needs to have twice the size of the balance type.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -684,38 +686,25 @@ pub mod pallet {
...
 				});
 
 			defensive_assert!(
 				sum_of_issuances >= burnt,
 				"burnt amount exceeds the total supply of all bonded currencies"
 			);
```
Ad96el, line 696: This can panic or not?

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: No strong opinion on this. I don’t think any other project is using a balance type larger than u128.

pallets/pallet-bonded-coins/src/curves/mod.rs
Ad96el: The macro is implementing the trait: 

```
		impl<T> $crate::core_::ops::ShlAssign<T> for $name where T: Into<$name> {
			fn shl_assign(&mut self, shift: T) {
				*self = *self << shift;
			}
		}
```

the `<<` operation can [panic](https://users.rust-lang.org/t/intentionally-overflow-on-shift/11859)

rflechtner: Ok but the panic would occur if the shift is larger than the size of the number, right? Which would be an insane misconfiguration. Should we just prevent against misconfiguration the way @ntn-x2 suggested [here](https://github.com/KILTprotocol/kilt-node/pull/794#discussion_r1838198725) and not check this at runtime?

ntn-x2: I am against removing the checks from the runtime. Anything that the type system does not guarantee against should be checked, and having it in the runtime does not imply removing the static tests for the config itself.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: Good thing is you can also configure it to be smaller, avoiding blowing it up to a U256 when it's completely unnecessary, as it would be if your balance is just a u64 (where a u128 would be enough)

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: this does not panic in production - it does though in non-optimised builds

pallets/pallet-bonded-coins/src/curves/mod.rs
rflechtner: Actually as long as we don't have this configurable this cannot panic, because the U256 is large enough to work with any Fixed type that exists

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: Added a TODO in cffd4bd2ae5a18d02722e18abdd6c0d14dc497f8

pallets/pallet-bonded-coins/src/curves/mod.rs
rflechtner: Added a TODO in cffd4bd2ae5a18d02722e18abdd6c0d14dc497f8

Ad96el (approved):

PR #799, opened 2024-11-15
fix: change traits

Ad96el:
We need to make the traits generic against the pallet-asset instance


rflechtner (approved):
LGTM. Are the original traits also generic?

PR #803, opened 2024-11-20
fix: mint into and burn into with min operation balance

Ad96el:
Since the funds are scaled down, we encountered a strange situation where a user could mint or burn funds without returning or taking any collateral.

For example, if a user wants to mint a single unit of a bonded currency, the amount to mint is scaled down based on the denomination specified in the bonding curve. With a denomination of 15, the provided value would be 1 * 10^-15. Such small values can cause side effects, as x^2 of 1 * 10^-15 results in 1 * 10^-30, which underflows and becomes zero.

This allowed users to mint bonded coins without providing any collateral. This behavior is now prevented through two mechanisms:

1. The calculated collateral must always be greater than 0.  
2. The pool creator must specify a `min_operation_balance` to enforce a minimum amount for minting or burning.

I think setting the min_operation_balance in an atomic tx is not a vise idea. This parameter should not be changeable.

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -269,6 +269,11 @@ pub mod pallet {
 		InvalidInput,
 		Internal,
 		Slippage,
+		/// The requested Operation could not be performed due to insufficient
+		/// balance.
+		InsufficientBalance,
+		/// The calculated collateral is zero.
+		ZeroCollateral,
 	}
 
 	#[pallet::composite_enum]
@@ -291,6 +296,7 @@ pub mod pallet {
 			currencies: BoundedVec<TokenMetaOf<T>, T::MaxCurrencies>,
 			denomination: u8,
 			transferable: bool,
+			min_operation_balance: u128,
 		) -> DispatchResult {
 			let who = T::PoolCreateOrigin::ensure_origin(origin)?;
 
@@ -352,6 +358,7 @@ pub mod pallet {
 					currency_ids,
 					transferable,
 					denomination,
+					min_operation_balance,
 				)),
 			);
 
@@ -477,6 +484,11 @@ pub mod pallet {
 			ensure!(pool_details.can_mint(&who), Error::<T>::NoPermission);
 
 			ensure!(
+				amount_to_mint >= pool_details.min_operation_balance.saturated_into(),
+				Error::<T>::InsufficientBalance
+			);
+
+			ensure!(
 				Self::get_currencies_number(&pool_details) <= currency_count,
 				Error::<T>::CurrencyCount
 			);
@@ -510,6 +522,7 @@ pub mod pallet {
 				pool_details.collateral_id.clone(),
 			)?;
 
+			ensure!(cost > Zero::zero(), Error::<T>::ZeroCollateral);
 			// fail if cost > max_cost
 			ensure!(cost <= max_cost, Error::<T>::Slippage);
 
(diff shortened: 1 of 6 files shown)


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -269,6 +269,11 @@ pub mod pallet {
 		InvalidInput,
 		Internal,
 		Slippage,
+		/// The requested Operation could not be performed due to insufficient
+		/// balance.
+		InsufficientBalance,
+		/// The calculated collateral is zero.
+		ZeroCollateral,
```
rflechtner, line 276: what happened to our ZeroAmount error? Is that gone?

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -269,6 +269,11 @@ pub mod pallet {
 		InvalidInput,
 		Internal,
 		Slippage,
+		/// The requested Operation could not be performed due to insufficient
+		/// balance.
+		InsufficientBalance,
```
rflechtner, line 274: I think that error is misleading, can we call this `BelowMinimum` and describe it as "The amount to mint or burn is below the minimum allowed?"

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -352,6 +358,7 @@ pub mod pallet {
 					currency_ids,
 					transferable,
 					denomination,
+					min_operation_balance,
```
rflechtner, line 361: should we force this to be at least 1 (e.g., simply replace if 0)? It makes no sense anyway to try to mint or burn 0 coins...

rflechtner (changes requested):

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -291,6 +296,7 @@ pub mod pallet {
 			currencies: BoundedVec<TokenMetaOf<T>, T::MaxCurrencies>,
 			denomination: u8,
 			transferable: bool,
+			min_operation_balance: u128,
```
ntn-x2, line 299: Why is this needed? Is it not enough to fail if the calculated collateral is `0`?

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Yes, I think I deleted that because the check was not really meaningful.

pallets/pallet-bonded-coins/src/lib.rs
Ad96el: Yeah, naming… Always difficult :grimacing: 

renamed in: [0e603c8](https://github.com/KILTprotocol/kilt-node/pull/803/commits/0e603c81763c2a2c67073a19887099e2be3c9167)

pallets/pallet-bonded-coins/src/lib.rs
ntn-x2: Does the pallet go in an inconsistent state if these checks were omitted? Or what's the worst that can happen if the pool owner misconfigures this value?

Ad96el: No, I don't think the pool goes into an inconsistent state. The relative error is just very high, but the minting and burning should still work. The user who wanted to mint the single unit can lose some money.

ntn-x2: Ok, then I don't have a strong opinion on this parameter.

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: Throwing an error because of this seems harsh, and 0 and 1 are identical in effect - minting / burning 0 will already always fail (due to ZeroCollateral), so silently raising a 0 to a 1 minimum just means throwing a different error instead (BelowMinimum). Do we need to bug people just trying to create a pool with default parameters with this detail?

Ad96el:
Superseded by: https://github.com/KILTprotocol/kilt-node/pull/808

PR #804, opened 2024-11-20
test: destroy

rflechtner:
## re/ https://github.com/KILTprotocol/ticket/issues/3671

Tests the three destruction extrinsics. All tests passing rn.

To be rebased after merging #791.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
+++ b/pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
@@ -204,7 +204,8 @@ fn fails_if_assets_cannot_be_destroyed() {
 		.build()
 		.execute_with(|| {
 			// Fails because asset is not in destroying state
-			assert!(BondingPallet::finish_destroy(origin.clone(), pool_id.clone(), 1).is_err());
+			BondingPallet::finish_destroy(origin.clone(), pool_id.clone(), 1)
+				.expect_err("Pool destruction should fail if any asset is not in destroying state.");
 
 			// Assets need to be in destroying state if pool is in destroying state
 			<Assets as Destroy<_>>::start_destroy(DEFAULT_BONDED_CURRENCY_ID, None)
--- a/pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
+++ b/pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
@@ -32,7 +32,7 @@ fn anyone_can_call_finish_destroy() {
 		.build()
 		.execute_with(|| {
 			// Assets need to be in destroying state if pool is in destroying state
-			<<Test as crate::Config>::Fungibles as Destroy<_>>::start_destroy(DEFAULT_BONDED_CURRENCY_ID, None)
+			<Assets as Destroy<_>>::start_destroy(DEFAULT_BONDED_CURRENCY_ID, None)
 				.expect("failed to set up test state: asset cannot be set to destroying");
 
 			let origin = RawOrigin::Signed(ACCOUNT_00).into();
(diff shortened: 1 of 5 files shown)


pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
Ad96el, line 1: I see an edge case regarding the deposits. If the deposit less the ED, it might be that the user got dusted, resulting into a failure to refund back his deposit. The pool life cycle would then be stuck. 

It can be avoided in the runtime, when the base deposit is equal or higher to the ED. Do we care about this test case?

pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
```diff
@@ -0,0 +1,285 @@
...
+
+	ExtBuilder::default()
+		.with_pools(vec![(pool_id.clone(), pool_details.clone())])
+		.with_native_balances(vec![(ACCOUNT_00, ONE_HUNDRED_KILT)])
+		.with_collaterals(vec![DEFAULT_COLLATERAL_CURRENCY_ID])
+		.with_bonded_balance(vec![(DEFAULT_BONDED_CURRENCY_ID, ACCOUNT_00, 100_000)])
+		.build()
+		.execute_with(|| {
+			// Fails because asset is not in destroying state
+			assert!(BondingPallet::finish_destroy(origin.clone(), pool_id.clone(), 1).is_err());
```
Ad96el, line 212: `assert_eq(BondingPallet::.finish_destroy(origin.clone(), pool_id.clone(), 1), Err(_));` is not possible?

pallets/pallet-bonded-coins/src/tests/transactions/start_destroy.rs
```diff
@@ -0,0 +1,412 @@
...
+		.with_collaterals(vec![DEFAULT_COLLATERAL_CURRENCY_ID])
+		.build()
+		.execute_with(|| {
+			let origin = RawOrigin::Signed(ACCOUNT_00).into();
+
+			assert_ok!(BondingPallet::start_destroy(origin, pool_id.clone(), 1));
+
+			// Verify that the pool state has been updated to destroying
+			let updated_pool = Pools::<Test>::get(&pool_id).unwrap();
+			assert!(matches!(updated_pool.state, PoolStatus::Destroying));
```
Ad96el, line 35: assert_eq! would look simpler. Same for the others further below

Ad96el (commented):
LGTM. One style suggestion.

pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
Ad96el: I think we have to use assert_eq. We introduced a neew lint, to avoid blank assert on results.

rflechtner: I think assert_eq only works for tests for equality, which don't allow matching patterns afaik. Same for assert_err!, which is why I'm not using that here. We can use a `matches!` expression, but the result is exactly the same as using the `is_err()`.

pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
rflechtner: The deposit is just reserved, never actually removed from the account, is it not?

pallets/pallet-bonded-coins/src/tests/transactions/finish_destroy.rs
rflechtner: I don't know, which one do you prefer?

Ad96el: Then leave it like that :+1:

pallets/pallet-bonded-coins/src/tests/transactions/start_destroy.rs
rflechtner: 1db6b8d7c70708533a100d0e28721449816fbe1a

PR #809, opened 2024-11-25
fix: runtime API implementations not exported

ntn-x2:
WIP. Fixes https://github.com/KILTprotocol/ticket/issues/3688.


Ad96el (approved):
LGTM. Can confirm that the API is present in the metadata after those changes.

PR #811, opened 2024-11-26
chore: exclude srtool and node_modules for taplo

Ad96el:


ntn-x2 (commented):
What's the issue here? I've never had any problems with this configuration in my PRs.

Ad96el:
It's not an issue in PRs. I use taplo in my terminal, and it can sometimes be annoying when I want to perform an operation, and it acts on all possible toml files. But we can also drop the changes.

ntn-x2:
Yes but if your CLI also formats those documents, shouldn't it show up as an error in the CI as well when we run the checks? Could it be your local CLI binary does not use the same config? I have nothing against this PR, but I'd just like to understand more, since I don't have the issue. But I use the `formatOnSave` setting of VSCode, so I never have to run the binary myself.

Ad96el:
The Taplo CLI uses the Taplo configuration from the repository. I am also using the CLI to check the formatting of the files. However, this causes Taplo to check TOML files in the node_modules, target, and target-analyze directories, which are not created in the CI. This is a bit annoying for local tests.

ntn-x2:
You can specify a path to the config file when running it from your terminal, and in the config file you can add more files to the ignore list, so they won't be checked.

Ad96el:
Excluding these files is what the PR currently does. The other TOML files are important and should be checked, but I could also specify the files manually in the CLI for local checks—it’s just annoying. I prefer running a single taplo check instead of multiple taplo check PATHS commands for the TOML files I might have touched during development.

Currently, TOML files from target and target-analyze are already excluded, and excluding TOML files from auto-generated code shouldn’t be an issue.

PR #812, opened 2024-11-26
refactor: decouple balance to fixed from config

rflechtner:
Unties the utility functions performing scaled conversion from a balance type to a fixed (and back) from the pallet config trait.
This also allows us to easily drop the requirement that the bonded coins and collateral need to use the same balance type.
Finally, I made sure that the symmetry of the two functions is reflected in their names and parameter order.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -137,7 +137,7 @@ pub mod pallet {
 			+ DestroyFungibles<Self::AccountId>
 			+ FungiblesMetadata<Self::AccountId>
 			+ FungiblesInspect<Self::AccountId>
-			+ MutateFungibles<Self::AccountId, Balance = CollateralCurrenciesBalanceOf<Self>>
+			+ MutateFungibles<Self::AccountId>
 			+ FreezeAccounts<Self::AccountId, Self::AssetId>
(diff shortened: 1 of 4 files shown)


Ad96el (dismissed):

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -300,7 +300,8 @@ pub mod pallet {
 	where
 		<CurveParameterTypeOf<T> as Fixed>::Bits:
 			Copy + ToFixed + AddAssign + BitOrAssign + ShlAssign + TryFrom<U256> + TryInto<U256>,
-		CollateralCurrenciesBalanceOf<T>: Into<U256> + TryFrom<U256> + TryInto<U256>,
+		CollateralCurrenciesBalanceOf<T>: Into<U256> + TryFrom<U256>,
+		FungiblesBalanceOf<T>: Into<U256> + TryFrom<U256>,
```
Ad96el, line 304: I think this has also to be specified in the benchmarking

rflechtner: 5b3a4ef3c3ee30c74d4e9f58521e3f29af0c0843

ntn-x2 (approved):

Ad96el (approved):

PR #813, opened 2024-11-27
feat: support gatekeeping web3names and account linkings

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3691.

This PR allows to dynamically (i.e., via the sudo origin) elect an account as the only authorised to submit `claim` extrinsic for any web3name pallet deployment, and `associate_account` and `associate_sender` extrinsics for any did-lookup deployment. 

I also updated our dotnames and unique linking deployments for both our runtimes to use this new feature, althought I had to rely on a trick with `storage_alias`es to avoid introducing a new pallet just for this. Unfortunately, the [pallet-parameters](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/parameters) was only introduced in 1.8.0, so we might want to migrate to using that once we update our codebase to 1.8.0. For now, I could not think of any other way to implement this feature without touching the pallets storage entries, which I did not want to.

For both our runtimes, these are the two new storage keys introduced:

* `0x8ea135058ec16554c8e3d230d658fbffd30ff375811804de60521a1654f58ebb` for the dotnames deployment authorization
* `0x41a63f711fa40ef5e1dc8f0ac115a906d4378bcb7f1d95ba1124c2140bfccdba` for the unique linking deployment authorization

These values can be updated with a `system.setStorage(key, value)` call, which must specify an account ID as the sole submitter of the extrinsics specified above. A `system.killStorage(keys)` call will set the relative entry to `None`, which means no gating is enforced and anyone can create. This is the default for our web3name and did linking deployments.

**Important: when deploying the new runtime, we will also need to set the storage value for these entries, and we don't need to wait for the new runtime to be live as writing it earlier has no effect on the rest of the runtime**.

## How to test

1. Spin up a chopsticks Peregrine setup
2. Set a desired account as the sole allowed submitter for dotnames
    * E.g., This call sets it to the sudo `0x000404808ea135058ec16554c8e3d230d658fbffd30ff375811804de60521a1654f58ebb80921cbc0ffe09a865dbf4ae1d0410aa17c656881fe86666da0f97939e3701b674` 
4. Try to claim a dotname with any DID while submitting the tx with an account different than the sudo account: this will not work
5. Try again with the sudo account and it will work.
6. Remove the authorised account with `killStorage`
    * This call removes it: `0x000504808ea135058ec16554c8e3d230d658fbffd30ff375811804de60521a1654f58ebb`
8. Try again with the first account: it will now work.

### DIFF
--- a/pallets/did/src/origin.rs
+++ b/pallets/did/src/origin.rs
@@ -88,8 +88,8 @@ where
 	}
 }
 
-impl<OuterOrigin, DidIdentifier, AccountId> EnsureOriginWithArg<OuterOrigin, DidIdentifier>
-	for EnsureDidOrigin<DidIdentifier, AccountId>
+impl<OuterOrigin, DidIdentifier, AccountId, ExpectedSubmitter> EnsureOriginWithArg<OuterOrigin, DidIdentifier>
+	for EnsureDidOrigin<DidIdentifier, AccountId, ExpectedSubmitter>
 where
 	OuterOrigin: Into<Result<DidRawOrigin<DidIdentifier, AccountId>, OuterOrigin>>
 		+ From<DidRawOrigin<DidIdentifier, AccountId>>
@@ -121,9 +121,9 @@ where
 }
 
 #[cfg(feature = "runtime-benchmarks")]
-impl<OuterOrigin, AccountId, DidIdentifier>
+impl<OuterOrigin, AccountId, DidIdentifier, ExpectedSubmitter>
 	kilt_support::traits::GenerateBenchmarkOrigin<OuterOrigin, AccountId, DidIdentifier>
-	for EnsureDidOrigin<DidIdentifier, AccountId>
+	for EnsureDidOrigin<DidIdentifier, AccountId, ExpectedSubmitter>
 where
 	OuterOrigin: Into<Result<DidRawOrigin<DidIdentifier, AccountId>, OuterOrigin>>
 		+ From<DidRawOrigin<DidIdentifier, AccountId>>,
--- a/pallets/did/src/origin.rs
+++ b/pallets/did/src/origin.rs
@@ -20,6 +20,7 @@ use frame_support::traits::{EnsureOrigin, EnsureOriginWithArg};
 use kilt_support::traits::CallSources;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
+use sp_core::Get;
 use sp_runtime::RuntimeDebug;
 use sp_std::marker::PhantomData;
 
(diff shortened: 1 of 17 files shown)


Ad96el:
> Important: when deploying the new runtime, we will also need to set the storage value for these entries, and we don't need to wait for the new runtime to be live as writing it earlier has no effect on the rest of the runtime.

Can we create a ticket for this? We can probably already tackle that with a placeholder for the needed account.

runtimes/common/src/dot_names/origin.rs
```diff
@@ -0,0 +1,83 @@
...
+pub(super) mod dot_names {
+	use frame_support::{pallet_prelude::OptionQuery, storage_alias, traits::PalletInfoAccess};
+	use sp_core::Get;
+	use sp_std::marker::PhantomData;
+
+	use crate::AccountId;
+
+	const LOG_TARGET: &str = "runtime::DotNames::AllowedDotNameClaimer";
+
+	// TODO: When upgrading to 1.8.0, which introduces a new `pallet-parameteres`,
```
Ad96el, line 28: I see a big warning, that the pallet is not audited, and should not be used in production. 

Could you also maybe elaborate, how the pallet would work to fix this problem? As far as I can see, the pallet is used to update a parameter of another pallet. Why do we need that?

pallets/did/src/origin.rs
```diff
@@ -36,19 +37,43 @@ impl<DidIdentifier, AccountId> DidRawOrigin<DidIdentifier, AccountId> {
...
+		self.submitter.clone()
+	}
 
-impl<OuterOrigin, DidIdentifier, AccountId> EnsureOrigin<OuterOrigin> for EnsureDidOrigin<DidIdentifier, AccountId>
+	fn subject(&self) -> DidIdentifier {
+		self.id.clone()
+	}
+}
+
+pub struct EnsureDidOrigin<DidIdentifier, AccountId, ExpectedSubmitter = ()>(
```
Ad96el, line 52: Its more `EnsureDidAuthorizedOrigin`. Feel free to resolve this.

Ad96el (commented):
Some questions.

runtimes/common/src/dot_names/origin.rs
ntn-x2: The pallet is basically a generic key-value store where each key has an authorised origin that can change it. So in our case the value we need could be stored in there and updated by governance. I did not read about production readiness tho.

pallets/did/src/origin.rs
ntn-x2: Yeah in the beginning I went for a new struct altogether, but then realised it's actually easier to change the existing one, so `EnsureDidOrigin` is the origin check we already have in place, and I don't think renaming it right now is a good idea.

runtimes/common/Cargo.toml
```diff
@@ -91,6 +91,7 @@ runtime-benchmarks = [
   "pallet-multisig/runtime-benchmarks",
   "pallet-tips/runtime-benchmarks",
   "pallet-treasury/runtime-benchmarks",
+  "pallet-web3-names/runtime-benchmarks",
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/813/commits/bd2c68b7b52ff9c6ac0e62e191c0920551a2200c

runtimes/common/src/dot_names/origin.rs
Ad96el: I am struggling to get why we need the new pallet. You have already described the process in the description of the PR

ntn-x2: The main limitation of this solution is that we don't have any try-runtime support, and it's effectively a storage element the runtime has no knowledge about, meaning it won't be checked in any sorts of tests. Using a pallet would give us a "standard" place to store all our stuff, and to migrate and check accordingly.

Ad96el: I see

Ad96el (approved):

PR #814, opened 2024-11-27
fix: benchmarking

Ad96el:
The user was missing some collateral, to pay the mint operation. Previously, the calculated collateral was rounded down to zero, allowing the user to mint coins without paying. 

The `force_start_refund` function was messy.


rflechtner (approved):
If you get the lints and tests to pass, this LGTM. It complains about redundant clones rn though.

ntn-x2 (approved):
Good for me. I guess once the big PR is merged in `develop`, we will test the benchmarks and catch any other issues.

PR #818, opened 2024-11-29
feat: bonded coins peregrine

Ad96el:
# Fixes: [3692](https://github.com/KILTprotocol/ticket/issues/3692) 
Includes the bonded coins to the peregrine runtime.


ntn-x2:
Wouldn't this depend on the resolution of https://github.com/KILTprotocol/kilt-node/pull/764#pullrequestreview-2468215124?

Ad96el:
Yes sorry. Forgot to put the on hold label.

runtimes/common/src/bonded_currencies/hooks.rs
```diff
@@ -0,0 +1,58 @@
...
+use sp_std::marker::PhantomData;
+
+use crate::bonded_currencies::AssetId;
+
+#[storage_alias]
+pub type NextAssetId<BondingPallet: PalletInfoAccess, T> =
+	StorageValue<BondingPallet, FungiblesAssetIdOf<T>, OptionQuery>;
+
+/// Struct to implement desired traits for [NextAssetId].
+pub struct NextAssetIdGenerator<T>(PhantomData<T>);
```
Ad96el, line 34: Was this envisioned?

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+		Left: fungible::Inspect<AccountId>,
+		Right: Inspect<AccountId, Balance = <Left as fungible::Inspect<AccountId>>::Balance>,
+		Criterion: Convert<AssetKind, Either<(), Right::AssetId>>,
+		AssetKind: AssetIdTraits,
+		AccountId,
+	> Inspect<AccountId> for WrapperNativeAndForeignAssets<Left, Right, Criterion, AssetKind, AccountId>
+{
+	fn decimals(asset: Self::AssetId) -> u8 {
+		match Criterion::convert(asset) {
+			Left(()) => 15u8,
```
Ad96el, line 268: Should I move this into const?

Ad96el: It is not used multiple time

Ad96el:
@ntn-x2 The PR is ready for review!

scripts/run_benches_for_runtime.sh
```diff
@@ -40,8 +40,8 @@ pallets=(
 	pallet-message-queue
 	cumulus-pallet-parachain-system
 	pallet_multisig
-	pallet-assets
 	pallet-asset-switch
+	pallet-assets
```
ntn-x2, line 44: Just for curiosity, was this the result of running some fmt on this file?

runtimes/common/src/dot_names/mod.rs
```diff
@@ -58,7 +58,10 @@ impl<const MIN_LENGTH: u32, const MAX_LENGTH: u32> TryFrom<Vec<u8>> for DotName<
 	type Error = DotNameValidationError;
 
 	fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
-		ensure!(value.len() >= MIN_LENGTH.saturated_into(), Self::Error::TooShort);
+		ensure!(
```
ntn-x2, line 61: Why are these changes here? Do you use a different fmt setup? Are you using the right nightly to format these?

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+// The KILT Blockchain is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+pub mod hooks;
```
ntn-x2, line 18: Mods come usually after the uses.

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+/// The AssetId for bonded assets.
+pub type AssetId = u32;
+
+/// Fixed point number used for creating bonding curves.
+pub type FloatInput = U75F53;
+
+/// Fixed point number used for doing calculation steps in the bonding curves.
+pub type Float = I75F53;
+
+pub struct TargetFromLeft<Target, L = Location>(PhantomData<(Target, L)>);
```
ntn-x2, line 50: A comment explaining this struct?

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+impl<Target: Get<L>, L: PartialEq + Eq> Convert<L, Either<(), L>> for TargetFromLeft<Target, L> {
```
ntn-x2, line 51: If `L` is already part of the trait, you can avoid adding it to the `TargetFromLeft` struct as well.

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+	fn convert(l: L) -> Either<(), L> {
+		Target::get().eq(&l).then(|| Left(())).map_or(Right(l), |n| n)
```
ntn-x2, line 53: This definitely could use some comments!

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+pub type NativeAndForeignAssets<Balances, Fungibles, Criterion, AssetKind, AccountId> =
+	UnionOf<Balances, Fungibles, Criterion, AssetKind, AccountId>;
+
+pub struct WrapperNativeAndForeignAssets<Left, Right, Criterion, AssetKind, AccountId>(
```
ntn-x2, line 60: Please explain why this is needed.

runtimes/common/src/bonded_currencies/mod.rs
ntn-x2: Also, shouldn't this use the decimal amount specified for the balances pallet? Or why are we returning `15` by default?

runtimes/common/src/bonded_currencies/hooks.rs
```diff
@@ -0,0 +1,56 @@
...
+	storage_alias,
+};
+use pallet_bonded_coins::{traits::NextAssetIds, FungiblesAssetIdOf};
+use sp_runtime::{ArithmeticError, BoundedVec, DispatchError};
+use sp_std::{marker::PhantomData, vec::Vec};
+
+use crate::bonded_currencies::AssetId;
+
+#[storage_alias]
+pub type NextAssetId<BondingPallet: PalletInfoAccess, T> =
```
ntn-x2, line 30: Is the next asset ID stored in the assets instance, or in the bonding curve instance? In the former case, the generic name should change. In the second, this is ok. Just double checking.

runtimes/common/src/bonded_currencies/hooks.rs
```diff
@@ -0,0 +1,56 @@
...
+	StorageValue<BondingPallet, FungiblesAssetIdOf<T>, OptionQuery>;
```
ntn-x2, line 31: I guess using `ValueQuery` and hence returning `0` would probably be a better idea.

runtimes/common/src/bonded_currencies/hooks.rs
```diff
@@ -0,0 +1,56 @@
...
+
+		let new_next_asset_id = next_asset_id.checked_add(n).ok_or(ArithmeticError::Overflow)?;
+
+		let asset_ids = (next_asset_id..new_next_asset_id)
+			.map(FungiblesAssetIdOf::<T>::from)
+			.collect::<Vec<FungiblesAssetIdOf<T>>>();
+
+		NextAssetId::<BondingPallet, T>::set(Some(new_next_asset_id.into()));
+		BoundedVec::try_from(asset_ids)
+			.map_err(|_| DispatchError::Other("Could not create Bounded vec in NextAssetIdGenerator"))
```
ntn-x2, line 54: I would rather return the bonding curve pallet's `Internal` error + logs here.

runtimes/peregrine/src/kilt/mod.rs
```diff
@@ -100,3 +106,32 @@ impl pallet_asset_switch::Config<KiltToEKiltSwitchPallet> for Runtime {
 	#[cfg(feature = "runtime-benchmarks")]
 	type BenchmarkHelper = crate::benchmarks::asset_switch::CreateFungibleForAssetSwitchPool1;
 }
+
+parameter_types! {
+	pub const NativeAsset: Location = Junctions::Here.into_location();
+}
+
+impl pallet_bonded_coins::Config for Runtime {
+	type BaseDeposit = runtime_common::constants::bonded_coins::BaseDeposit;
```
ntn-x2, line 115: I think the assets for the bonding curve should go in the same module. So if the bonded coins config is implemented here, I would move over the bonded assets config here as well.

runtimes/peregrine/src/benchmarks/mod.rs
```diff
@@ -59,6 +60,9 @@ use pallet_membership as pallet_technical_membership;
 #[allow(unused_imports)]
 use pallet_web3_names as pallet_dot_names;
```
ntn-x2, line 62: Why the whitespace?

ntn-x2 (commented):
What about the proxy configs?

runtimes/peregrine/src/benchmarks/mod.rs
Ad96el: I kinda overcomplicated it: [aaeeeac](https://github.com/KILTprotocol/kilt-node/pull/818/commits/aaeeeacfbf989c2201fde73a478cc47f270fdd48)

runtimes/peregrine/src/lib.rs
```diff
@@ -146,6 +146,7 @@ construct_runtime! {
 
 		AssetSwitchPool1: pallet_asset_switch::<Instance1> = 48,
 		Fungibles: pallet_assets = 49,
+		BondedFungibles: pallet_assets::<Instance1> = 50,
```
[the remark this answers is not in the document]
Ad96el: Updated in: [b827db6](https://github.com/KILTprotocol/kilt-node/pull/818/commits/b827db6507e6d842620a9f4e61999a34dd797fb1)

runtimes/common/src/bonded_currencies/hooks.rs
Ad96el: I can remember, I had a strange error when I was using `ValueQuery`. But It seems to compile today: [d95534e](https://github.com/KILTprotocol/kilt-node/pull/818/commits/d95534ea19ddd0031a10a8c5faef9df932954c50)

Not sure what was wrong.

runtimes/common/src/bonded_currencies/hooks.rs
Ad96el: I renamed the generic in: [10b4b2e](https://github.com/KILTprotocol/kilt-node/pull/818/commits/10b4b2ef17b4570d41e9e0119832e9f7642f0bf3)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: I added documentation in: [57155ad](https://github.com/KILTprotocol/kilt-node/pull/818/commits/57155adffd2905b964e38a3d2e5b9d44e48c5dd5) 

I will also rename the struct in another commit.

Ad96el: renaming in: [178e292](https://github.com/KILTprotocol/kilt-node/pull/818/commits/178e292fc380abdcebd14809014d1bd39f481e83)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Added some explanation in: [57155ad](https://github.com/KILTprotocol/kilt-node/pull/818/commits/57155adffd2905b964e38a3d2e5b9d44e48c5dd5)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Yeah true: [57155ad](https://github.com/KILTprotocol/kilt-node/pull/818/commits/57155adffd2905b964e38a3d2e5b9d44e48c5dd5)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Comment in: [57155ad](https://github.com/KILTprotocol/kilt-node/pull/818/commits/57155adffd2905b964e38a3d2e5b9d44e48c5dd5)

runtimes/common/src/dot_names/mod.rs
Ad96el: I think I do. 

These are vscode settings. The tool chain is installed.
```
  "rust-analyzer.rustfmt.extraArgs": [
        "+nightly-2023-10-02"
  ],
```
Not sure why this is happening

Ad96el: saved without fmt in: [20432ec](https://github.com/KILTprotocol/kilt-node/pull/818/commits/20432ec66223bcc1977eccee118695f2ce62f12c)

scripts/run_benches_for_runtime.sh
Ad96el: No changes :thinking:

Ad96el: Could be. I am using shfmt.

Ad96el: I don’t think so. If I remember correctly, I initially added only the Peregrine-specific pallets to the pallet list but later reverted that change.

runtimes/common/src/constants.rs
```diff
@@ -158,6 +158,31 @@ pub mod attestation {
...
+	/// 1 Storage entry in the assets pallet.
+	/// AssetId has 8 bytes. Two times stored in the storage. -> 16 bytes
+	/// struct AssetMetadata has 18 bytes + name + symbol -> 18 + 8 = 26 bytes
+	/// https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/assets/src/types.rs#L188
+	pub const DEPOSIT_PER_CURRENCY: Balance = deposit(1, 42);
+	pub const MAX_CURRENCIES: u32 = 50;
+	pub const MAX_DENOMINATION: u8 = 16;
+	pub const MAX_STRING_LENGTH: u32 = crate::constants::assets::STRING_LIMIT;
+
+	parameter_types! {
```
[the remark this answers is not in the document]
Ad96el: Updated in: [e6800a2](https://github.com/KILTprotocol/kilt-node/pull/818/commits/e6800a25e99bf3a3008323265692814805fe0013)

Ad96el:
Proxy settings are added in: [d9c47cf](https://github.com/KILTprotocol/kilt-node/pull/818/commits/d9c47cfb87030abd54bf55bf85b7faf4ce8813ef)

runtimes/common/src/bonded_currencies/hooks.rs
```diff
@@ -0,0 +1,60 @@
...
+
+		let asset_ids = (next_asset_id..new_next_asset_id)
+			.map(FungiblesAssetIdOf::<T>::from)
+			.collect::<Vec<FungiblesAssetIdOf<T>>>();
+
+		NextAssetId::<BondedFungibles, T>::set(new_next_asset_id.into());
+		BoundedVec::try_from(asset_ids.clone()).map_err(|_| {
+			log::error!(target: LOG_TARGET, "Failed to convert asset ids to bounded vec {:?}", asset_ids);
+			Error::<T>::Internal.into()
+		})
```
rflechtner, line 58: Antonio mentioned that the trait could also use a simple Vec - guess we could still change this to avoid this pointless size limit. What we would rather want to check (but that should be part of the pallet logic) is that the number of ids returned is equal to the number of ids requested.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -116,7 +119,7 @@ pub mod pallet {
 
 	pub(crate) type CurveParameterInputOf<T> = <T as Config>::CurveParameterInput;
 
-	pub(crate) type PoolDetailsOf<T> = PoolDetails<
+	pub type PoolDetailsOf<T> = PoolDetails<
```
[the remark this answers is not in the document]
Ad96el: Yes that's intended. I wanted to calculate the max size at the runtime to determine the accurate deposit.

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -868,7 +868,6 @@ mod benchmarks {
 		make_free_for_deposit::<T>(&pool_account);
 		let collateral_id = create_default_collateral_asset::<T>();
 		T::Collaterals::touch(collateral_id.clone(), &pool_account, &pool_account).expect("Touching should work");
-
 		set_collateral_balance::<T>(collateral_id, &pool_account, 10000u128);
```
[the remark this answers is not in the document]
Ad96el: Yes I think so sorry

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+pub mod hooks;
+
+/// The AssetId for bonded assets.
+pub type AssetId = u32;
+
+/// Fixed point number used for creating bonding curves.
+pub type FloatInput = U75F53;
+
+/// Fixed point number used for doing calculation steps in the bonding curves.
+pub type Float = I75F53;
```
rflechtner, line 48: why do we call these floats even though they aren't?

runtimes/common/src/bonded_currencies/mod.rs
rflechtner: if the balances pallet allows us to read the native currency's denomination, it would indeed make sense to do so

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+	fn decimals(asset: Self::AssetId) -> u8 {
+		match Criterion::convert(asset) {
+			Left(()) => 15u8,
+			Right(a) => Right::decimals(a),
+		}
+	}
+
+	fn name(asset: Self::AssetId) -> Vec<u8> {
+		match Criterion::convert(asset) {
+			Left(()) => Vec::new(),
```
rflechtner, line 266: same for name and symbol; if we were to expose pool info via a runtime call this would be really helpful for example.

runtimes/common/src/dot_names/mod.rs
rflechtner: you added the `<usize>` right? how come this was suddenly necessary?

runtimes/common/src/bonded_currencies/hooks.rs
Ad96el: Yeah, good point. I will open up another PR.

runtimes/peregrine/src/system/proxy.rs
```diff
@@ -205,7 +217,20 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 							| pallet_web3_names::Call::unban { .. }
 							| pallet_web3_names::Call::update_deposit { .. }
 							| pallet_web3_names::Call::change_deposit_owner { .. }
-					),
+					)
+					| RuntimeCall::BondedCurrencies(
+						// Excludes `finish_destroy`
```
rflechtner, line 222: that's pretty pointless, the call is permissionless

runtimes/common/src/bonded_currencies/hooks.rs
Ad96el: https://github.com/KILTprotocol/kilt-node/pull/826

rflechtner (commented):
Looks pretty good, I have one request though based on the demo this week. I was horrified to see that apparently the SCALE type info for the Fixed types marks them as the type of their underlying integer type, in this case a `u128`/`i128`. Threw up in my mouth a little when seeing that you have to encode the coeficients as a Fixed, then take the encoded bytes and re-code them as an integer, which you then paste into the pjs apps. If this type info is in the metadata, this will unfortunately affect all tooling and we need to do something about it. I see two routes we can explore:

- Choose a different type for the `CurveParameterInput` type - we could even check if using floats here would pose a problem?
- See if we can override the `Encode`/`Decode` or `scale_info::TypeInfo` traits of the type we use for `CurveParameterInput`. I'd rather have it typed as a `[u8; 8]` then misrepresented as an integer.

runtimes/peregrine/src/system/proxy.rs
Ad96el: -.- true.

runtimes/common/src/bonded_currencies/hooks.rs
rflechtner: let's just remove this purely theoretical error case by switching to a Vec - the length bound is meaningless anyway, given that there is another length restriction that must actually hold: the Vec length must be equal to `n`

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: The balances pallet is not exposing such information. The denomination are metadata, set in the chain spec. At runtime level there are unknown. 
Same for symbol and name. 

I have moved the denomination into a constant in: [bc4dbae](https://github.com/KILTprotocol/kilt-node/pull/818/commits/bc4dbae37ee31eca96bb419b7174c20e89b00ba1). 

I think symbol and name are not necessary. They are not used in the pallet, and we will not need them in the runtime api.

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Do you have a better name in mind? `FixedPoint` maybe?

runtimes/peregrine/src/system/proxy.rs
Ad96el: I allow all tx: [f619872](https://github.com/KILTprotocol/kilt-node/pull/818/commits/f6198720463b18e6a82c596d035323c470432099)

runtimes/common/src/bonded_currencies/mod.rs
rflechtner: Yeah, that works

Ad96el: Updated in:[0ff6f03](https://github.com/KILTprotocol/kilt-node/pull/818/commits/0ff6f03673c168f11a300f0f141b51824213083f)

scripts/run_benches_for_runtime.sh
ntn-x2: Then this change could probably be reverted as well since it's not really useful.

runtimes/common/src/constants.rs
```diff
@@ -164,11 +164,10 @@ pub mod bonded_coins {
 	/// The size is checked in the runtime by a test.
 	pub const MAX_POOL_BYTE_LENGTH: u32 = 970;
 	pub const BASE_DEPOSIT: Balance = deposit(1, MAX_POOL_BYTE_LENGTH);
-	/// 1 Storage entry in the assets pallet.
-	/// AssetId has 8 bytes. Two times stored in the storage. -> 16 bytes
-	/// struct AssetMetadata has 18 bytes + name + symbol -> 18 + 8 = 26 bytes
+	pub const ASSET_ID_BYTE_LENGTH: u32 = 16;
```
ntn-x2: They do not need to be `pub`

runtimes/peregrine/src/kilt/mod.rs
```diff
@@ -112,19 +114,19 @@ parameter_types! {
 }
 
 impl pallet_bonded_coins::Config for Runtime {
-	type BaseDeposit = runtime_common::constants::bonded_coins::BaseDeposit;
+	type BaseDeposit = ConstU128<{ BASE_DEPOSIT }>;
```
ntn-x2, line 117: I think it's good to keep namespacing constants with the module they belong to.

runtimes/common/src/bonded_currencies/mod.rs
```diff
@@ -65,12 +65,12 @@ impl<Target: Get<L>, L: PartialEq + Eq> Convert<L, Either<(), L>> for TargetFrom
 
 /// Wrapper struct for [UnionOf] to implement the [metadata::Inspect] trait,
 /// needed for the pallet_bonded_coins module.
-pub struct WrapperNativeAndForeignAssets<Left, Right, Criterion, AssetKind, AccountId>(
+pub struct NativeAndForeignAssets<Left, Right, Criterion, AssetKind, AccountId>(
```
ntn-x2, line 70: Generics `Left` and `Right` should rather be `Native` and `ForeignAssets`, I would suggest.

runtimes/peregrine/src/kilt/mod.rs
ntn-x2: The new assets deployment solely exists for bonding curve functionality, so to me it feels fair that they go together, in whatever module.

ntn-x2 (commented):
Few more nitpicks.

runtimes/common/src/constants.rs
Ad96el: removed all unnecessary pubs in: [5f97947](https://github.com/KILTprotocol/kilt-node/pull/818/commits/5f979479d8831bbf3992dfaa7476a7e96032e38a)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Renamed in: [37299a7](https://github.com/KILTprotocol/kilt-node/pull/818/commits/37299a78ddb518cae0ae5a5450bd6a1145ec444c)

runtimes/peregrine/src/kilt/mod.rs
Ad96el: made them explicit in: [5578ec5](https://github.com/KILTprotocol/kilt-node/pull/818/commits/5578ec5bc61d7e7b841915b0f37cd2f72dfc7231)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: I hard coded KILT in the runtime api pr: https://github.com/KILTprotocol/kilt-node/pull/828. 

I hope it's fine to hard code those, otherwise I have to introduce another generic, to fetch the right symbol and  name, as they are different in the different runtimes.

Ad96el:
> Choose a different type for the CurveParameterInput type - we could even check if using floats here would pose a problem?

Floats are not implementing the `TypeInfo` trait and can not be used. We can think of using `BoundedVec<u8>` which has the drawback, that the user ends up with a coefficient, he didn't want.  

> See if we can override the Encode/Decode or scale_info::TypeInfo traits of the type we use for CurveParameterInput. I'd rather have it typed as a [u8; 8] then misrepresented as an integer.

Overwriting traits is not possible in rust. We either have to introduce a wrapper struct and implement those traits, which feels like not worth it. I would suggest we keep that in a backlog, and we can try to tackle that once the audit is there. I created a ticket for that: [#3729](https://github.com/KILTprotocol/ticket/issues/3729)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: reverted in: [afffda3](https://github.com/KILTprotocol/kilt-node/pull/818/commits/afffda333d5e3253f9a6867124444d7325f922fa)

scripts/run_benches_for_runtime.sh
Ad96el: reverted in: [afffda3](https://github.com/KILTprotocol/kilt-node/pull/818/commits/afffda333d5e3253f9a6867124444d7325f922fa)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: Oh, I did that my whole live wrong. Existential crises. Ops… 

I am not sure where I changed that, but I moved the mods after the uses.

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: NVM we need them in the runtime api: [8330fa8](https://github.com/KILTprotocol/kilt-node/pull/818/commits/8330fa81335a63e6cf78d77ccae89ab28049ade1)

runtimes/common/src/dot_names/mod.rs
Ad96el: Oh, yes. I think I executed `cargo clippy --fix`. I try to investigate why some unrelated crates requires some changes

runtimes/common/src/bonded_currencies/mod.rs
ntn-x2: I would suggest you move the value of decimals for peregrine into the peregrine runtime crate, and use that also when generating the chainspec in the binary. Then, you can use the same const here.

ntn-x2: Sorry but your commit history seems to be a mess and all the commits you link to give a 404 error.

runtimes/peregrine/src/kilt/mod.rs
ntn-x2: Sorry to be picky, but we do import the `constants` module, so you can get rid of the `runtime_common` namespace.

ntn-x2:
@Ad96el I resolved most comments and still waiting for few to be addressed by you. The commit history seems to be messed up and I could not successfully click on any of the commits you link in your answers, so I hope I did not miss anything important.

runtimes/peregrine/src/kilt/mod.rs
Ad96el: Changed it back in: [4296c3b](https://github.com/KILTprotocol/kilt-node/pull/818/commits/4296c3bf94f6a59af86e247192cafa43688f266f)

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: I changed a bit the implementation. i introduced another generic in the type to fetch the metadata: [0dcf5b3](https://github.com/KILTprotocol/kilt-node/pull/818/commits/0dcf5b38aaf642cb694d3e8cc7cb222218a091ea) 

It's adding a bit more complexity. Let me know what you think

Ad96el: yeah sorry I was thinking rebasing would be a good idea...

runtimes/common/src/bonded_currencies/mod.rs
Ad96el: I changed my mind. I have already introduced the metadata in this pr. Please check out this commit: [0dcf5b3](https://github.com/KILTprotocol/kilt-node/pull/818/commits/0dcf5b38aaf642cb694d3e8cc7cb222218a091ea)

runtimes/peregrine/src/kilt/mod.rs
```diff
@@ -112,9 +114,21 @@ parameter_types! {
 	pub const NativeAsset: Location = Junctions::Here.into_location();
 }
 
+pub struct MetadataProvider;
```
ntn-x2, line 117: This looks to me an overkill. I would also not put this stuff in `constants`, but would declare them in each runtime. So basically I would introduce two more consts next to where you define `CURRENCY_SYMBOL`. As for solving the problem with `NativeAndForeignAssets`, I would simply move that inside Peregrine runtime, and define a different one for Spiritnet later on.

ntn-x2: I thought about this, and actually I think it's ok to leave it in `runtime_common`. Please use types instead of tuples for the return type, and perhaps quickly check if there's already a trait that could be used, before declaring our own.

Ad96el: I couldn't find a trait for the native assets. There is one for `fungibles` which is not perfectly suited. 

Let me know what you think: [77ce5ce](https://github.com/KILTprotocol/kilt-node/pull/818/commits/77ce5ce2fa74fe1c89ceb62f31fd42c4b9fe1895)

rflechtner (approved):
All good from my side

runtimes/peregrine/src/system/mod.rs
```diff
@@ -53,6 +53,12 @@ pub const SS_58_PREFIX: u16 = 38;
 /// Currency Symbol
 pub const CURRENCY_SYMBOL: &[u8] = b"PILT";
 
+/// The denomination of native currency.
+pub const DENOMINATION: u8 = 15;
```
ntn-x2, line 57: Why do we still need these values here if we now have the `InspectMetadata` trait?

Ad96el: I am using these values for the implementation of the `InspectMetadata` trait. I could define them directly within the implementation, but I kinda like the const definition with a declarative name and documentation. Additionally, I am using these values in the parachain node implementation for generating a new chain spec.

ntn-x2: I think it would be best to have a single source of truth. You could require the trait in the node implementation directly, or then get rid of the trait and use constants everywhere, but not both.

Ad96el: [19762e4](https://github.com/KILTprotocol/kilt-node/pull/818/commits/19762e4bb89b367442a29cb4c809c5c394764468)

runtimes/common/src/bonded_coins/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+pub type FixedPointUnderlyingType = <FixedPoint as Fixed>::Bits;
+
+/// Struct to implement the desired [Convert] trait needed for the
+/// [NativeAndForeignAssets] type.
+/// The generic type Target is used to determine the type of the asset id
+/// for the Either::Left variant.
+pub struct TargetFromLeft<Target>(PhantomData<Target>);
+
+/// Metadata trait for native asset.
+pub trait InspectMetadata<AssetId> {
```
ntn-x2, line 67: This trait actually belongs to `kilt_support`, and not `runtime_common`.

ntn-x2: Also, what's the need for `AssetId`? It's not used anywhere in the trait.

ntn-x2 (commented):
Two more points, then we're good!

runtimes/common/src/bonded_coins/mod.rs
Ad96el: Ah yeah. Good catch: [7af4e61](https://github.com/KILTprotocol/kilt-node/pull/818/commits/7af4e61675178db659f264da90c7a70061475c6a)

ntn-x2 (approved):

PR #819, opened 2024-11-29
feat: better handling of DID deletions

ntn-x2:
WIP. Fixes https://github.com/KILTprotocol/ticket/issues/3673.

### DIFF
--- a/pallets/did/src/lib.rs
+++ b/pallets/did/src/lib.rs
@@ -88,6 +88,7 @@ pub mod errors;
 pub mod migrations;
 pub mod origin;
 pub mod service_endpoints;
+pub mod traits;
 
 #[cfg(test)]
 mod mock;
@@ -171,6 +172,7 @@ pub mod pallet {
 			DidEncryptionKey, DidSignature, DidVerifiableIdentifier, DidVerificationKey, RelationshipDeriveError,
 		},
 		service_endpoints::{utils as service_endpoints_utils, ServiceEndpointId},
+		traits::DeletionHelper,
 	};
 
 	/// The current storage version.
@@ -328,6 +330,10 @@ pub mod pallet {
 
 		/// Migration manager to handle new created entries
 		type BalanceMigrationManager: BalanceMigrationManager<AccountIdOf<Self>, BalanceOf<Self>>;
+
+		/// Helper trait to aid in cleaning up DID-related resources across all
+		/// runtime pallets.
+		type DeletionHelper: DeletionHelper<Self>;
 	}
 
 	#[pallet::pallet]
@@ -455,6 +461,9 @@ pub mod pallet {
 		/// The number of service endpoints stored under the DID is larger than
 		/// the number of endpoints to delete.
 		MaxStoredEndpointsCountExceeded,
+		/// The DID is linked to other runtime elements that would be left
+		/// dangling if the DID were to be deleted.
+		NonZeroReferences,
 		/// An error that is not supposed to take place, yet it happened.
 		Internal,
 	}
@@ -1495,6 +1504,10 @@ pub mod pallet {
 			ensure!(
 				current_endpoints_count <= endpoints_to_remove,
 				Error::<T>::MaxStoredEndpointsCountExceeded
+			);
+			ensure!(
+				T::DeletionHelper::linked_resources_count(&did_subject).is_zero(),
+				Error::<T>::NonZeroReferences
 			);
 
 			// This one can fail, albeit this should **never** be the case as we check for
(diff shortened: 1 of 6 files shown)


ntn-x2:
@rflechtner @Ad96el just a ping to keep this in your inbox, since we need to discuss about the approach proposed here.

pallets/did/src/lib.rs
```diff
@@ -1254,11 +1268,62 @@ pub mod pallet {
...
+			// matches the origin.
+			} else if let Ok(did_origin) = T::EnsureOrigin::ensure_origin(origin) {
+				ensure!(did_origin.subject() == did, Error::<T>::BadDidOrigin);
+			// We don't accept any other origin -> fail.
+			} else {
+				return Err(DispatchError::BadOrigin);
+			};
+
+			// The cleanup logic does not need to get benchmarked as it consumes the
+			// provided weight.
```
Ad96el, line 1294: I think this is not necessary true. `remaining_weight` is only getting updated by the performed operation or not? 

So the processing in the loop and the event creation has to be benchmarked or not?

Ad96el (commented):
I’m not a huge fan of this deletion process. We generally aim to keep operations simple and atomic, allowing them to be batched together to achieve the desired state change. Introducing so much logic for something so minor doesn’t seem worth it to me. I also dnt think this design would  improve the user experience.

We should avoid introducing inflationary storage maps that are unknown to the pallet. This doesn’t feel like a good Substrate practices and should only be maybe only considered as a last resort.

I would strongly prefer a simple hook that determines whether a DID can be deleted or not. Any error returned to the user should provide enough guidance to make it clear what steps they need to take.

Sorry for not paying closer attention to the discussion earlier - I would have raised my concerns sooner.

pallets/did/src/lib.rs
ntn-x2: Yes this would need to be refined properly, this is just a pre-draft, but I wanted to start a conversation about it.

ntn-x2:
@Ad96el I totally agree with you, @rflechtner was of a different opinion tho. So I think it's worth discussing about this a bit further, and then choose a strategy. I also prefer not overcomplicating the runtime and simply fail, and let the client figure out what needs to happen before, with a proper strategy that perhaps combines a good error message + a support runtime API. Let's give @rflechtner one more chance to express his opinion before we decide the way forward.

rflechtner:
I am not opposed to just rejecting attempts to delete a DID in order to keep things simple, but let me be clear that the client will _NOT_ solve this for the user, so it's going to be the user who's got to figure out what needs to be done. Good, expressive error messages are vital in this case, which could be a problem given that substrate errors are pretty limited - there's no more info than can be expressed in an enum variant, is there?

The only runtime api that I see as helpful would be the one providing extrinsics via a runtime api. I generally like this idea, still it gives me a bit of a headache because it seems really dangerous to sign transactions composed by external clients, and can only be really recommended if you run a node yourself.

Generally I think this discussion is exactly the of the kind of questions that we would tackle in the tech board.

ntn-x2:
@rflechtner 

> Good, expressive error messages are vital in this case, which could be a problem given that substrate errors are pretty limited - there's no more info than can be expressed in an enum variant, is there?

I just went over all the FRAME pallets, and there's no single instance where what's returned as part of a _pallet_ error is more than a simple enum variant, with no additional info. But that applies for pallet errors, meaning, we can have a runtime API `check_deletion_requirements` - naming to be changed -  which can return expressive errors.

> The only runtime api that I see as helpful would be the one providing extrinsics via a runtime api. I generally like this idea, still it gives me a bit of a headache because it seems really dangerous to sign transactions composed by external clients, and can only be really recommended if you run a node yourself.

This can be left to clients to decide what they want to do with this runtime API. I think it helps to have it there, but of course we will put a disclaimer that the result of this call still has to be checked before being taken, signed and submitted. Honestly, I don't think this is any different than the security assumptions we have today, especially since the check for the metadata Merkle root: we are relying on the full node to give us the right metadata. Nevertheless, a node could give us malicious metadata and we could be signing a transfer tx when we think we're trying to claim a w3n.

In general, I would see perhaps two different APIs: one that returns a list of "linked" elements, which is trivially an enum that we can expand on the more DID-linked stuff we add, and the second is a vec of `Call`s containing the `delete` operation for each of those elements. Clients (or users for what matter) can do as they please with the information returned, knowing that if any of those returns a non-empty collection, the `did.delete` call will fail.

ntn-x2:
Closing in favor of https://github.com/KILTprotocol/kilt-node/pull/843.

PR #820, opened 2024-11-29
bench: add benchmarks for new deployments for Peregrine runtime

ntn-x2:
It just updates the weight files for the two new pallets, so that it can be actually used.


ntn-x2:
@Ad96el FYI I'm merging this.

Ad96el:
LGTM

PR #821, opened 2024-12-02
chore: update srtool version to show correct WASM meta hashes

ntn-x2:
WIP. Fixes https://github.com/KILTprotocol/ticket/issues/3162.


ntn-x2:
Opened a PR to srtool to address an issue with the `wasm32-unknown-unknown` target when overriding the `RUSTC` version: https://github.com/paritytech/srtool/pull/90

PR #822, opened 2024-12-02
refactor: get next asset id via hook

rflechtner:
Adresses the discussion at https://github.com/KILTprotocol/kilt-node/pull/764#discussion_r1862257153

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/traits.rs
+++ b/pallets/pallet-bonded-coins/src/traits.rs
@@ -83,7 +83,10 @@ where
 	}
 }
 
+/// A trait for getting the next n asset ids to be used during pool creation.
 pub trait NextAssetIds<T: Config> {
-	type Error: Into<DispatchError> + Debug;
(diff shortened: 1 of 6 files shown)


pallets/pallet-bonded-coins/src/traits.rs
```diff
@@ -82,3 +82,11 @@ where
 		AssetsPallet::<T, I>::set_team(origin.into(), id.into(), issuer.into(), admin.into(), freezer.into())
 	}
 }
+
+/// A trait for getting the next n asset ids to be used during pool creation.
+pub trait NextAssetIds<T: Config> {
+	/// Generic error type.
+	type Error: Into<DispatchError>;
+	/// Get the next `n` asset ids.
+	fn try_get(n: u32) -> Result<BoundedVec<FungiblesAssetIdOf<T>, T::MaxCurrencies>, Self::Error>;
```
ntn-x2, line 91: The `n` parameter already contains the number of IDs to return, so I would have not returned a `BoundedVec` but a simple `Vec`. This is a nit pick but it's also ok the way it is.

ntn-x2 (commented):
LFG 🔥

PR #823, opened 2024-12-03
feat: implement `MutateHold` for `pallet-deposit-storage`

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3702.

This PR implements the `MutateHold` trait for the deposit storage pallet by introducing a `PalletDepositStorageReason` type that, in the runtime, is required to implement the runtime's `RuntimeHoldReason`.

Because the deposit pallet assumes that the deposit payer can reclaim the deposit back, we can't add the `MutateHold` implementation without either 1. requiring a migration to distinguish which deposits are reclaimable and which are not, or 2. introducing a new storage map that only keeps track of "system" deposits, aka deposits that are part of some runtime logic and that are only supposed to be changed by interacting with the `MutateHold` trait and not directly via the pallet exposed extrinsic(s).

## How to use it in the bonding curves

When switching to the deposit pallet as the deposit fungible, the bonding curve pallet has to define a namespace, and a unique key per deposit to be taken. For instance, the tuple (asset ID, account ID) could be such a unique key. Example of integration PR: https://github.com/KILTprotocol/kilt-node/pull/827. I can provide support for integrating this into the runtime once the runtimes have been updated to have the required pallets.

### DIFF
--- a/pallets/pallet-deposit-storage/src/fungible/mod.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/mod.rs
@@ -39,7 +39,7 @@ where
 	type Balance = <T::Currency as Inspect<T::AccountId>>::Balance;
 
 	fn total_issuance() -> Self::Balance {
-		<T::Currency as Inspect<T::AccountId>>::minimum_balance()
+		<T::Currency as Inspect<T::AccountId>>::total_issuance()
 	}
 
 	fn minimum_balance() -> Self::Balance {
--- a/pallets/pallet-deposit-storage/src/fungible/mod.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/mod.rs
@@ -25,7 +25,7 @@ use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_runtime::{traits::Zero, DispatchError, DispatchResult};
 
-use crate::{deposit::DepositEntry, Config, DepositKeyOf, HoldReason, Pallet, SystemDeposits};
+use crate::{deposit::DepositEntry, Config, DepositKeyOf, Error, HoldReason, Pallet, SystemDeposits};
 
 #[cfg(test)]
 mod tests;
@@ -81,13 +81,12 @@ pub struct PalletDepositStorageReason<Namespace, Key> {
 
 impl<Namespace, Key> From<PalletDepositStorageReason<Namespace, Key>> for HoldReason {
 	fn from(_value: PalletDepositStorageReason<Namespace, Key>) -> Self {
-		// All the deposits ever taken like this will count towards the same hold reason.
+		// All the deposits ever taken like this will count towards the same hold
+		// reason.
 		Self::Deposit
 	}
 }
 
-// This trait is implemented by forwarding everything to the `Currency`
-// implementation.
 impl<T> InspectHold<T::AccountId> for Pallet<T>
 where
 	T: Config,
@@ -98,11 +97,17 @@ where
 		<T::Currency as InspectHold<T::AccountId>>::total_balance_on_hold(who)
 	}
 
+	// We return the balance of hold of a specific (namespace, key), which prevents
+	// mistakes since everything in the end is converted to `HoldReason`.
 	fn balance_on_hold(reason: &Self::Reason, who: &T::AccountId) -> Self::Balance {
-		<T::Currency as InspectHold<T::AccountId>>::balance_on_hold(
-			&T::RuntimeHoldReason::from(reason.clone().into()),
-			who,
-		)
+		let Some(deposit_entry) = SystemDeposits::<T>::get(&reason.namespace, &reason.key) else {
+			return 0u32.into();
+		};
+		if deposit_entry.deposit.owner == *who {
+			deposit_entry.deposit.amount
+		} else {
+			0u32.into()
+		}
 	}
 }
 
@@ -137,7 +142,7 @@ where
 			who,
 			amount,
 		)?;
-		SystemDeposits::<T>::mutate_exists(&reason.namespace, &reason.key, |maybe_existing_deposit_entry| {
+		SystemDeposits::<T>::try_mutate_exists(&reason.namespace, &reason.key, |maybe_existing_deposit_entry| {
 			match maybe_existing_deposit_entry {
 				// If this is the first registered deposit, create a new storage entry.
 				None => {
@@ -150,18 +155,23 @@ where
 							reason: T::RuntimeHoldReason::from(reason.clone().into()),
 						})
 					}
+					Ok(())
 				}
-				// If this is a previous deposit, update the storage entry or remove completely if balance on hold is
-				// set to `0`.
 				Some(existing_deposit_entry) => {
+					// The pallet assumes each (namespace, key) points to a unique deposit, so the
+					// same combination cannot be used for a different account.
+					if existing_deposit_entry.deposit.owner != *who {
+						return Err(Error::<T>::DepositExisting);
+					}
 					if amount.is_zero() {
 						*maybe_existing_deposit_entry = None;
 					} else {
 						existing_deposit_entry.deposit.amount = amount;
 					}
+					Ok(())
 				}
 			}
-		});
+		})?;
 		Ok(())
 	}
 }
--- a/pallets/pallet-deposit-storage/src/fungible/mod.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/mod.rs
@@ -128,10 +128,11 @@ where
 					if amount > Zero::zero() {
 						*maybe_existing_deposit_entry = Some(DepositEntry {
 							deposit: Deposit {
-								amount: amount,
+								amount,
 								owner: who.clone(),
 							},
 							reason: reason.clone().into(),
+							reclaimable: false,
 						})
 					}
 				}
--- a/pallets/pallet-deposit-storage/src/fungible/mod.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/mod.rs
@@ -18,14 +18,19 @@
 
 use frame_support::traits::{
 	fungible::{Dust, Inspect, InspectHold, MutateHold, Unbalanced, UnbalancedHold},
-	tokens::{Fortitude, Preservation, Provenance, WithdrawConsequence},
+	tokens::{Fortitude, Precision, Preservation, Provenance, WithdrawConsequence},
 };
 use kilt_support::Deposit;
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
-use sp_runtime::{traits::Zero, DispatchError, DispatchResult};
+use sp_runtime::{
+	traits::{CheckedSub, Zero},
+	DispatchError, DispatchResult,
+};
 
 use crate::{deposit::DepositEntry, Config, DepositKeyOf, Error, HoldReason, Pallet, SystemDeposits};
+
+const LOG_TARGET: &str = "runtime::pallet_deposit_storage::mutate-hold";
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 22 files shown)


pallets/pallet-deposit-storage/src/lib.rs
```diff
@@ -171,6 +174,20 @@ pub mod pallet {
 		DepositEntryOf<T>,
 	>;
 
+	/// Storage of all system deposits. They are the same as user deposits, but
+	/// cannot be claimed back by the payers. Instead, some on chain logic must
+	/// trigger their release.
+	#[pallet::storage]
+	#[pallet::getter(fn system_deposits)]
+	pub(crate) type SystemDeposits<T> = StorageDoubleMap<
```
ntn-x2, line 182: Changing the existing storage would require a migration. I think it's anyway better to divide between "system" deposits which can only be released by the runtime and "regular" deposits that can be claimed back by the user.

rflechtner: Fine by me! I don't know much about the architecture of the deposit pallet, but if hold reasons are sufficiently informative to identify the pallet, and if there's a hook that's executed whenever someone tries to reclaim a deposit through the pallet (instead of the release being triggered by another pallet), then I could also see that this may be configured to simply fail when the deposit relates to the bonding curve pallet (for example) _OR_ to try and call `finish_destroy` there, which would lead to the deposit being reclaimed unless the pool is not ready to be destroyed.

ntn-x2: Sorry I am not sure I am able to fully understand your message. If pallets interact with the deposit storage pallet via the `MutateHold` trait, then any deposit that needs to be removed would have to go through some other runtime components. The deposit storage pallet ALSO exposes one extrinsic of its own, which allows a deposit to be reclaimed back by the user. To be honest, I would simply, in the future, remove such an extrinsic and "force" all pallets to interact with the deposit storage via simply the `MutateHold` trait, without any complex hooks and similar. So to your point, if you call `T::MutateHold::release(<right_unique_reason>, <user_address>)`, it should be enough.

pallets/pallet-deposit-storage/src/fungible/tests/burn.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[test]
+fn burn_held() {
+	ExtBuilder::default()
+		.with_balances(vec![(OWNER, 100_000)])
+		.build()
+		.execute_with(|| {
+			let reason = PalletDepositStorageReason::default();
+
+			<Pallet<TestRuntime> as MutateHold<AccountId32>>::hold(&reason, &OWNER, 10)
```
Ad96el, line 43: Using `StorageDepositPallet` would look simpler

pallets/pallet-deposit-storage/src/fungible/tests/inspect_hold.rs
```diff
@@ -0,0 +1,60 @@
...
+use frame_support::traits::fungible::{InspectHold, MutateHold};
+use sp_runtime::{traits::Zero, AccountId32};
+
+use crate::{
+	fungible::tests::mock::{DepositNamespace, ExtBuilder, TestRuntime, OTHER_ACCOUNT, OWNER},
+	Pallet, PalletDepositStorageReason,
+};
+
+#[test]
+fn balance_on_hold() {
```
Ad96el, line 28: can we also maybe add a test for creating a second hold with a different `HoldReason`? I would expect two SystemDeposits accumulated into the one holdReason in the Currency pallet. You prob have to change the mock setup

pallets/pallet-deposit-storage/src/fungible/tests/transfer.rs
```diff
@@ -0,0 +1,102 @@
...
+	ExtBuilder::default()
+		.with_balances(vec![(OWNER, 100_000), (OTHER_ACCOUNT, 1)])
+		.build()
+		.execute_with(|| {
+			let reason = PalletDepositStorageReason::default();
+
+			<Pallet<TestRuntime> as MutateHold<AccountId32>>::hold(&reason, &OWNER, 10)
+				.expect("Failed to hold amount for user.");
+
+			<Pallet<TestRuntime> as MutateHold<AccountId32>>::transfer_on_hold(
```
Ad96el, line 46: What is the expected behavior of transfer_on_hold?

Ad96el (commented):
Love it

pallets/pallet-deposit-storage/src/fungible/tests/transfer.rs
ntn-x2: From the docs:

```
        /// Transfer held funds into a destination account.
	///
	/// If `mode` is `OnHold`, then the destination account must already exist and the assets
	/// transferred will still be on hold in the destination account. If not, then the destination
	/// account need not already exist, but must be creatable.
	///
	/// If `precision` is `BestEffort`, then an amount less than `amount` may be transferred without
	/// error.
	///
	/// If `force` is `Force`, then other fund-locking mechanisms may be disregarded. It should be
	/// left as `Polite` in most circumstances, but when you want the same power as a `slash`, it
	/// may be `Force`.
	///
	/// The actual amount transferred is returned, or `Err` in the case of error and nothing is
	/// changed.
```

ntn-x2: To be fair, I did not extensively test all the `MutateHold` functions. It is mostly like a smoke test to make sure we don't deviate from the "traditional" implementation.

pallets/pallet-deposit-storage/src/fungible/tests/inspect_hold.rs
ntn-x2: But why would you test the `Currency` pallet interactions?

pallets/pallet-deposit-storage/src/fungible/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+
+	fn total_balance_on_hold(who: &T::AccountId) -> Self::Balance {
+		<T::Currency as InspectHold<T::AccountId>>::total_balance_on_hold(who)
+	}
+
+	// We return the balance of hold of a specific (namespace, key), which prevents
+	// mistakes since everything in the end is converted to `HoldReason`.
+	fn balance_on_hold(reason: &Self::Reason, who: &T::AccountId) -> Self::Balance {
+		let Some(deposit_entry) = SystemDeposits::<T>::get(&reason.namespace, &reason.key) else {
+			return 0u32.into();
```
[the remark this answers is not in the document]
ntn-x2: We don't know the concrete type, so we can't use `u128`. But good suggestion with the `Zero` trait: https://github.com/KILTprotocol/kilt-node/pull/823/commits/3202fba4a67df1e05f6598fd50c68f8ccd5c6daf.

pallets/pallet-deposit-storage/src/fungible/tests/burn.rs
Ad96el: yes but its a minor thing

pallets/pallet-deposit-storage/src/fungible/tests/inspect_hold.rs
Ad96el: Actually, that is possible. The deposit is storing the account.

Ad96el: as a try_runtime test

ntn-x2: You helped me identify a very big bug. I addressed it in here: https://github.com/KILTprotocol/kilt-node/pull/823/commits/6f3e1ace0d705d8cedc9a5b12e97a5b4e0d2e641.

Simply we can't forward the `set_balance_on_hold` to the underlying currency directly, as that would override the balance for the overarching `HoldReason::Deposit`. What we need to do (I've done it in the commit), is to call either `increase_balance_on_hold` or `decrease_balance_on_hold` depending on weather we are reserving or releasing holds.

This was a very very good catch that I did not write a test for before, but now I did.

ntn-x2: Hey, I opened a new PR on top of this where I add `try-state` checks. Please check that out, so we can merge that on top of this before merging this. Now I feel much more confident everything works as expected, also thanks to you 😀

https://github.com/KILTprotocol/kilt-node/pull/830

pallets/pallet-deposit-storage/src/fungible/tests/burn.rs
ntn-x2: I think I'll skip this, as we have both approaches across our codebase, so I think it's not too big of a deal. I personally prefer the current approach, which, once again, is more Rusty instead of relying on the result of the `construct_runtime` macro.

pallets/pallet-deposit-storage/src/fungible/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+	/// value of `1`.
+	fn set_balance_on_hold(reason: &Self::Reason, who: &T::AccountId, amount: Self::Balance) -> DispatchResult {
+		SystemDeposits::<T>::try_mutate_exists(&reason.namespace, &reason.key, |maybe_existing_deposit_entry| {
+			match maybe_existing_deposit_entry {
+				// If this is the first registered deposit for this reason, create a new storage entry.
+				None => {
+					if amount > Zero::zero() {
+						// Increase the held amount for the final runtime hold reason by `amount`.
+						<T::Currency as UnbalancedHold<T::AccountId>>::increase_balance_on_hold(
+							&T::RuntimeHoldReason::from(reason.clone().into()),
```
Ad96el, line 157: I see multiple clone().into() 

can we maybe move that into a variable?

pallets/pallet-deposit-storage/src/fungible/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+					} else {
+						// If trying to increase the amount held, increase the held amount for the final
+						// runtime hold reason by the (amount - stored amount) difference.
+						if amount >= existing_deposit_entry.deposit.amount {
+							let amount_to_hold = amount
+								.checked_sub(&existing_deposit_entry.deposit.amount)
+								.ok_or_else(|| {
+									log::error!(target: LOG_TARGET, "Failed to evaluate {:?} - {:?}", amount, existing_deposit_entry.deposit.amount);
+									Error::<T>::Internal
+								})?;
```
Ad96el, line 201: From my point of view, we can also `saturating_sub` here. It is impossible that this operation can fail. We checked the conditions. But it is also fine for me to leave it like that.

Ad96el (dismissed):
Two minior things

pallets/pallet-deposit-storage/src/fungible/mod.rs
ntn-x2: I mean, the runtime cost of both is the same since they check for underflows, but while with saturating we don't know for sure, with checked we do, so why not using the checked version? We've done that elsewhere as well, in case invariants are broken.

ntn-x2: It's also good for tests, since the code could change and you could start triggering `silent` saturating stuff, while with this you'd get an explicit `Internal` error, so I don't see any advantages of using saturating.

pallets/pallet-deposit-storage/src/fungible/mod.rs
ntn-x2: Good catch! https://github.com/KILTprotocol/kilt-node/pull/823/commits/44d9745f80725462d4ca86e1d98692879448e775

pallets/pallet-deposit-storage/src/fungible/mod.rs
Ad96el: If possible, I would be curious about the case.

PR #826, opened 2024-12-04
fix: check return asset ids from hook

Ad96el:
Based on the discussion: https://github.com/KILTprotocol/kilt-node/pull/818#discussion_r1869222126, we have to check whether the `NextAssetIds` hook is returning the right amount of asset ids.

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -363,8 +363,15 @@ pub mod pallet {
 
 			let currency_length = currencies.len();
 
-			let currency_ids: BoundedVec<FungiblesAssetIdOf<T>, T::MaxCurrenciesPerPool> =
-				T::NextAssetIds::try_get(currency_length.saturated_into()).map_err(|err| err.into())?;
+			let currency_ids = T::NextAssetIds::try_get(currency_length.saturated_into())
+				.map_err(|_| Error::<T>::Internal)
+				.and_then(|ids| {
+					if ids.len() != currency_length {
+						log::error!(target: LOG_TARGET, "NextAssetIds::try_get returned wrong number of ids");
+						return Err(Error::<T>::Internal);
+					}
+					Ok(ids)
+				})?;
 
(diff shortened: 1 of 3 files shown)


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -363,8 +363,15 @@ pub mod pallet {
 
 			let currency_length = currencies.len();
 
-			let currency_ids: BoundedVec<FungiblesAssetIdOf<T>, T::MaxCurrenciesPerPool> =
-				T::NextAssetIds::try_get(currency_length.saturated_into()).map_err(|err| err.into())?;
+			let currency_ids = T::NextAssetIds::try_get(currency_length.saturated_into())
+				.map_err(|_| Error::<T>::Internal)
```
rflechtner, line 367: Still very much opposed to the overuse of non-informative "Internal" errors. Why mask hook errors instead of showing what is actually going on? If in the hook I throw an `Arithmetic::Overflow` because the space of ids is exhausted, I would expect this to be seen by the caller instead of a non-informative "well, that wasn't supposed to happen".

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -363,8 +363,15 @@ pub mod pallet {
...
+				.and_then(|ids| {
+					if ids.len() != currency_length {
+						log::error!(target: LOG_TARGET, "NextAssetIds::try_get returned wrong number of ids");
+						return Err(Error::<T>::Internal);
+					}
```
rflechtner, line 372: this, on the other hand, makes sense, because it indicates inactionable misconfiguration.

rflechtner (commented):
Should we also re-type the hook's return type from a BoundedVec to a simple Vec here?

Ad96el:
I also changed the trait in: [9baa808](https://github.com/KILTprotocol/kilt-node/pull/826/commits/9baa80884e29903420bdc5efaf6efd19c315de7b). Let's merge this first before merging the runtime implementation.

ntn-x2 (approved):

rflechtner (approved):
❤️

PR #827, opened 2024-12-04
[DO NOT MERGE] Example adaptation of bonding curve hold reason generation

ntn-x2:
Example PR. For more context refer to https://github.com/KILTprotocol/kilt-node/pull/823.

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -192,7 +192,11 @@ pub mod pallet {
 		/// The type used for pool ids
 		type PoolId: Parameter + MaxEncodedLen + From<[u8; 32]> + Into<Self::AccountId>;
 
-		type RuntimeHoldReason: From<HoldReason>;
+		/// Reason for each deposit. The combination of (pool ID, account ID)
+		/// must be unique for each individual deposit across all the bonding
+		/// curves.
+		type HoldReason: From<(Self::PoolId, Self::AccountId)>;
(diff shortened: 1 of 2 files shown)


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -1274,8 +1279,9 @@ pub mod pallet {
 
 			Pools::<T>::remove(&pool_id);
 
+			let hold_reason = T::HoldReason::from((pool_id.clone(), pool_details.owner.clone()));
 			T::DepositCurrency::release(
-				&T::RuntimeHoldReason::from(HoldReason::Deposit),
+				&T::RuntimeHoldReason::from(hold_reason),
```
rflechtner, line 1284: given that the hold reason is unique to a pool, we wouldn't even have to store the amount held in this pallet any more, right? We could just release any amount held under this reason?

ntn-x2: Exactly. That's the whole point why I tried to make the deposit storage pallet work. Everything will be stored in there under the `BondedCurves` namespace and `(pool_id, owner)` key, or even simply `pool_id` now that I think about it, if there is a single deposit taken for each pool and each pool ID is unique.

ntn-x2: Plus you don't need to store the owner anymore, as you can use the `InspectHold` trait for a specific reason and user, to know whether that user has a reserved deposit for the reason or not.

PR #830, opened 2024-12-07
test: add try-state tests for pallet-deposit-storage

ntn-x2:
Based on top of https://github.com/KILTprotocol/kilt-node/pull/823.

It adds `try-runtime` checks for both the `MutateHold` implementation as well as the rest of the deposits, so that one feature does not interfere with the other.

### DIFF
--- a/pallets/pallet-deposit-storage/src/fungible/mod.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/mod.rs
@@ -35,6 +35,9 @@ const LOG_TARGET: &str = "runtime::pallet_deposit_storage::mutate-hold";
 #[cfg(test)]
 mod tests;
 
+#[cfg(any(test, feature = "try-runtime"))]
+pub(super) mod try_state;
+
 // This trait is implemented by forwarding everything to the `Currency`
 // implementation.
 impl<T> Inspect<T::AccountId> for Pallet<T>
@@ -88,7 +91,7 @@ impl<Namespace, Key> From<PalletDepositStorageReason<Namespace, Key>> for HoldRe
 	fn from(_value: PalletDepositStorageReason<Namespace, Key>) -> Self {
 		// All the deposits ever taken like this will count towards the same hold
 		// reason.
-		Self::Deposit
+		Self::FungibleImpl
 	}
 }
 
--- a/pallets/pallet-deposit-storage/src/fungible/try_state.rs
+++ b/pallets/pallet-deposit-storage/src/fungible/try_state.rs
@@ -1,0 +1,65 @@
+// KILT Blockchain – https://botlabs.org
+// Copyright (C) 2019-2024 BOTLabs GmbH
+
+// The KILT Blockchain is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The KILT Blockchain is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+use std::collections::BTreeMap;
+
+use frame_support::{ensure, traits::fungible::InspectHold};
+use frame_system::pallet_prelude::BlockNumberFor;
+use kilt_support::Deposit;
+use sp_runtime::{traits::CheckedAdd, TryRuntimeError};
+
+use crate::{deposit::DepositEntry, AccountIdOf, BalanceOf, Config, HoldReason, SystemDeposits};
+
+pub(crate) fn check_fungible_consistency<T>(_n: BlockNumberFor<T>) -> Result<(), TryRuntimeError>
+where
+	T: Config,
+{
+	// Sum together all the deposits stored as part of the `MutateHold`
+	// implementation, and fail if any of them does not have the expected
+	// `crate::HoldReason::FungibleImpl` reason.
+	let sum_of_deposits = SystemDeposits::<T>::iter_values().try_fold(
+		BTreeMap::<AccountIdOf<T>, BalanceOf<T>>::new(),
+		|mut sum,
+		 DepositEntry {
+		     reason,
+		     deposit: Deposit { amount, owner },
+		 }| {
+			ensure!(
+				reason == HoldReason::FungibleImpl.into(),
+				TryRuntimeError::Other("Found a deposit reason different than the expected `HoldReason::FungibleImpl`")
+			);
+
+			// Fold the deposit amount for the current user.
+			sum.entry(owner)
+				.and_modify(|s| *s = s.checked_add(&amount).expect("Failed to fold deposits for user."));
+
+			Ok::<_, TryRuntimeError>(sum)
+		},
+	)?;
+	// We verify that the total balance on hold for the `HoldReason::FungibleImpl`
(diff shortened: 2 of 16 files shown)


pallets/pallet-deposit-storage/src/fungible/mod.rs
```diff
@@ -88,7 +91,7 @@ impl<Namespace, Key> From<PalletDepositStorageReason<Namespace, Key>> for HoldRe
 	fn from(_value: PalletDepositStorageReason<Namespace, Key>) -> Self {
 		// All the deposits ever taken like this will count towards the same hold
 		// reason.
-		Self::Deposit
+		Self::FungibleImpl
```
ntn-x2, line 94: This helps keeping the two functionalities of "regular" deposit and "system" deposits completely separated from each other, and having two different reasons also allows us to write much more accurate and strict `try-state` checks.

pallets/pallet-deposit-storage/src/fungible/try_state.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) fn check_fungible_consistency<T>(_n: BlockNumberFor<T>) -> Result<(), TryRuntimeError>
+where
+	T: Config,
+{
+	// Sum together all the deposits stored as part of the `MutateHold`
+	// implementation, and fail if any of them does not have the expected
+	// `crate::HoldReason::FungibleImpl` reason.
+	let sum_of_deposits = SystemDeposits::<T>::iter_values().try_fold(
+		BTreeMap::<AccountIdOf<T>, BalanceOf<T>>::new(),
+		|mut sum,
```
Ad96el, line 38: Its rather `user_map`

pallets/pallet-deposit-storage/src/fungible/try_state.rs
```diff
@@ -0,0 +1,66 @@
...
+			ensure!(
+				reason == HoldReason::FungibleImpl.into(),
+				TryRuntimeError::Other("Found a deposit reason different than the expected `HoldReason::FungibleImpl`")
+			);
+
+			// Fold the deposit amount for the current user.
+			let entry = sum.entry(owner).or_default();
+			*entry = entry.checked_add(&amount).expect("Failed to fold deposits for user.");
+
+			Ok::<_, TryRuntimeError>(sum)
```
Ad96el, line 52: I kinda prefer a function definition then the type injection here | sum, DepositEntry {. . }| -> Result<BtreeMap<..>, TryRuntimeError>. But it is a minor thing.

pallets/pallet-deposit-storage/src/lib.rs
```diff
@@ -115,6 +118,7 @@ pub mod pallet {
 	#[pallet::composite_enum]
 	pub enum HoldReason {
 		Deposit,
+		FungibleImpl,
```
Ad96el, line 121: I think a better name would be `SystemDeposit`

pallets/pallet-deposit-storage/src/try_state.rs
```diff
@@ -0,0 +1,91 @@
...
+		BTreeMap::<(AccountIdOf<T>, Vec<u8>), BalanceOf<T>>::new(),
+		|mut sum,
+		 DepositEntry {
+		     reason,
+		     deposit: Deposit { amount, owner },
+		 }| {
+			// Regular deposits should not interfere with the `MutateHold` implementation
+			// state.
+			ensure!(
+				reason != HoldReason::FungibleImpl.into(),
```
Ad96el, line 65: I think, if we ever add a new `Reason` we still want to only check the Deposit reason right? maybe we can make this check explicit.

Ad96el (commented):
Some minor comments

pallets/pallet-deposit-storage/src/fungible/try_state.rs
ntn-x2: I think I disagree. `sum_of_deposits` might not be the best name, but what does `user_map` tell you about the variable? You already know it's a map since Rust is a strongly typed language.

pallets/pallet-deposit-storage/src/try_state.rs
ntn-x2: Sorry, I am not sure I understand your point?

pallets/pallet-deposit-storage/src/lib.rs
ntn-x2: Mhh ok, but what if the pallet will implement a new trait, and we need a new deposit reason? In that case, it would need to be different than this one, while it could still be stored within the same storage map.

pallets/pallet-deposit-storage/src/fungible/try_state.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/830/commits/f9d2aa7af4c8dda7c250f5a0442bfdccf1b7b2cd

ntn-x2: Ops, I forgot it in the other function: https://github.com/KILTprotocol/kilt-node/pull/830/commits/cee766cd9e20f814cc2b6535fa1cb789719839f0

pallets/pallet-deposit-storage/src/fungible/try_state.rs
Ad96el: Yeah, okay. I think the variable is showing the balance of the user. So maybe `users_balance`? But feel free to resolve this conversation.

pallets/pallet-deposit-storage/src/try_state.rs
Ad96el: `Reason == HoldReason::FigibleImpl` or is this not intendant? If yes, what if a new Reason is introduced?

ntn-x2: We check for `reason != HoldReason::FungibleImpl.into(),`, at least in this line, so I still don't get your point. Here we're only interested in checking that `HoldReason::FungibleImpl` is only used for the trait implementation. If there's a new deposit reason, we don't care.

pallets/pallet-deposit-storage/src/fungible/try_state.rs
ntn-x2: Renamed to `users_deposits` in https://github.com/KILTprotocol/kilt-node/pull/830/commits/ff293323c921a3c8270d60af2e56a59d28b97180.

pallets/pallet-deposit-storage/src/lib.rs
Ad96el: Okay, maybe I am missing something. I was thinking the `FungibleImpl` reason is introduced to separate regular and system deposits from each other.

Can you perhaps explain what you mean with 'implementing a new trait'?

ntn-x2: `FungibleImpl`, in my opinion, is to "namespace" deposits taken when implementing the `MutateHold` trait. If we were to implement, let's say, the `fungibles::MutateHold` trait tomorrow, I would rather see a new hold reason `FungiblesImpl` which we would use when implementing that trait. Does it make sense?

ntn-x2: After all, that's what hold reasons are for, right? So that one implementation does not interfere (i.e., cannot release) deposits taken with a different hold reason, i.e., with a different trait implementation.

pallets/pallet-deposit-storage/src/try_state.rs
Ad96el: Okay I see

pallets/pallet-deposit-storage/src/lib.rs
ntn-x2: i hope this clarifies better the use of the `HoldReason` for this pallet?

Ad96el: Okay, I think I understand. Fine for me :+1:

Ad96el (approved):

PR #835, opened 2024-12-19
feat: add support for metadata hash checks

ntn-x2:
We introduce a new feature for peregrine and spiritnet runtimes, which adds the metadata hash signed extension, for wallets that need it.

The Gitlab build pipeline has also been updated to include the new feature only when building the production WASMs.

## How to test

Change any integration tests in the SDK based on this PR (https://github.com/KILTprotocol/sdk-js/pull/924), and verify the metadata is verified (e.g., by enabling logs via a Chopsticks deployment).

### DIFF
--- a/runtimes/spiritnet/build.rs
+++ b/runtimes/spiritnet/build.rs
@@ -19,9 +19,14 @@
 use substrate_wasm_builder::WasmBuilder;
 
 fn main() {
-	WasmBuilder::new()
+	let builder = WasmBuilder::new()
 		.with_current_project()
 		.export_heap_base()
-		.import_memory()
-		.build()
+		.import_memory();
+
+	#[cfg(feature = "metadata-hash")]
+	// TODO: Can we re-use some consts?
+	let builder = builder.enable_metadata_hash("PILT", 15);
+
+	builder.build()
 }
--- a/runtimes/spiritnet/build.rs
+++ b/runtimes/spiritnet/build.rs
@@ -26,7 +26,7 @@ fn main() {
 
 	#[cfg(feature = "metadata-hash")]
 	// TODO: Can we re-use some consts?
-	let builder = builder.enable_metadata_hash("PILT", 15);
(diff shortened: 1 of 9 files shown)


ntn-x2:
Ccing @rflechtner here cause it might be interesting for him as well.

ntn-x2:
There's a unit test by Frequency that manually computes the metadata hash and includes it in the `SignedExtra`: https://github.com/frequency-chain/frequency/blob/main/e2e/signed-extensions/checkMetadataHash.test.ts.

runtimes/spiritnet/build.rs
```diff
@@ -19,9 +19,14 @@
...
+	let builder = WasmBuilder::new()
 		.with_current_project()
 		.export_heap_base()
-		.import_memory()
-		.build()
+		.import_memory();
+
+	#[cfg(feature = "metadata-hash")]
+	// TODO: Can we re-use some consts?
+	let builder = builder.enable_metadata_hash("PILT", 15);
```
Ad96el, line 29: "KILT"

Ad96el: You should maybe use the `MetadataProvider` in the kilt module

Ad96el (commented):
Nice changes. Does this require some updates on the client side? Or is the metadata hash already present in the signed extra of the tx the nodes are currently ignoring them?

Ad96el:
According to the documentation, the tx fmt is defined in the metadata. So there is no extra step to take for us, I think.

runtimes/spiritnet/build.rs
ntn-x2: This is a build script, it can't use stuff from libraries.

ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/835/commits/0a1f2d006d8977f8432c5c5918edc1ddf9190661

Ad96el (approved):

PR #836, opened 2025-01-06
feat: increase string limit 

Ad96el:
# Fixes: [#3746](https://github.com/KILTprotocol/ticket/issues/3746)


ntn-x2:
Do bonded coin symbol and name use the same limit?

Ad96el:
Yes, they share the same limit. In the asset-pallet, there is a single config to limit the length of the symbol and name. Not much what we can do here.

PR #840, opened 2025-01-08
feat: change gatekeeping logic to avoid complex deployment setups

ntn-x2:
Right now, if the new runtime is enacted and there's no authorised account set, the calls are considered permissionless. To avoid this, and to allow us to set an account later on, I changed the logic so that by default it's considered unusable, and only when an account is set, the stored account is checked. I introduced also a new type to allow for permissionless dispatch.

The change does not require any migrations or changes to storage, as it simply introduces a semantic change to how those storage values are interpreted.

## How to test

Use chopsticks to try to claim a DOT name before anything is set, it will fail with `BadOrigin`. Set the new value with a `sudo.setStorage` call, and verify that only that account can submit `dotname.claim` or `uniqueLinking.linkAccount` transactions.

### DIFF
--- a/pallets/did/src/origin.rs
+++ b/pallets/did/src/origin.rs
@@ -49,7 +49,32 @@ impl<DidIdentifier: Clone, AccountId: Clone> CallSources<AccountId, DidIdentifie
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Ad96el (approved):
LGTM

PR #843, opened 2025-01-09
feat: prevent DID deletions if there's linked resources

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3673.

I defined a new collection of hooks for the DID pallet, which can be extended with new functions if we need to check conditions for some other DID operations. I provide a bunch of implementations to check for related resources.

~The only limitation to the current solution is that we don't have a precise worst case weight calculation for the hook, since we would need to benchmark the storage accesses, and the pallets use the old benchmarking logic which is not very flexible. I will work on the benchmarks and on updating the actual weights in a separate PR.~
Apparently, [it's enough to use the expected `MaxEncodedLen` as the proof size](https://substrate.stackexchange.com/questions/11842/calculate-proof-size-of-a-function-that-reads-state/11843#11843), so I switched to that in the [final commit](https://github.com/KILTprotocol/kilt-node/pull/843/commits/72fc198d24b84e19b7d04a1da0134742f0f2b9d4).

## How to test

Spin up a Chopsticks instance, claim a DID, claim a web3name, try to delete the DID: it will fail. Delete the web3name. Try to delete the DID: it will succeed.
The same can be tried with linked accounts.


support/src/mock.rs
```diff
@@ -171,3 +176,109 @@ impl AsRef<[u8]> for SubjectId {
 		self.0.as_ref()
 	}
 }
+
+// Mock currency that implements all required traits, allowing test runtimes to
```
Ad96el, line 180: Maybe use doc strings

runtimes/spiritnet/src/kilt/did/mod.rs
```diff
@@ -86,6 +91,53 @@ impl DeriveDidCallAuthorizationVerificationKeyRelationship for RuntimeCall {
...
+pub struct DidLifecycleHooks;
+
+impl did::traits::DidLifecycleHooks<Runtime> for DidLifecycleHooks {
+	type DeletionHook = EnsureNoNamesAndNoLinkedAccountsOnDidDeletion;
+}
+
+// Read size is given by the `MaxEncodedLen`: https://substrate.stackexchange.com/a/11843/1795.
+// Since the trait is not `const`, we have unit tests that make sure the
+// `max_encoded_len()` function matches this const.
+const WORST_CASE_WEB3_NAME_STORAGE_READ_SIZE: u64 = 33;
```
Ad96el, line 103: Normally those const are in the runtime common they are also duplicated in the peregrine runtime

runtimes/spiritnet/src/kilt/did/tests/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// If you feel like getting in touch with us, you can do so at info@botlabs.org
+
+mod did_hooks;
```
Ad96el, line 19: I am not saying the test folder is bad here, but we should maybe create a ticket for the other tests. Currently, all size checks are in the root test.rs file. We should split it up.

runtimes/common/src/did/deletion_hooks/tests/mock.rs
```diff
@@ -0,0 +1,231 @@
...
+	type WeightInfo = ();
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = ();
+}
+
+#[derive(Default)]
+pub(super) struct ExtBuilder(
+	Vec<(AccountId32, Option<Web3Name>, bool)>,
+	Vec<(AccountId32, Option<Web3Name>, bool)>,
+);
```
Ad96el, line 167: Not super important, but a struct would be better. I was first confused by the 0 and 1 in the impl ExtBuilder. But as I said not important. Feel free to resolve this comment

runtimes/common/src/did/deletion_hooks/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+	const READ_WEIGHT_TIME: u64,
+	const READ_WEIGHT_SIZE: u64,
+	Web3NameDeployment,
+>(PhantomData<Web3NameDeployment>);
+
+impl<T, const READ_WEIGHT_TIME: u64, const READ_WEIGHT_SIZE: u64, Web3NameDeployment> did::traits::DidDeletionHook<T>
+	for EnsureNoLinkedWeb3NameDeletionHook<READ_WEIGHT_TIME, READ_WEIGHT_SIZE, Web3NameDeployment>
+where
+	T: did::Config + pallet_web3_names::Config<Web3NameDeployment, Web3NameOwner = DidIdentifierOf<T>>,
+	Web3NameDeployment: 'static,
```
Ad96el, line 42: I would have anticipated, the deployment requires the instance.

runtimes/common/src/did/deletion_hooks/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+{
+	const MAX_WEIGHT: Weight = Weight::from_parts(READ_WEIGHT_TIME, READ_WEIGHT_SIZE);
```
Ad96el, line 44: Are you sure this is the accurate MAX_WEIGHT? I think the weight here is a good approximation for the lookup: pallet_web3_names::Names::<T, Web3NameDeployment>::contains_key(did) but not for the function call itself.

Ad96el (commented):
Sorry for the late review. I have some questions

runtimes/common/src/did/deletion_hooks/mod.rs
Ad96el: Oh, okay, I see how this works in the runtime implementation.

runtimes/spiritnet/src/kilt/did/mod.rs
ntn-x2: Yeah that's because the concrete type that is read is runtime-specific. Tomorrow we might want to change the maximum length of a Dotname on Spiritnet, and that const will go up. I would keep them runtime-specific as we don't have the concrete type defined in `runtime-common` but we have it in the runtime.

runtimes/spiritnet/src/kilt/did/tests/mod.rs
ntn-x2: What do you mean? I think we should update the structure once we have more tests to run. Right now with these tests this structure makes sense to me, or not? What would you change with the current set of tests?

pallets/did/src/traits/lifecycle_hooks/deletion/tests.rs
```diff
@@ -0,0 +1,84 @@
...
+struct True;
+
+impl DidDeletionHook<TestRuntime> for True {
+	const MAX_WEIGHT: Weight = Weight::from_all(20);
+
+	fn can_delete(_did: &DidIdentifierOf<TestRuntime>) -> Result<(), Weight> {
+		Ok(())
+	}
+}
+
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/843/commits/8b47e0c72135a22406473568943aa7045306afea

support/src/mock.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/843/commits/f178c82e18668ed97eb552ec736572915c67bbfd

runtimes/common/src/did/deletion_hooks/tests/mock.rs
ntn-x2: https://github.com/KILTprotocol/kilt-node/pull/843/commits/3efbafbaec763c784dc06ddc0ce66ea5162d6962

runtimes/spiritnet/src/kilt/did/tests/mod.rs
ntn-x2: Btw the size checks are inside the `did_hooks.rs` file, not in the root.

Ad96el: Yes the structure makes sense.
I just mean in the root test.rs folder, we have tests like `test_derive_did_verification_relation_single` which should be moved into the did/tests.

runtimes/common/src/did/deletion_hooks/mod.rs
Ad96el: Yes, not much, but it is still not the accurate weight, just a good approximation. 
Let's keep it like that.

runtimes/spiritnet/src/kilt/did/tests/mod.rs
ntn-x2: Oh I see. Made a ticket for it as well: https://github.com/KILTprotocol/ticket/issues/3753.

Ad96el (approved):
LGTM

pallets/did/src/lib.rs
```diff
@@ -1512,6 +1527,17 @@ pub mod pallet {
...
+			// Make sure this check happens after the line where we check if a DID exists,
+			// else we would start getting `CannotDelete` errors when we should be getting
+			// `NotFound`.
+			ensure!(
+				<<T::DidLifecycleHooks as DidLifecycleHooks<T>>::DeletionHook as DidDeletionHook<T>>::can_delete(
+					&did_subject,
+				)
+				.is_ok(),
+				Error::<T>::CannotDelete
+			);
```
rflechtner, line 1539: Not sure if this is feasible in substrate, but I'd much rather raise errors in the hook and have them propagated, so that we can inform callers on which deletion conditions have not been met.

rflechtner: I can see this approach hitting a wall though if we can't come up with a way to define errors that are not already defined in any of the pallets. Substrate seems to be somewhat limited in that regard.

ntn-x2: The only thing we can do is return `u8` or `u16` error codes, which are returned by the runtime hook. Then we would need to come up with a scheme that would be able to encode in 8 or 16 bits all the different "components" that are linked to the DID, to be able to return an aggregate error of all the pending stuff instead of returning the error of the first item that was encountered.

ntn-x2: Since we will add runtime APIs to ease deletion, I would say that a dry run of this extrinsic + a call to the API to return the list of call is sufficient to solve this issue. I would not over-complicate the runtime code.

PR #847, opened 2025-01-15
chore: change struct order

Ad96el:
In SCALE encoding, the struct name is irrelevant for the encoding, only the order matters. By changing the order of the properties in the events, we can hopefully reduce the amount of breaking changes.

### DIFF
--- a/pallets/attestation/src/lib.rs
+++ b/pallets/attestation/src/lib.rs
@@ -242,10 +242,10 @@ pub mod pallet {
 			/// The claim hash of the credential for which the attestation entry
 			/// was deleted.
 			claim_hash: ClaimHashOf<T>,
-			/// Who authorized the deletion of the attestation.
-			authorized_by: AuthorizedByOf<T>,
 			/// The ctype of the attested credential.
 			ctype_hash: CtypeHashOf<T>,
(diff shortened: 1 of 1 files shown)


pallets/attestation/src/lib.rs
```diff
@@ -226,26 +226,26 @@ pub mod pallet {
...
 			attester: AttesterOf<T>,
-			/// The ctype of the attested credential.
-			ctype_hash: CtypeHashOf<T>,
 			/// The claim hash of the credential for which the attestation entry
 			/// was deleted.
 			claim_hash: ClaimHashOf<T>,
+			/// Who authorized the deletion of the attestation.
+			authorized_by: AuthorizedByOf<T>,
+			/// The ctype of the attested credential.
+			ctype_hash: CtypeHashOf<T>,
```
rflechtner, line 248: let's stick to the order introduced by AttestationRevoked, which has the exakt same parameters
```suggestion
			/// The ctype of the attested credential.
			ctype_hash: CtypeHashOf<T>,
			/// Who authorized the deletion of the attestation.
			authorized_by: AuthorizedByOf<T>,
```

ntn-x2 (approved):

rflechtner (approved):
LGTM

PR #851, opened 2025-01-16
fix: weight for DID deletion hook

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3752.

I realized I over-complicated the previous solution. We never ever have considered the total size of a storage map in our benchmarks, so I don't see why we would need to start now.

In our case, the deletion hook has the worst case if it returns `true`, which is the default behaviour, hence we don't need any special `setup` logic for the benchmarks. If needed in the future, that is easy to add behind a `feature = "runtime-benchmarks"` cfg instruction.

This solution simplifies weight computation by a lot, since the additional weight will be considered when running the benchmarks already, and does not need any specialised benchmarking logic.


Ad96el (approved):
:neckbeard:

PR #852, opened 2025-01-20
feat: add new DID runtime APIs to aid with DID deletion

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/3742.

The PR introduces two new runtime APIs: one that returns an `enum` of "linked resources", which some clients might find useful to understand why a DID deletion failed, and one that returns a list of `RuntimeCall`s, that could be extended with a `did.delete()` call, put together into a `utility.batch*`, and submitted to the chain, and be sure the DID deletion will work as expected (as long as all the DID origin requirements are met).


Ad96el (approved):

runtimes/common/src/did/runtime_apis.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Debug, Encode, Decode, TypeInfo, PartialEq, Eq, PartialOrd, Ord)]
+pub enum LinkedDidResource<Web3Name, DotName> {
+	/// A Web3name.
+	Web3Name(Web3Name),
+	/// A Dotname.
+	DotName(DotName),
+	/// An account linked to the DID and resolvable by or to a Web3name.
+	Web3NameAccount(LinkableAccountId),
+	/// An account linked to the DID and resolvable by or to a Dotname.
+	DotNameAccount(LinkableAccountId),
```
rflechtner, line 33: that's because we have the two instances of the linking pallet?

rflechtner (approved):
Seems reasonable! The enums can serve as the missing error cases to show to users via a UI to help users identify which clean up steps they need to perform. And for the transactions I would assume the client just needs logic to batch them in a DID-origin transaction, which the sdk already has. Did you try this out?

runtimes/common/src/did/runtime_apis.rs
ntn-x2: Yes, exactly.

PR #856, opened 2025-01-28
fix: add required migration

ntn-x2:
I added the required migration for the new deployments, and made the `BumpStorageVersion` type a bit more generic to work with any pallets that might need the same treatment in the future.

I tried the new migrations with both Peregrine (new runtime) and Spiritnet (previous runtime), and I got no errors.

### DIFF
--- a/runtimes/common/src/migrations.rs
+++ b/runtimes/common/src/migrations.rs
@@ -17,12 +17,12 @@
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
 use frame_support::{
-	pallet_prelude::StorageVersion,
-	traits::{GetStorageVersion, OnRuntimeUpgrade},
+	traits::{GetStorageVersion, OnRuntimeUpgrade, PalletInfoAccess, StorageVersion},
 	weights::Weight,
 };
 use sp_core::Get;
-use sp_std::marker::PhantomData;
+use sp_std::{fmt::Debug, marker::PhantomData};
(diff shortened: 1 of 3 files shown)


runtimes/common/src/migrations.rs
```diff
@@ -17,55 +17,68 @@
...
 
 	#[cfg(feature = "try-runtime")]
 	fn post_upgrade(_state: sp_std::vec::Vec<u8>) -> Result<(), sp_runtime::TryRuntimeError> {
-		if pallet_assets::Pallet::<T>::on_chain_storage_version() < TARGET_PALLET_ASSETS_STORAGE_VERSION {
+		let (on_chain_version, current_version) = (T::on_chain_storage_version(), T::current_storage_version());
+
+		if on_chain_version < current_version {
 			Err(sp_runtime::TryRuntimeError::Other(
-				"pallet_assets storage version was not updated to v1.",
+				"Pallet storage version was not updated to the latest version.",
```
Ad96el, line 81: Maybe also add the name here

Ad96el (approved):

runtimes/common/src/migrations.rs
ntn-x2: Good point: https://github.com/KILTprotocol/kilt-node/pull/856/commits/d46bc49de09b333638ca2d05e90d2fbb7a12c6a5

PR #857, opened 2025-02-03
chore: chopsticks utils

Ad96el:
Adds some basic chopsticks functionality + a CLI.

### DIFF
--- a/integration-tests/chopsticks/src/cli.ts
+++ b/integration-tests/chopsticks/src/cli.ts
@@ -18,8 +18,18 @@ program
 	.description('Executes a transaction on the network')
 	.argument('<endpoint>', 'The endpoint of the network')
 	.argument('<rawTx>', 'The raw transaction to execute')
-	.option('--origin <String>', 'The origin of the transaction', 'Root')
-	.option('--port <Number>', 'The RPC port', '8888')
+	.option(
+		'--origin',
+		'The origin of the transaction Either "Root" or "Signed". Default is "Root"',
+		(value) => {
+			if (value !== 'Root' && value !== 'Signed') {
+				throw new Error('Invalid origin. Must be either "Root" or "Signed"')
+			}
+			return value
+		},
+		'Root'
+	)
+	.option('--port', 'The RPC port', '8888')
 	.action(async (endpoint, rawTx, options) => {
 		const { origin, port } = options
 
--- a/integration-tests/chopsticks/src/cli.ts
+++ b/integration-tests/chopsticks/src/cli.ts
@@ -1,6 +1,6 @@
 import { Command } from 'commander'
 
-import { createTestNetwork, scheduleTxCommand } from './command/index.js'
+import { createTestNetwork, scheduleTxCommand, stateTransition } from './command/index.js'
 
 const program = new Command()
 
@@ -22,8 +22,16 @@ program
 	.option('--port <Number>', 'The RPC port', '8888')
 	.action(async (endpoint, rawTx, options) => {
 		const { origin, port } = options
-		console.log(endpoint, rawTx, origin, port)
+
 		await scheduleTxCommand(endpoint, rawTx, origin, +port)
+	})
+
+program
+	.command('stateTransition')
+	.description('Shows the state transition of the network by the latest block')
+	.argument('<endpoint>', 'The endpoint of the network')
+	.action(async (endpoint) => {
+		await stateTransition(endpoint)
 	})
 
 program.parse()
--- a/integration-tests/chopsticks/src/cli.ts
+++ b/integration-tests/chopsticks/src/cli.ts
@@ -30,8 +30,11 @@ program
 	.command('stateTransition')
 	.description('Shows the state transition of the network by the latest block')
 	.argument('<endpoint>', 'The endpoint of the network')
-	.action(async (endpoint) => {
-		await stateTransition(endpoint)
+	.option('--block', 'The block number to do the state transition', 'undefined')
+	.action(async (endpoint, options) => {
+		const { block } = options
+		const blockNumber = block === 'undefined' ? undefined : +block
+		await stateTransition(endpoint, blockNumber)
 	})
 
 program.parse()
--- a/integration-tests/chopsticks/src/cli.ts
+++ b/integration-tests/chopsticks/src/cli.ts
@@ -1,0 +1,29 @@
+import { Command } from 'commander'
+
+import { createTestNetwork, scheduleTxCommand } from './command/index.js'
+
+const program = new Command()
+
+program.name('chopsticks utils').description('CLI to mange some chopsticks instances').version('0.0.1')
+
+program
+	.command('spinUp')
+	.description('Spin up the network')
+	.action(async () => {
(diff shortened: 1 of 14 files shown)


integration-tests/chopsticks/src/command/network.ts
```diff
@@ -0,0 +1,28 @@
...
+	// 2. Create the network specific config
+	const kiltConfig = mainChains.kilt.getConfig({})
+
+	// 2. Add the different parachainContexts here. The order of parachainContexts and parachainStorage should match.
+	const parachainOptions = [kiltConfig]
+	const parachainStorage = [kiltStorage]
+
+	const { relayChainContext, parachainContexts } = await setupNetwork(relayConfig, parachainOptions)
+
+	parachainStorage.map(async (storage, index) => await setStorage(parachainContexts[index], storage))
```
rflechtner, line 21:
```suggestion
	await Promise.all(parachainStorage.map((storage, index) => setStorage(parachainContexts[index], storage)))
```

nothing is awaited here without an `await Promise.all()` wrapping this statement (right now it just returns an array of promises which are not observed). The await inside of the mapping function is pointless too (your function returns a Promise regardless of whether you await inside it or not).

integration-tests/chopsticks/src/command/network.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	await Promise.all(
+		[...parachainContexts, relayChainContext].map((context) => {
+			context.pause()
+		})
```
rflechtner, line 26:
```suggestion
		[...parachainContexts, relayChainContext].map((context) => context.pause())
```
Nothing to await if there's nothing returned; either remove the curly brackets or add an explicit return before `context.pause()`

integration-tests/chopsticks/src/command/stateTransition.ts
```diff
@@ -0,0 +1,16 @@
...
+	if (blockNumber) {
+		options.push(`--block=${blockNumber}`)
+	}
+
+	try {
+		const { stdout } = await execa('yarn', options)
+		console.log(stdout)
+	} catch (error) {
+		console.error('Error:', error)
+	}
```
rflechtner, line 15: There's two ways of writing this; which one you use is a matter of preference.

```suggestion
	await execa('yarn', options)
	  .then(({stdout})=>{console.log(stdout)})
	  .catch(error => {console.error('Error:', error)})
```

integration-tests/chopsticks/src/network/index.ts
```diff
@@ -10,7 +10,10 @@ import { ChainConfigs } from './types.js'
 function getRequiredEnvVariable(name: string): string {
 	const value = process.env[name]
 	if (value === undefined) {
-		throw new Error(`Environment variable ${name} is not set.`)
+		if (process.env.NODE_ENV === 'test') {
+			console.error(`Error: Environment variable ${name} is not set. Some tests might not work.`)
+		}
+		return ''
```
rflechtner, line 16: what's the reason you are no longer throwing here?

integration-tests/chopsticks/src/cli.ts
```diff
@@ -0,0 +1,40 @@
...
+	.action(async () => {
+		await createTestNetwork()
+	})
+
+program
+	.command('scheduleTx')
+	.description('Executes a transaction on the network')
+	.argument('<endpoint>', 'The endpoint of the network')
+	.argument('<rawTx>', 'The raw transaction to execute')
+	.option('--origin <String>', 'The origin of the transaction', 'Root')
```
rflechtner, line 21: it's unclear what the options are; maybe turn this into a choice option?

integration-tests/chopsticks/src/network/utils.ts
```diff
@@ -72,7 +71,7 @@ export async function setStorage(context: Config, storage: Record<string, Record
 }
 
 /// Schedules a transaction with root privileges at the given block number. If no block is provided, the transaction will be scheduled for the next block.
-export async function scheduleTx({ api }: { api: ApiPromise }, call: SubmittableExtrinsic<'promise'>, at = undefined) {
+export async function scheduleTx({ api }: { api: ApiPromise }, call: string, at = undefined, origin = 'Root') {
```
rflechtner, line 74: Also here providing the origin choices as the typescript type of this parameter would help users understand which origins can be used.

rflechtner (changes requested):
There's two classic gotchas with regards to array mapping and Promises, but LGTM otherwise!

integration-tests/chopsticks/src/command/stateTransition.ts
rflechtner: Yeah then()/catch() are methods on promises and allow you to handle their results.

integration-tests/chopsticks/src/network/index.ts
Ad96el: Yeah, I found it a bit inconvenient to throw an error if the variables are not set. But I think it's better to throw an error and not run into an undefined state: [4807d9e](https://github.com/KILTprotocol/kilt-node/pull/857/commits/4807d9ee08a46a18f5ace5fb49cf1996b0ac3df4)

integration-tests/chopsticks/src/cli.ts
Ad96el: It is not super easy to do an option choice, so I did a callback for the option: [981999f](https://github.com/KILTprotocol/kilt-node/pull/857/commits/981999fb72240deb97ab1c440b7fef85a05a6912) 

But to be honest, I would like to keep it generic. They might be different origins, such as `TechnicalCommitee` or `TipperOrigin` which we would deny now. Other chains can also have different origin names.

integration-tests/chopsticks/src/network/utils.ts
Ad96el: [293e6d3](https://github.com/KILTprotocol/kilt-node/pull/857/commits/293e6d33a5be1f1c90b4963c0f25051936f55b4f)

integration-tests/chopsticks/src/network/index.ts
```diff
@@ -10,7 +10,10 @@ import { ChainConfigs } from './types.js'
 function getRequiredEnvVariable(name: string): string {
 	const value = process.env[name]
 	if (value === undefined) {
-		throw new Error(`Environment variable ${name} is not set.`)
+		if (process.env.NODE_ENV === 'test') {
+			throw new Error(`Error: Environment variable ${name} is not set. Some tests might not work.`)
+		}
+		return ''
```
rflechtner, line 16: so in non-test mode this now returns an empty string for an env that is not set? A bit counterintuitive given that the function name is `getRequiredEnvVariable`

rflechtner (approved):
LGTM

integration-tests/chopsticks/src/network/index.ts
Ad96el: [17323b1](https://github.com/KILTprotocol/kilt-node/pull/857/commits/17323b1de7b3ecbaec010e830b2dd6c62228dd27)

PR #865, opened 2025-03-03
fix: make sure the reset_team trait actually sets the owner

rflechtner:
### Fixes https://github.com/KILTprotocol/ticket/issues/3802

Changes the reset_team trait to actually set the owner passed to the function.

The trait implementation is a back-port of a trait implemented in latest versions of pallet-assets (https://docs.rs/pallet-assets/41.0.0/pallet_assets/pallet/struct.Pallet.html#impl-ResetTeam%3C%3CT+as+Config%3E::AccountId%3E-for-Pallet%3CT,+I%3E). Changing ownership of an asset is part of this trait but was not implemented here as not needed.
Nevertheless, we're adding this now to avoid mismatched expectations.

### TBD

- [x] Update benchmarking to cover the case where the current owner != the desired owner.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/benchmarking.rs
+++ b/pallets/pallet-bonded-coins/src/benchmarking.rs
@@ -368,6 +368,10 @@ mod benchmarks {
 		let curve = get_linear_bonding_curve::<CurveParameterTypeOf<T>>();
 		let pool_id = create_pool::<T>(curve, bonded_currencies.clone(), Some(account_origin), None, None);
 
+		// Although these would rarely happen in practice, for benchmarking we assume
+		// the worst case where the owner must be changed as well
+		assert!(T::Fungibles::owner(bonded_currencies[0].clone()) != Some(pool_id.clone().into()));
+
 		let admin: AccountIdOf<T> = account("admin", 0, 0);
 		let freezer: AccountIdOf<T> = account("freezer", 0, 0);
 		let fungibles_team = PoolManagingTeam {
@@ -377,12 +381,19 @@ mod benchmarks {
 
 		let max_currencies = T::MaxCurrenciesPerPool::get();
 		#[extrinsic_call]
-		_(origin as T::RuntimeOrigin, pool_id, fungibles_team, max_currencies);
+		_(
+			origin as T::RuntimeOrigin,
+			pool_id.clone(),
+			fungibles_team,
+			max_currencies,
+		);
 
 		// Verify
 		bonded_currencies.iter().for_each(|asset_id| {
 			assert_eq!(T::Fungibles::admin(asset_id.clone()), Some(admin.clone()));
 			assert_eq!(T::Fungibles::freezer(asset_id.clone()), Some(freezer.clone()));
+			assert_eq!(T::Fungibles::owner(asset_id.clone()), Some(pool_id.clone().into()));
+			assert_eq!(T::Fungibles::issuer(asset_id.clone()), Some(pool_id.clone().into()));
 		});
 	}
 
--- a/pallets/pallet-bonded-coins/src/benchmarking.rs
+++ b/pallets/pallet-bonded-coins/src/benchmarking.rs
@@ -379,11 +379,13 @@ mod benchmarks {
 			freezer: freezer.clone(),
 		};
 
+		let pool_id_for_call = pool_id.clone();
+
 		let max_currencies = T::MaxCurrenciesPerPool::get();
 		#[extrinsic_call]
 		_(
 			origin as T::RuntimeOrigin,
-			pool_id.clone(),
+			pool_id_for_call,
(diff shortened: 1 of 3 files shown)


rflechtner:
@ntn-x2 pinging you to have a look at this issue. What do you think about the implementation of ownership change?

pallets/pallet-bonded-coins/src/benchmarking.rs
```diff
@@ -377,12 +381,19 @@ mod benchmarks {
 
 		let max_currencies = T::MaxCurrenciesPerPool::get();
 		#[extrinsic_call]
-		_(origin as T::RuntimeOrigin, pool_id, fungibles_team, max_currencies);
+		_(
+			origin as T::RuntimeOrigin,
+			pool_id.clone(),
```
ntn-x2, line 386: Do not `clone()` as part of the benchmark itself, but do it before.

pallets/pallet-bonded-coins/src/traits.rs
```diff
@@ -80,21 +82,37 @@ pub trait ResetTeam<AccountId>: Inspect<AccountId> {
...
 		admin: AccountIdOf<T>,
 		issuer: AccountIdOf<T>,
 		freezer: AccountIdOf<T>,
 	) -> DispatchResult {
-		let owner = AssetsPallet::<T, I>::owner(id.clone()).ok_or(DispatchError::Unavailable)?;
-		let origin = RawOrigin::Signed(owner);
-		AssetsPallet::<T, I>::set_team(origin.into(), id.into(), issuer.into(), admin.into(), freezer.into())
+		let current_owner = AssetsPallet::<T, I>::owner(id.clone()).ok_or(DispatchError::Unavailable)?;
+		if current_owner != owner {
+			AssetsPallet::<T, I>::transfer_ownership(
```
ntn-x2, line 103: Isn't there an internal function that would eliminate the need to create a `SignedOrigin` parameter?

pallets/pallet-bonded-coins/src/traits.rs
```diff
@@ -80,21 +82,37 @@ pub trait ResetTeam<AccountId>: Inspect<AccountId> {
...
-		_owner: AccountIdOf<T>,
+		owner: AccountIdOf<T>,
 		admin: AccountIdOf<T>,
 		issuer: AccountIdOf<T>,
 		freezer: AccountIdOf<T>,
 	) -> DispatchResult {
-		let owner = AssetsPallet::<T, I>::owner(id.clone()).ok_or(DispatchError::Unavailable)?;
-		let origin = RawOrigin::Signed(owner);
-		AssetsPallet::<T, I>::set_team(origin.into(), id.into(), issuer.into(), admin.into(), freezer.into())
+		let current_owner = AssetsPallet::<T, I>::owner(id.clone()).ok_or(DispatchError::Unavailable)?;
```
ntn-x2, line 101: Do we need this check? Would not this be checked from within the `transfer_ownership` function? If yes, this is redundant. If not, why do we need it? Just so that we can run the `if`?

pallets/pallet-bonded-coins/src/traits.rs
rflechtner: No, there's just the transfer_ownership and reset_team calls that we can use in this version. That's why we need this trait implementation.

pallets/pallet-bonded-coins/src/traits.rs
rflechtner: We need the current owner to construct a SignedOrigin which is authorised to perform ownership changes. We can omit the check if the current owner is the new owner, but this would increase weights somewhat. We'd be reading the owner again, and performing the same check in the transfer_ownership function. But, yeah, not strictly required - should I get rid of the check?

ntn-x2: Repeated storage reads are cached anyway, so they don't incur in any additional weight. If the check is not strictly forbidden, I would remove it, also because if for any reason the underlying logic changes, this duplicated logic risks to go out of sync. So, yeah, if not strictly required I would remove it. If weight is a concern, it is not since reads are and only committed once.

pallets/pallet-bonded-coins/src/benchmarking.rs
rflechtner: e1670e61c1a127962c55817f2a536f8cf78fa076

pallets/pallet-bonded-coins/src/traits.rs
rflechtner: Ok, done in dc4f8f2cf7a1d86c39f7f075e7ed54a15b4d30c9

ntn-x2 (approved):

PR #868, opened 2025-03-05
fix: increase MaxConsumers to allow accounts to hold more than 15 different bonded coins

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3798

Increases consumers limits to 100 for peregrine, spiritnet, mocks & tests.
Configs for dip-templates are left as-is.

It's hard for me to tell if this is going to have an impact on benchmarking / transaction weights though, how can we check this?

## Checklist:

    * Link the PR/Ticket here


ntn-x2 (approved):
I don't think it has any impact on benchmarks and weights.

PR #869, opened 2025-03-05
refactor: batch updating of asset team

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3801

Changes the `reset_team` extrinsic to perform batch updates in order to avoid that asset management teams differ between assets of one pool.

Recommended by auditors and desired by our team - see linked ticket.

### TBD

- [x] Update benchmarks to account for different numbers of currencies on a pool.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -449,18 +449,17 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Changes the managing team of a bonded currency which is issued by
-		/// this pool. The new team will be set to the provided team. The
-		/// currency index is used to select the currency that the team will
-		/// manage. The origin account must be a manager of the pool.
+		/// Changes the managing team of all bonded currencies issued by this
+		/// pool, setting it to the provided `team`. The origin account must be
+		/// a manager of the pool.
 		///
 		/// # Parameters
 		/// - `origin`: The origin of the call, requiring the caller to be a
 		///   manager of the pool.
 		/// - `pool_id`: The identifier of the pool.
 		/// - `team`: The new managing team.
-		/// - `currency_idx`: The index of the currency in the bonded currencies
-		///   vector.
+		/// - `currency_count`: The number of bonded currencies vector linked to
+		///   the pool. Required for weight estimations.
 		///
 		/// # Returns
 		/// - `DispatchResult`: The result of the dispatch.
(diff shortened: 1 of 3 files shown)


rflechtner:
> Makes sense to enforce this within the pallet. This must be mentioned somewhere in the docs tho.

I updated the extrinsic's description, is there any other place where the change should be documented? I don't think we have actual docs for this pallet rn.

ntn-x2:
No I think that's all there is for now.

rflechtner:
@ntn-x2 I updated the benchmarks in 701410cb54ee4d17d1921caf914cb8ac91787960, based on the implementation of `start_destroy`'s benchmarks. Let me know if there are any issues here.

pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -469,40 +468,40 @@ pub mod pallet {
 		/// - `Error::<T>::PoolUnknown`: If the pool does not exist.
 		/// - `Error::<T>::NoPermission`: If the caller is not a manager of the
 		///   pool.
-		/// - `Error::<T>::IndexOutOfBounds`: If the currency index is out of
-		///   bounds.
+		/// - `Error::<T>::CurrencyCount`: If the actual number of currencies in
+		///   the pool is larger than `currency_count`.
 		/// - Other errors depending on the types in the config.
 		#[pallet::call_index(1)]
 		#[pallet::weight(T::WeightInfo::reset_team())]
```
rflechtner, line 475:
```suggestion
		#[pallet::weight(T::WeightInfo::reset_team(currency_count.to_owned()))]
```
I'm realising this needs to be changed, but in order for this to work it seems that I'd have to re-run benchmarks? I'll definitely need guidance for that, I have never done this before.

ntn-x2: But in general, what you would need to do, is to pre-populate the storage with as many pools as the theoretical maximum, and then use the range of possible inputs to pass to the extrinsic. You can see an example of it for the `create_pool_polynomial` benchmark here: https://github.com/KILTprotocol/kilt-node/blob/67c3d65a8adb6b0ac9578492d882c1f9f117fcdd/pallets/pallet-bonded-coins/src/benchmarking.rs#L247. So I think it's sufficient to pre-populate the storage with N assets where N is the benchmark input, and then simply call `reset_team(N)`. That should be enough.

rflechtner: I'm not sure I understand, https://github.com/KILTprotocol/kilt-node/commit/701410cb54ee4d17d1921caf914cb8ac91787960 already makes changes in line with the snippet you linked to, but I still can't make the change I suggested in https://github.com/KILTprotocol/kilt-node/pull/869#discussion_r1983083354 because `default_weights.rs` has not been updated (the compiler complains that `this function takes 0 arguments but 1 argument was supplied`). Should I just manually mess around in the `default_weights.rs` for now until it compiles, even though it's an autogenerated file?

ntn-x2: Just leave the previous weight for now. Once the new benchmarks are run, the pallet will complain that it now requires one parameter, and then you can fill in the result of the benchmarks. Sorry I did not see the commit you previously linked. But yeah, just leave the old weight for the extrinsic. As soon as the new benchmarks are run, the code won't compile and you will update that. As a temporary solution, you COULD also update the weight file, since it will be overwritten anyway upon new benchmarks, but I would not recommend that.

PR #870, opened 2025-03-06
feat: disallow re-using token names and symbols within a pool

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3803

Disallows re-using the same token symbol or name. Empty names and symbols are always allowed.

## Checklist:

    * Link the PR/Ticket here


ntn-x2 (approved):

rflechtner:
Failing tests suggest that `HashSet` is not available to be used in the runtime, so I replaced it with `BTreeSet`

PR #871, opened 2025-03-06
fix: do not allow lock states that are equivalent to unlocked

rflechtner:
## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/pallets/pallet-bonded-coins/src/lib.rs
+++ b/pallets/pallet-bonded-coins/src/lib.rs
@@ -557,7 +557,8 @@ pub mod pallet {
 		/// - `origin`: The origin of the call, requiring the caller to be a
 		///   manager of the pool.
 		/// - `pool_id`: The identifier of the pool to be locked.
-		/// - `lock`: The locks to be applied to the pool.
+		/// - `lock`: The locks to be applied to the pool. At least one lock
+		///   flag must be set to `false` for a valid lock.
 		///
 		/// # Returns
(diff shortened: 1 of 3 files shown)


pallets/pallet-bonded-coins/src/lib.rs
```diff
@@ -568,11 +569,22 @@ pub mod pallet {
...
+		/// - `Error::<T>::InvalidInput`: If all lock flags are `true` (all
+		///   operations enabled), which would be equivalent to an unlocked
+		///   (active) pool state.
 		#[pallet::call_index(3)]
 		#[pallet::weight(T::WeightInfo::set_lock())]
 		pub fn set_lock(origin: OriginFor<T>, pool_id: T::PoolId, lock: Locks) -> DispatchResult {
 			let who = T::DefaultOrigin::ensure_origin(origin)?;
 
+			ensure!(
+				lock != (Locks {
```
ntn-x2, line 581: Can we do the same without allocating a new `Locks` object?

ntn-x2 (approved):

pallets/pallet-bonded-coins/src/lib.rs
rflechtner: Good point, that's bad. I wanted to avoid the scenario where we add a field to the locks and then forget adding a check for this new field here, but I think fa4bde344f88c9f4274eab98c8aa743c03cbd0c2 does avoid that pretty well too

PR #877, opened 2025-03-27
fix: remove messageQueue

Ad96el:
## fixes Chopsticks 

We decided to remove the `messageQueue` pallet from the snapshot tests since its events are not reliably reproducible and cause flaky tests. This PR removes all occurrences.


rflechtner (approved):
LGTM!

PR #878, opened 2025-03-27
chore: update deny rules

Ad96el:


.cargo-deny.toml
```diff
@@ -18,6 +18,9 @@ ignore = [
   { id = "RUSTSEC-2024-0384", reason = "`instant` is unmaintained but a Substrate dependency. Re-verify upon next polkadot-sdk updates." },
   { id = "RUSTSEC-2024-0388", reason = "`derivative` is unmaintained but a Substrate dependency. Re-verify upon next polkadot-sdk updates." },
   { id = "RUSTSEC-2024-0421", reason = "`idna` has a security vulnerability but a Substrate dependency. Re-verify upon next polkadot-sdk updates." },
+  { id = "RUSTSEC-2024-0436", reason = "`paste` is unmaintained but a Substrate dependency. Re-verify upon next polkadot-sdk updates." },
+  { id = "RUSTSEC-2025-0009", reason = "`ring` has a security vulnerability but a Substrate dependency. Re-verify upon next polkadot-sdk updates." },
```
rflechtner, line 22: while there is a 0.16 version of ring in the tree that cannot be updated, there is also a 0.17 version that can, did you do that?

rflechtner (approved):

.cargo-deny.toml
Ad96el: Version 17 has no security issue. What do you exactly mean?

PR #882, opened 2025-04-22
feat: Permissioned collators

Ad96el:
## fixes [#3822](https://github.com/KILTprotocol/ticket/issues/3822)

This PR introduces permissioned collators to the Peregrine and Spirtinet runtimes. The council or governance is now allowed to add and remove members through the newly deployed pallet-membership. This pallet is then used by a new session manager implementation. The old session manager implementation remains untouched—it's only moved out of the runtime.

After the migration, the parachain-staking pallet becomes essentially obsolete, as collators no longer receive rewards for block production.

A follow-up PR will reduce rewards in the parachain-staking pallet accordingly.

The round information (which determines session length) still resides in the parachain-staking pallet. Before it can be removed, a new source for round length must be implemented—either directly in the runtime or via a custom pallet that includes collator information.

### How to test
- Navigate to the zombienet directory and spin up a Peregrine runtime.
- Perform the runtime upgrade (sudo -> system -> set_code).
- Add a new collator via sudo or start a council motion (sudo -> permissionedCollators -> add_member).
- Force a new round (sudo -> parachainStaking -> forceNewRound).
- Verify in the session pallet that the new collator is queued. (To use a new collator, spin up a node with the correct keys. Using an existing collator is usually easier.)
- Force another new round.
- Only the new collator should be producing blocks now.


Ad96el:
I decided to add the benchmarking in this PR.

runtimes/peregrine/src/system/mod.rs
```diff
@@ -280,3 +318,21 @@ impl pallet_scheduler::Config for Runtime {
...
 }
+
+type CollatorMembershipProvider = pallet_membership::Instance3;
+impl pallet_membership::Config<CollatorMembershipProvider> for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type AddOrigin = RootOrMoreThanHalfCouncil;
+	type RemoveOrigin = RootOrMoreThanHalfCouncil;
+	type SwapOrigin = EnsureRoot<AccountId>;
+	type ResetOrigin = EnsureRoot<AccountId>;
+	type PrimeOrigin = EnsureRoot<AccountId>;
```
rflechtner, line 329: We don't have a use for a prime member, so ideally we disallow setting one. Can we set a 'never origin' here, if this exists?

runtimes/spiritnet/src/system/mod.rs
```diff
@@ -141,13 +142,50 @@ impl_opaque_keys! {
...
+
+	fn start_session(_start_index: sp_staking::SessionIndex) {
+		// We don't care
+	}
+
+	fn end_session(_end_index: sp_staking::SessionIndex) {
+		// We don't care
+	}
+}
+
```
rflechtner, line 181: is the duplication necessary? Can't we define this in runtime-support?

Ad96el: Yes, this was bothering me too. The issue was that including pallet-session in the runtime common crate caused circular dependencies. It wasn't entirely clear to me why this happened only when the pallet-session runtime benchmark feature was enabled. Disabling the feature allowed it to compile for all targets: [8a71d6e](https://github.com/KILTprotocol/kilt-node/pull/882/commits/8a71d6ed28d6cf9516fdd1ce191f85811404e5c0) 

was not easy to figure out :smiling_face_with_tear:

runtimes/peregrine/src/lib.rs
```diff
@@ -116,6 +116,7 @@ construct_runtime! {
 		ParachainStaking: parachain_staking = 21,
 		Authorship: pallet_authorship = 20,
 		AuraExt: cumulus_pallet_aura_ext = 24,
+		PermissionedCollator: pallet_membership::<Instance3> = 25,
```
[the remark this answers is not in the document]
Ad96el: [6407d31](https://github.com/KILTprotocol/kilt-node/pull/882/commits/6407d31edcc964c88b63f4fc71623f986f0fa19e)

runtimes/peregrine/src/system/mod.rs
Ad96el: I also updated some other config implementation: [cc706fa](https://github.com/KILTprotocol/kilt-node/pull/882/commits/cc706fa9a0be759f9334eec88430dff89741a3d6)

runtimes/peregrine/src/system/mod.rs
```diff
@@ -141,13 +142,50 @@ impl_opaque_keys! {
...
+		// We don't care
+	}
+}
+
 impl pallet_session::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type ValidatorId = AccountId;
 	type ValidatorIdOf = ConvertInto;
 	type ShouldEndSession = ParachainStaking;
 	type NextSessionRotation = ParachainStaking;
```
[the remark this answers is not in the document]
Ad96el: Introduced in: [12a5b7f](https://github.com/KILTprotocol/kilt-node/pull/882/commits/12a5b7f7a387e4a445abd8604a011f0541242850)

The session length is 2 hours. I guess the length does not hurt. I can also increase it to one day.

runtimes/peregrine/src/system/mod.rs
```diff
@@ -141,13 +142,50 @@ impl_opaque_keys! {
...
+		System::register_extra_weight_unchecked(
+			<Runtime as frame_system::Config>::DbWeight::get().reads(2),
+			frame_support::pallet_prelude::DispatchClass::Mandatory,
+		);
+
+		if collators.is_empty() {
+			// we never want to pass an empty set of collators. This would brick the chain.
+			log::error!("💥 keeping old session because of empty collator set!");
+			return None;
+		}
```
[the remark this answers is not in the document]
Ad96el: [2dcd021](https://github.com/KILTprotocol/kilt-node/pull/882/commits/2dcd0210157c976cbe8c9542eb04266297f37188)

runtimes/peregrine/src/system/mod.rs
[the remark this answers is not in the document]
rflechtner: 2 hours sounds good to me

abdulmth (approved):

runtimes/common/src/session.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::constants::staking::DefaultBlocksPerRound;
+
+type AccountIdOf<Runtime> = <Runtime as frame_system::Config>::AccountId;
+
+/// The session manager for the collator set.
+pub struct SessionManager<Runtime>(PhantomData<Runtime>);
+
+impl<Runtime> SessionManagerTrait<AccountIdOf<Runtime>> for SessionManager<Runtime>
+where
+	Runtime: MembershipConfig<Instance3> + pallet_session::Config,
```
ntn-x2, line 38: The instance itself should also be generic, since you can't assume there are no additional instances deployed on the actual runtime.

runtimes/common/src/session.rs
```diff
@@ -0,0 +1,81 @@
...
+
+		let has_collator_keys = collators.iter().any(|collator| {
+			pallet_session::NextKeys::<Runtime>::contains_key(<Runtime as pallet_session::Config>::ValidatorId::from(
+				collator.clone(),
+			))
+		});
+
+		SystemPallet::<Runtime>::register_extra_weight_unchecked(
+			<Runtime as frame_system::Config>::DbWeight::get()
+				.reads(2u64.saturating_add(collators.len().saturated_into::<u64>())),
```
ntn-x2, line 59: What's the second read, besides `let collators = MembershipPallet::<Runtime, Instance3>::members().to_vec();`?

runtimes/common/src/session.rs
```diff
@@ -0,0 +1,81 @@
...
+			frame_support::pallet_prelude::DispatchClass::Mandatory,
+		);
+
+		if collators.is_empty() || !has_collator_keys {
+			// we never want to pass an empty set of collators. This would brick the chain.
+			log::error!("💥 keeping old session because of empty collator set!");
```
ntn-x2, line 65: I think a different error message depending on which condition is `false` would massively reduce headaches in the future.

runtimes/common/src/constants.rs
```diff
@@ -309,6 +311,8 @@ pub mod staking {
 		pub const NetworkRewardStart: BlockNumber = super::treasury::INITIAL_PERIOD_LENGTH;
 		/// The rate in percent for the network rewards
 		pub const NetworkRewardRate: Perquintill = NETWORK_REWARD_RATE;
+		/// The maximum number of collators
+		pub const MaxCollators: u32 = MAX_COLLATORS;
```
ntn-x2, line 315: Is this a leftover? I couldn't see where it's used.

ntn-x2 (commented):
Minor things. Also, it seems the PR description is outdated, since I can't see where the staking pallet is being used? Session length has a fixed duration now, so the only thing that the staking pallet does, is to provide collators until there is a well formed collator set (meaning non empty and with all session keys set) that is returned by the new session manager.

runtimes/common/src/session.rs
Ad96el: We read the collators, and we check for each collator, if they have some keys in the session pallet. Makes in total 1 + collators.len() of reads. I added one more for overestimating the weights for being an upper bound. Do you think this is not necessary?

runtimes/common/src/constants.rs
Ad96el: The constant is used in the pallet membership pallet: https://github.com/KILTprotocol/kilt-node/pull/882/files#diff-ccc6b55434b4eb75cfff1f592e229e32a3c0927c8aa151af5bd557aa2b0c42e4R299

runtimes/common/src/session.rs
Ad96el: Can I not assume, that only a single instance3 is in the runtime present? Collisions are not possible, are they? 

Should I change the implementation to:

```
impl<Runtime, Instance> SessionManagerTrait<AccountIdOf<Runtime>> for SessionManager<Runtime>
where
	Runtime: MembershipConfig<Instance> + pallet_session::Config,
```

?

runtimes/common/src/session.rs
Ad96el: Fair point. I created a new PR: https://github.com/KILTprotocol/kilt-node/pull/887

runtimes/common/src/session.rs
ntn-x2: I don't think it's necessary, but even if it is, please add a comment as to why the weight does not match what's actually supposed to be.

runtimes/common/src/session.rs
ntn-x2: Yes, perhaps with a `Instance: 'static` bound if the compiler complains, but that should be the proper way of doing it.

runtimes/common/src/constants.rs
ntn-x2: Ah ok, it was confusing since it's in the `staking` module.

runtimes/common/src/session.rs
Ad96el: I changed it to the right calculation: [535b36b](https://github.com/KILTprotocol/kilt-node/pull/887/commits/535b36be4de61dafd1d8ccad175f17ff567ef6b4)

runtimes/common/src/session.rs
Ad96el: Added in the instance in the generic: [234920a](https://github.com/KILTprotocol/kilt-node/pull/887/commits/234920a857e334a0ebb33fa52c60964ab6b76c12)

PR #883, opened 2025-04-23
chore(deps): bump tokio due to RUSTSEC-2025-0023

rflechtner:
## fixes RUSTSEC-2025-0023

Cargo deny raises warnings about an issue in the Tokyo library, which has soundness issues in the version used here.
Fortunately this can easily be bumped to a patched version.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4608,7 +4608,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite 0.2.14",
- "socket2 0.5.7",
+ "socket2 0.4.10",
  "tokio",
  "tower-service",
  "tracing",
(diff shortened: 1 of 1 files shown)


Ad96el (dismissed):

Cargo.lock
```diff
@@ -4608,7 +4608,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite 0.2.14",
- "socket2 0.5.7",
+ "socket2 0.4.10",
```
rflechtner, line 4611: this is strange - not sure why a package in an unrelated packaged would have downgraded suddenly. Stranger still, if I run `cargo update -p socket2@0.4.10` this change is reverted. @Ad96el any idea what's going on?

Ad96el: The dependency comes from hyper, which pins the version to >=0.4.7, <0.6.0. It seems Cargo isn't fully deterministic in this case - possibly selecting a version randomly. This might happen if another crate was previously compiled with a stricter version constraint.

https://github.com/hyperium/hyper/blob/v0.14.29/Cargo.toml#L42

rflechtner: ok, good to know. To keep the change set minimal, I reverted this change in 2a263a882e6e51e6ea0d2f26bc9fa8ad4f8f37df

Ad96el (approved):

PR #901, opened 2025-07-03
fix: reset_team weights

rflechtner:
Fixes an oversight in weights computation, where the number of currencies in a pool was not taken into account for the bonded-coins pallet's `reset_team` extrinsic.
Also updates the default weights accordingly.

Given that the `run_benches_for_pallets.sh` script no longer works, I've benchmarked this manually using the frame-omni-bencher. Double-check that I've done it right, given that it's my first time doing so.

## Checklist:

    * Link the PR/Ticket here


Ad96el (approved):
LGTM. I'm curious why this wasn't catchedduring the release. The node should not compile after running the benchmarks, no?

rflechtner:
> LGTM. I'm curious why this wasn't catched during the release. The node should not compile after running the benchmarks, no? 

Yes indeed, but the benchmarks weren't actually run, that's why this wasn't caught.

PR #902, opened 2025-07-03
fix: allow fast unstaking for all

rflechtner:
Allows unstaking for all collator candidates and delegators in the staking pallet with minimum delay. 
Also prevents additional candidates or delegators from joining, and existing ones from increasing their stake.

__Note that collator selection is no longer dependent on the staking pallet, so removal of all candidates is inconsequential for the collator set.__

## Checklist:

    * Link the PR/Ticket here


runtimes/common/src/constants.rs
```diff
@@ -256,9 +256,9 @@ pub mod staking {
 	pub const DEFAULT_BLOCKS_PER_ROUND: BlockNumber = 2 * HOURS;
 
 	#[cfg(feature = "fast-gov")]
-	pub const STAKE_DURATION: BlockNumber = 30;
+	pub const STAKE_DURATION: BlockNumber = 1;
 	#[cfg(not(feature = "fast-gov"))]
-	pub const STAKE_DURATION: BlockNumber = 7 * DAYS;
+	pub const STAKE_DURATION: BlockNumber = 1;
```
abdulmth: There's no need to specify features here because the value is the same in both cases.

runtimes/common/src/constants.rs
```diff
@@ -256,9 +256,9 @@ pub mod staking {
...
 
 	#[cfg(feature = "fast-gov")]
 	pub const MIN_COLLATORS: u32 = 4;
```
abdulmth, line 261: Don't we want to change this too?

Ad96el: I don't think this constant is very important. It's only used in set_max_selected_candidates as a [validation parameter](https://github.com/KILTprotocol/kilt-node/blob/develop/pallets/parachain-staking/src/lib.rs#L832).
The pallet isn't used to fetch collators anyway

runtimes/peregrine/src/system/mod.rs
```diff
@@ -55,6 +55,15 @@ parameter_types! {
...
 }
 
+match_types! {
+	pub type IncreaseStakingCalls: impl Contains<RuntimeCall> = {
+		RuntimeCall::ParachainStaking(parachain_staking::Call::<Runtime>::join_candidates { .. }) |
+		RuntimeCall::ParachainStaking(parachain_staking::Call::<Runtime>::join_delegators { .. }) |
+		RuntimeCall::ParachainStaking(parachain_staking::Call::<Runtime>::delegator_stake_more { .. }) |
+		RuntimeCall::ParachainStaking(parachain_staking::Call::<Runtime>::candidate_stake_more { .. })
+	};
+}
```
Ad96el, line 65: Did you test those filter in a chopsticks' env?

runtimes/common/src/constants.rs
rflechtner: True that

runtimes/common/src/constants.rs
rflechtner: Yes, this constraints the setting determining how many candidates are selected as collators. Which has become irrelevant now, and I would want to avoid fiddling with settings where not required, given that the pallet is pretty complex and it's hard to foresee all consequences of such changes.

runtimes/peregrine/src/system/mod.rs
rflechtner: I did, and it worked just as intended

Ad96el (dismissed):
LGTM

runtimes/common/src/constants.rs
rflechtner: b2519d400cb96497720157719390e2ff440e8112

